PR #44, opened 2021-01-09
feat: allow TimeOfDay comparison against string

jimtng:
Allows the following code

```ruby
if TimeOfDay.now < '15:00'
#do something
end
```

Update README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -685,7 +685,7 @@ or
 rule 'Log an entry if started between 3:30:04 and midnight using TimeOfDay objects' do
   on_start
   run { logger.info ("Started at #{TimeOfDay.now}")}
-  between TimeOfDay.new(h: 3, m: 30, s: 4)..TimeOfDay.midnight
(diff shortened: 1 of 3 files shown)


boc-tothefuture:
👍

boc-tothefuture:
Should we add something in the README?

jimtng:
> Should we add something in the README?

I'll work on this. I also have some other ideas for TimeOfDay which may belong in a separate PR.

jimtng:
Writing docs isn't my strongest suit. I'd welcome any suggestions / changes for README

boc-tothefuture:
I pushed a test case, if you think the test case makes sense I am ready to merge.

jimtng:
> I pushed a test case, if you think the test case makes sense I am ready to merge.

cucumber tests still an alien to me. However, it does look reasonable, and I tried it and it passed.

PR #49, opened 2021-01-10
feat:support OpenHAB Actions

### LINKED ISSUE #8: Actions are not supported

jimtng:
feat: support OpenHAB Actions
resolve #8 
Add notify(msg, email:) as a shorthand for NotificationAction.send[Broadcast]Message()
Add say and play_sound
Add action to load the actions from bindings

### DIFF
--- a/docs/usage/misc/actions.md
+++ b/docs/usage/misc/actions.md
@@ -1,0 +1,78 @@
+---
+layout: default
+title: Actions
+nav_order: 2
+has_children: false
+parent: Misc
+grand_parent: Usage
+---
+
+# Actions
+
+All OpenHAB's actions including those provided by add-ons are available, notably:
+* Audio
+* Voice
+* Things
+* Ephemeris
+* Exec
+* HTTP
+* Ping
+
+From add-ons, e.g.:
+* Transformation
+* PersistenceExtensions (from Persistence add-on)
+* NotificationAction (from OpenHAB cloud add-on)
+
+For convenience, the following methods are implemented:
+| Method     | Parameters                                                    | Description                                                                                                                          |
+| ---------- | ------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------ |
+| action     | scope, thing_uid                                              | Return the Action object for the given scope and thinguid                                                                            |
+| notify     | msg, email: (optional)                                        | When an email is specified, calls NotificationAction.sendNotification. Otherwise, calls NotificationAction.sendBroadcastNotification |
+| say        | text, volume: (optional), voice: (optional), sink: (optional) | Calls Voice.say()                                                                                                                    |
+| play_sound | filename, volume: (optional), sink: (optional)                | Calls Audio.playSound()                                                                                                              |
+
+## Example
+
+Run the TTS engine and output the default audio sink. For more information see [Voice](https://www.openhab.org/docs/configuration/multimedia.html#voice)
+```ruby
+rule 'Say the time every hour' do
+  every :hour
+  run { say "The time is #{TimeOfDay.now}" }
+end
+```
+
+```ruby
+rule 'Play an audio file' do
+  every :hour
+  run { play_sound "beep.mp3", volume: 100 }
+end
+```
+
+Send a broadcast notification via the OpenHAB Cloud
+```ruby
+rule 'Send an alert' do
+  changed Alarm_Triggered, to: ON
+  run { notify 'Red Alert!' }
+end
+```
+
+Send an email using the Mail binding
+```ruby
+rule 'Send an Email' do
+  every :day
+  run do
+    mail = action('mail', 'mail:smtp:local')
+    mail.sendEmail('me@example.com', 'subject', 'message')
+  end
+end
+```
+
+Execute a command line
+```ruby
+rule 'Run a command' do
+  every :day
+  run do
+    Exec.executeCommandLine('/bin/true')
+  end
+end
+```
--- a/features/actions.feature
+++ b/features/actions.feature
@@ -10,26 +10,41 @@ Feature:  Openhab Action Support
       rule 'Notify a user' do
         on_start
         run do
-          user = 'user@openhab.com'
-          notify(user: user, msg: 'Test Alert')
+          user = 'user@example.com'
+          notify('Test Alert', email: user)
           logger.info("Notified #{user}")
          end
       end
       """
     When I deploy the rule
-    Then It should log 'Notified user@openhab.com' within 5 seconds
+    Then It should log 'Notified user@example.com' within 5 seconds
 
   @not_implemented
   Scenario: send to all users
     Given a rule
       """
-      rule 'Notify a user' do
+      rule 'Notify all user' do
         on_start
         run do
-          notify_all('Test Alert')
+          notify('Test Alert')
           logger.info("Notified All Users")
          end
       end
       """
     When I deploy the rule
     Then It should log 'Notified All users' within 5 seconds
+
+  Scenario: Execute a command line
+    Given a rule
+      """
+      java_import java.time.Duration
+      rule 'Execute command line' do
+        on_start
+        run do
+          output = Exec.executeCommandLine(Duration.ofSeconds(2), '/bin/echo', 'Hello, World!').chomp
+          logger.info("executeCommandLine output: '#{output}'")
+        end
+      end
+      """
+    When I deploy the rule
+    Then It should log "executeCommandLine output: 'Hello, World!'" within 5 seconds
--- a/lib/openhab/core/dsl/actions.rb
+++ b/lib/openhab/core/dsl/actions.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 require 'java'
-require 'core/duration'
+require 'openhab/osgi'
 
 module OpenHAB
   module Core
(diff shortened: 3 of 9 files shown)


jimtng:
I removed OSGI::service_references, but I found that it's being used in spec/core/osgi_spec.rb. What is the purpose / significance of this file?

boc-tothefuture:
This is awesome! 

> I removed OSGI::service_references, but I found that it's being used in spec/core/osgi_spec.rb. What is the purpose / significance of this file?

Before moving to cucumber I was doing rspec testing, but that was not a good fit.  I am going to remove the rspec tests. 

Could we merge this but not announce it or announce it and hedge it a bit?  I wanted to take a look at the actions and see if there are ruby wrapper we should be put in place for them.  

Just as an example, I think notify instead of being
```
NotificationAction.broadcast or NotificationAction.notify
```
we would have
```
notify(msg, user:) # Where email is optional, with no email its broadcast otherwise its a specific user.
```

So not sure if we want to announce this just to potentially change it very soon..  WDYT?

jimtng:
I didn't know you could do `NotificationAction.broadcast`. I thought it's `NotificationAction.sendBroadcastNotification` ?
IMO we could just allow this as is, so people can use NotificationAction.sendBroadcastNotification. This will ease migration from jython.

We could implement `notify(msg, user)` in addition to this, which will use the underlying NotificationAction anyway?

I could write this `notify` (or you could, since I'm about to go to sleep), and push it into this same PR so it's done now, to avoid an extra feat: entry in the changelog? :) 

Would it go in the same actions.rb ?

jimtng:
struggling with git... please hold.

jimtng:
I'm not sure how to write the test for this. Firstly we need to check whether the relevant addon is installed, e.g. to test `notify` the OpenHAB Cloud Connector needs to be installed. Then how could you verify whether the notification was sent by the cloud?

About the easiest one to create a test for is probably the Thing action.

Or the test could just verify that the standard classes are available?

jimtng:
This should wait for the metadata feature + its osgi implementation

jimtng:
should be ready to be merged

jimtng:
How do we get cucumber to install a binding in the test environment? I'd like to install the openhab cloud and mail binding to be able to test some of the actions.

lib/openhab/core/dsl/actions.rb
```diff
@@ -10,9 +10,78 @@ module DSL
...
+        #
+        # @return [void]
+        #
+        def notify(msg, email: nil)
+          unless NotificationAction
+            raise NoMethodError, 'NotificationAction is not available. Please install the OpenHAB cloud addon'
+          end
+
+          if email
+            NotificationAction&.sendNotification email, msg
```
boc-tothefuture, line 51: You put the check above with a good error message.  This would now fail silently here with the &. is that what we want?  Or I guess the question I am asking is why use the &. here?

lib/openhab/core/dsl/actions.rb
```diff
@@ -10,9 +10,78 @@ module DSL
...
+        #
+        # Play an audio file via OpenHAB sound service, Audio.playSound()
+        #
+        # @param filename [String] The sound file to play
+        # @param sink [String] Specify a particular sink to output the speech
+        # @param volume [PercentType] Specify the volume for the speech
+        #
+        # @return [void]
+        #
+        def play_sound(filename, sink: nil, volume: nil)
```
boc-tothefuture, line 81: I don't use playSound.. is the sink required or is it optional?

lib/openhab/core/dsl/actions.rb
```diff
@@ -10,9 +10,78 @@ module DSL
...
+
+        #
+        # Return an OpenHAB Action object for the given scope and thing
+        #
+        # @param scope [String] The action scope
+        # @param thing_uid [String] Thing UID
+        #
+        # @return [Object] OpenHAB action
+        #
+        def action(scope, thing_uid)
```
boc-tothefuture, line 33: Are actions always attached to things?

features/actions.feature
```diff
@@ -10,26 +10,41 @@ Feature:  Openhab Action Support
...
+    Then It should log 'Notified All users' within 5 seconds
+
+  Scenario: Execute a command line
+    Given a rule
+      """
+      java_import java.time.Duration
+      rule 'Execute command line' do
+        on_start
+        run do
+          output = Exec.executeCommandLine(Duration.ofSeconds(2), '/bin/echo', 'Hello, World!').chomp
```
boc-tothefuture, line 44: Should we support '2.seconds' here instead?  
WDYT?

docs/usage/misc/actions.md
```diff
@@ -0,0 +1,78 @@
...
+  run { notify 'Red Alert!' }
+end
+```
+
+Send an email using the Mail binding
+```ruby
+rule 'Send an Email' do
+  every :day
+  run do
+    mail = action('mail', 'mail:smtp:local')
```
boc-tothefuture, line 64: For the other elements in our library we have items[], things[], etc. 
Should this be:
actions['mail']? 

And I make this comment below, but if actions are attached to things, should it be:
```
things['mail:smtp:local].sendEmail()?
```

I haven't uses actions beyond the basic notify, exec, etc.

lib/openhab/core/dsl/actions.rb
jimtng: The & is not necessary, but I think it makes no difference. I can remove it if you prefer?

lib/openhab/core/dsl/actions.rb
jimtng: all but the filename are optional. When not provided, it'll use the default

lib/openhab/core/dsl/actions.rb
jimtng: It just translates to $actions.get(scope, thinguid). AFAIK both are required, so yes, it seems to be always attached to things. The "scope" is which binding provides it.

features/actions.feature
jimtng: This is just how the raw Exec action works. We can provide a special method e.g. exec_command_line which we can support `2.seconds`. I am not yet familiar with the duration handling.

docs/usage/misc/actions.md
jimtng: I'm not sure either actions['mail'] syntax nor things['mail:smtp:local'] syntax would work. Certainly not the things[] syntax.

Perhaps actions[scope][thinguid].methodname_of_the_action could work?

Another example for you (from jython):
```
actions.get("telegram", "telegram:telegramBot:mybotname").sendTelegram(notification_msg)
actions.get("lgwebos", thing.getUID().toString()).launchBrowser('https://www.openhab.org')
actions.get("mqtt", "mqtt:broker:mosquitto").publishMQTT("zigbee/bridge/config/permit_join", "")
```

It is always a pair of `[ bindingname, thingid ]`

boc-tothefuture:
> How do we get cucumber to install a binding in the test environment? I'd like to install the openhab cloud and mail binding to be able to test some of the actions.

Look at thing.feature, it installs the astro binding in the background section.

docs/usage/misc/actions.md
jimtng: If I were using multiple calls to the action, I'd do:
```
livingroomtv_actions = actions['lgwebos']['livingroomtv:thing:uid']
livingroomtv_actions.launchBrowser(...)
livingroomtv_actions.dosomethingelse(....)
```

lib/openhab/core/dsl/actions.rb
boc-tothefuture: I think we should remove it.  The &. indicates we are Ok with the result being nil and I don't think we are here.  I think we want to fail.

features/actions.feature
boc-tothefuture: I am not sure we need to... I need to experiment, It is possible if in the Ruby duration class we create a method called "to_java" and have it return the Java Duration here, Jruby it might just automatically convert it for us.

docs/usage/misc/actions.md
boc-tothefuture: Can I play around for a few minutes?  If we don't like it we don't have to use it? I think there might be a way.  If it ends up too messy, we can skip it.  
What would be the best and easiest binding to test an action with?

features/actions.feature
boc-tothefuture: Do you want me to check this?

docs/usage/misc/actions.md
jimtng: Is there a Ruby construct that lets you have actions[scope][thinguid] translate into $actions.get(scope, thinguid) without having to create two levels of objects?

This is what I can think of with my limited knowledge:
```
        class ActionThingUid
          def initialize(scope)
            @scope = scope
          end

          def [](thing_uid)
            $actions.get(@scope, thing_uid)
          end
        end

        class ActionScope
          def initialize
            @scopes = []
          end

          def [](scope)
            @scopes[scope] = ActionThingUid.new(scope) unless @scopes.key? scope
            @scopes[scope]
          end
        end

        def actions
          @actions ||= ActionScope.new
        end
```

jimtng: > What would be the best and easiest binding to test an action with?

Try the mail binding. You'll have to configure a thing to use, e.g.

```
Thing mail:smtp:local [ hostname="localhost", sender="openhab@yourdomain.com" ]
```

Then you could use $actions.get('mail', 'mail:smtp:local').sendEmail(toaddress, subject, message)

features/actions.feature
jimtng: > Do you want me to check this?

yes please

lib/openhab/core/dsl/actions.rb
jimtng: I'll remove the &, just waiting on the actions[][] and duration below

features/actions.feature
boc-tothefuture: Ok, talked with the JRuby group and its not possible.  If we want this it would be better to return Java.Time.Duration rather than the duration object I built.  I will open an story to look at that. But it shouldn't hold this up.

docs/usage/misc/actions.md
jimtng: I've tested my code, so I can implement either 
```
actions('mail', 'mail:smtp:local').sendEmail(...)
```
or
```
actions['mail']['mail:smtp:local'].sendEmail(...)
```

The first option is simpler in implementation, but the second one isn't too complicated although it involves creating new objects and variables. 

Which one do you prefer, @boc-tothefuture ?

boc-tothefuture: I am working on something here. Will be a few but then I can share it for your feedback.

lib/openhab/core/dsl/actions.rb
```diff
@@ -10,9 +10,78 @@ module DSL
...
       module Actions
-        #  def notify(user:, msg:); end
+        java_import org.openhab.core.library.types.PercentType
+        include Logging
 
-        #  def notifyAll(message); end
+        NotificationAction = nil # Avoid the (NameError) uninitialized constant error
+        OpenHAB::OSGI.services('org.openhab.core.model.script.engine.action.ActionService')&.each do |service|
+          java_import service.actionClass.to_s
+          logger.info("Loaded ACTION: #{service.actionClass}")
```
jimtng, line 19: I forgot to change this to trace. Will do this on the next commit

boc-tothefuture:
Can you take a look here?  https://github.com/boc-tothefuture/openhab-jruby/tree/osgi-actions

It is not complete because it doesn't yet work with the automatic JRubyized versions of methods (send_mail instead of sendMail).  I am waiting to hear back from the JRuby team on how to handle that.  Also, missing YARD documentation, testing, etc.   

But I wanted to share and get feedback.  The basic premise is that actions are attached to things and therefore, should be methods on things.  So for example, this now works:

```ruby
  things['mail:smtp:test'].sendMail('noone@nowhere.com', 'test', 'test')
```

I think we should keep the work you did on the action method as well to ease conversion from Jython and others.  

WDYT?

boc-tothefuture:
I have all your changes in my branch as well.. is the best course of action here to close this PR and open the one from my branch for discussion?  Let me know what you think.

jimtng:
> Can you take a look here? https://github.com/boc-tothefuture/openhab-jruby/tree/osgi-actions

I got an error:
```
org.jruby.embed.EvalFailedException: (NoMethodError) undefined method `action_list' for #<Java::OrgOpenhabCoreAutomationModuleScriptInternalDefaultscope::ScriptThingActions:0x2184bce>
```

I haven't had a chance to make it work yet, but in general, the syntax would be good.



> I have all your changes in my branch as well.. is the best course of action here to close this PR and open the one from my branch for discussion? Let me know what you think.

Sure, it doesn't matter as long as it gets merged. Right now my 'production' environment is a mish mash of too many things. It would be nice to reset and use the main branch with only a few things that I'm playing with.

boc-tothefuture:
Yeah.. I forgot to add a critical file in that first push.

PR #52, opened 2021-01-10
feat: TimeOfDay.parse understands am/pm

jimtng:
This will make TimeOfDay.parse accept '7pm', '7 pm', '7:00pm', and '7 PM' as 19:00. The check is delegated to Java's DateTimeFormatter, so '13pm' will raise an error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1477,16 +1477,16 @@ end
 
 TimeOfDay class can be used in rules for time related logic. Methods:
 
-| Method      | Parameter | Description                                                 | Example                                                  |
-| ----------- | --------- | ----------------------------------------------------------- | -------------------------------------------------------- |
-| parse       | String    | Creates a TimeOfDay object with a given time string.        | curfew_start = TimeOfDay.parse '19:30'                   |
-| now         |           | Creates a TimeOfDay object that represents the current time | TimeOfDay.now > curfew_start, or TimeOfDay.now > '19:30' |
-| MIDNIGHT    |           | Creates a TimeOfDay object for 00:00                        | TimeOfDay.MIDNIGHT                                       |
-| NOON        |           | Creates a TimeOfDay object for 12:00                        | TimeOfDay.now < TimeOfDay.NOON                           |
-| constructor | h, m, s   | Creates a TimeOfDay with the given hour, minute, second     | TimeOfDay.new(h: 17, m: 30, s: 0)                        |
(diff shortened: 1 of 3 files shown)


boc-tothefuture:
Do you want to take a look [here](https://github.com/jackc/tod/blob/master/lib/tod/time_of_day.rb) and compare? 

It should probably raise an error, but I am open to discussion.

jimtng:
Perhaps it's better to raise an error rather than causing a silent logic error.

jimtng:
I'm encountering really weird bugs on Java DateTimeFormatter, and also Locale, when running in the cucumber test environment (tmp/openhab). The code runs fine on my actual openhab instance but caused an error that seems to originate from inside Java. It doesn't even have Locale.ENGLISH, very strange. In the process of troubleshooting, I tried updating openhab to 3.0.0 release (#54), deleted the tmp/openhab and re-installed it using rake, but the error persisted.

jimtng:
This should be good to go now.

boc-tothefuture:
> I'm encountering really weird bugs on Java DateTimeFormatter, and also Locale, when running in the cucumber test environment (tmp/openhab). The code runs fine on my actual openhab instance but caused an error that seems to originate from inside Java. It doesn't even have Locale.ENGLISH, very strange. In the process of troubleshooting, I tried updating openhab to 3.0.0 release (#54), deleted the tmp/openhab and re-installed it using rake, but the error persisted.

What was the issue here, how did you fix it?

jimtng:
> > I'm encountering really weird bugs on Java DateTimeFormatter, and also Locale, when running in the cucumber test environment (tmp/openhab). The code runs fine on my actual openhab instance but caused an error that seems to originate from inside Java. It doesn't even have Locale.ENGLISH, very strange. In the process of troubleshooting, I tried updating openhab to 3.0.0 release (#54), deleted the tmp/openhab and re-installed it using rake, but the error persisted.
> 
> What was the issue here, how did you fix it?

TBH, I am not sure what was wrong. It seems that by changing the "am/pm" to lowercase (`string.downcase`) instead of uppercase, everything was good again. It got me stumped for many hours.

jimtng:
LocalTime.parse with DateTimeFormatter supports recognising optional fields, so 'h', 'hh', 'h:mm', and 'h:mm:ss' can be parsed without the extra code. Cucumber tests confirmed it.

PR #53, opened 2021-01-11
fix: cucumber deploy_rule checking

jimtng:
Cucumber test issued an error when checking whether the rule has been deployed when it shouldn't be.

When there is a runtime error in the rule, it's a rule problem, not a deployment problem. Before this fix, the test will fail saying that the deployment failed, not the test that failed. This is because it was checking for the log line that was inserted at the end of the rule. So when there's an error in the rule, that "Processing Complete for uid" never made it to the log.

This PR moves the logging to say "Processing Started for uid" before the test rule is executed, and the deploy_rule looks for this instead.

This solved two types of problems:
- As per PR #43 when the test rule has "sleep" in it, the Processing Complete didn't get written to the log file until after the script had finished executing. This makes it too late for checking item states that changed before sleep started.
- When there's a runtime error in the rule/script, the Processing Complete logging is never reached. This makes deploy_rule think that the rule failed to deploy when it had actually been deployed and run.

This PR needs to be merged first before #43 so that the cucumber test on that PR could work.

### DIFF
--- a/features/step_definitions/openhab_rules.rb
+++ b/features/step_definitions/openhab_rules.rb
@@ -1,13 +1,25 @@
-# frozen_string_literal: true
+# frozen_string_literal: false
 
 require 'securerandom'
 
+def require_openhab
+  "require 'openhab'"
+end
+
 def doc_string_to_rule(doc_string)
-  "require 'openhab'\n\n#{doc_string}\n"
+  "#{require_openhab}\n\n#{doc_string}\n"
+end
+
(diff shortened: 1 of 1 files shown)


boc-tothefuture:
Did you test this against the rest of the tests?  The original code is trying to catch when the rule has been processed so that it knows it can move onto the next step.  There is a fair amount of timing involved.

boc-tothefuture:
I see in the other PR that yes in fact you did! 

Thanks!

PR #64, opened 2021-01-12
fix: TimeOfDay.parse am/pm case insensitive

jimtng:
DateTimeFormatter uses the system Locale. US Locale needs upper case AM/PM whilst some other locale need lowercase am/pm. This PR creates a DateTimeFormatter that is case insensitive.

### DIFF
--- a/lib/openhab/core/dsl/time_of_day.rb
+++ b/lib/openhab/core/dsl/time_of_day.rb
@@ -13,7 +13,8 @@ module OpenHAB
       # @since 0.0.1
       module Tod
         java_import java.time.LocalTime
-        java_import java.time.format.DateTimeFormatter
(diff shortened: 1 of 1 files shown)


jimtng:
The alternative implementation is by using string.upcase and using DateTimeFormatter with Locale::US. This would result in shorter code.

```
            local_time = LocalTime.parse(string.upcase, DateTimeFormatter.ofPattern(format, Locale::US))
```

The version that I submitted is longer, but makes it clear that it's case insensitive by calling  `parseCaseInsensitive`. I'm still specifying a locale in case some other locale could have their time interpretation in some unexpected way. I'm no expert in internationalisation. Either version is fine I think. WDYT?

PR #70, opened 2021-01-14
feat: add features for metadata

jimtng:
Added some features to the new metadata support. Suggestions / criticisms / outright rejection are welcome. I am just presenting some ideas and the code is admittedly not the greatest, and I'd appreciate any suggestions.

Please see the README section on Metadata for a list of added features.

In this proposal, the MetadataItem class can be used by the user to create a complete metadata value and configuration, as it's the only way to do it in one step. To do so, currently it needs to make a long reference to `OpenHAB::Core::DSL::MonkeyPatch::Items::Metadata::MetadataItem.new`

I've also created some test scenarios but it is not comprehensive. I can add more scenarios later. The point of this now is to present the implementation ideas.


boc-tothefuture:
Wow!  
I would say this looks pretty good.  I have been heavily focused on trying to get builds to work so we can have confidence in our releases so I have not picked this up.

```
Item1.meta['namespace1'] = MetadataItem.new(value: 'new', config: {"x"=>"a", "y"=>"b"})
```
The only thing that jumps out to me in the flow is that one.  Why do we have to do the MetadataItem.new? 
```
Item1.meta['namespace1']= value: 'new', config: {"x"=>"a", "y"=>"b"}
```
I haven't tested but would that work?  Or does ruby not allowed named arguments for []=?

jimtng:
I did some experiments, and this syntax is the one I find to work. It is passed as an array. I tried named parameters, didn't seem to work, but don't take my word for it. There may be a better way, I'm still new to ruby!

```
Item1.meta['namespace1'] = 'new',  {"x"=>"a", "y"=>"b"}
```

jimtng:
And just pushed another one, allowing to set just the config, without touching the value.

Full examples:

```ruby
# Check namespace's existence
Item1.meta['namespace'].nil?
Item1.meta.key?('namespace')

# Access item's metadata value
Item1.meta['namespace1'].value

# Access namespace1's configuration
Item1.meta['namespace1']['config1']

# Set item's metadata value, preserving its config
# Item1's metadata before: { namespace1="value" [ config1="foo", config2="bar" ] }
Item1.meta['namespace1'].value = 'new value'
# Item1's metadata after: { namespace1="new value" [ config1="foo", config2="bar" ] }

# Set item's metadata config, preserving its value
# Item1's metadata before: { namespace1="value" [ config1="foo", config2="bar" ] }
Item1.meta['namespace1'].config = { 'scooby'=>'doo' }
# Item1's metadata after: { namespace1="value" [ scooby="doo" ] }

# Set a namespace to a new value and config in one line
# Item1's metadata before: { namespace1="value" [ config1="foo", config2="bar" ] }
Item1.meta['namespace1'] = 'new value', { 'scooby'=>'doo' }
# Item1's metadata after: { namespace1="new value" [ scooby="doo" ] }

# Set item's metadata value and clear its previous config
# Item1's metadata before: { namespace1="value" [ config1="foo", config2="bar" ] }
Item1.meta['namespace1'] = 'new value'
# Item1's metadata after: { namespace1="value" }

# Set item's metadata config, set its value to nil, and wiping out previous config
# Item1's metadata before: { namespace1="value" [ config1="foo", config2="bar" ] }
Item1.meta['namespace1'] = { 'newconfig' => 'value' }
# Item1's metadata after: { namespace1=nil [ config1="foo", config2="bar" ] }

# Update namespace1's specific configuration, preserving its value and other config
# Item1's metadata before: { namespace1="value" [ config1="foo", config2="bar" ] }
Item1.meta['namespace1']['config1'] = 'doo'
# Item1's metadata will be: { namespace1="value" [ config1="doo", config2="bar" ] }

# Add a new configuration to namespace1
# Item1's metadata before: { namespace1="value" [ config1="foo", config2="bar" ] }
Item1.meta['namespace1']['config3'] = 'boo'
# Item1's metadata after: { namespace1="value" [ config1="foo", config2="bar", config3="boo" ] }

# Delete a config
# Item1's metadata before: { namespace1="value" [ config1="foo", config2="bar" ] }
Item1.meta['namespace1'].delete('config2')
# Item1's metadata after: { namespace1="value" [ config1="foo" ] }

# Add a namespace and set it to a value
# Item1's metadata before: { namespace1="value" [ config1="foo", config2="bar" ] }
Item1.meta['namespace2'] = 'qx'
# Item1's metadata after: { namespace1="value" [ config1="foo", config2="bar" ], namespace2="qx" }

# Enumerate Item1's namespaces
Item1.meta.each { |namespace, value, config| logger.info("Item1's namespace: #{namespace}='#{value}' #{config}") }

# Delete a namespace
Item1.meta.delete('namespace1')

# Delete all metadata of the item
Item1.meta.clear

# Does this item have any metadata?
Item1.meta.any?
```

boc-tothefuture:
This is really coming together.  Now that I have builds working I can look at this in detail.

jimtng:
I'll add some more test scenarios

jimtng:
I've added some more test scenarios. I could use help on the docs.

boc-tothefuture:
Can you merge the main branch in? That should pull in all the testing too.

boc-tothefuture:
I am playing around a bit to see if we can:

`TestSwitch.meta['test']= config: 'bit', value: { 'x' => 'y' }`

to work.. but it doesn't appear easy.

jimtng:
> Can you merge the main branch in? That should pull in all the testing too.

I've merged the main branch, although, since this PR is against your `metadata` branch, I think you should merge on your branch so this PR doesn't show all the merged files too.

jimtng:
> I am playing around a bit to see if we can:
> 
> `TestSwitch.meta['test']= config: 'bit', value: { 'x' => 'y' }`
> 
> to work.. but it doesn't appear easy.

I tried this too, but didn't work. I guess the []= is an assignment syntax, not a function call, whereas config: and value: is a method parameter/argument syntax. That's why I had to settle with the array syntax like in the examples

To have named parameter, it'd have to be `TestSwitch.meta['test'].set value: xx, config: {xxxx: yyy}`.

boc-tothefuture:
I also gave it a try..  It is still a method (it shows up when I try and do meta programming with method_missing)..  But the parsing of it is somehow special and won't accept named arguments.


This is a tricky one in general to get right and feel natural.  I took the examples posted above, removed the comments and just read them, then I checked to see if they did what I would expect. 

I had a couple questions and wanted your thoughts on a few.

> Item1.meta.key?('namespace')
This makes sense because we are accessing it like a hash.  Should we also have: `Item1.meta.namespace? 'namespace` ?

> Item1.meta['namespace1'] = 'new value', { 'scooby'=>'doo' }
I like what you did here.  I wish we could do named parameters, but we can't. 

> Item1.meta['namespace1'] = 'new value'
I am surprised this wipes out the config, rather than just setting the value.  I think it might surprise users to. Can we preserve the config or do you think we should wipe it out? 

> Item1.meta['namespace1'] = { 'newconfig' => 'value' }
Same thing as above for the value.  Can we just make it that if you pass a string you set the value and retain the config, and if you pass a hash you set the config and retain the value. 

If that is the case than i don't think we need the config= and value= methods?  WDYT?

> Item1.meta['namespace1'].delete('config2')
Should we (and does OpenHAB support deleting the value? Do we need  `delete_value`, `delete_config` methods?

> Item1.meta['namespace2'] = 'qx'
Would `Item1.meta['namespace2'] = 'qx', {'foo' => 'bar'}` create a new namespace, set the value to 'qx' and config to the supplied hash?


Great work on this one.. Some complicated element here.




Item1.meta.each { |namespace, value, config| logger.info("Item1's namespace: #{namespace}='#{value}' #{config}") }

Item1.meta.delete('namespace1')
Item1.meta.clear
Item1.meta.any?

jimtng:
> I am surprised this wipes out the config, rather than just setting the value. I think it might surprise users to. Can we preserve the config or do you think we should wipe it out?

I believe it should wipe out the other part that is not in the assigned value. 

When you assign something to a variable, you'd expect that the variable doesn't hold anything else inside, other than what you're assigning. That's why I think it's expected that when you do `Item.meta['ns'] = 'value'` then 'value' is all that is now inside the 'ns' metadata, and nothing else. 


```ruby
a = [ 1, 2, 3 ,4 ]
a = [ 9 ]
# you'd expect that a now contains just [ 9 ] and not [ 9, 1, 2, 3, 4 ]
```

That is why `Item.meta['ns'].value = X` will do exactly that, replace the `value` with 'x'. Since you are not referring to .config, you know the config will remain untouched.

Another way to look at this:

```ruby
Item.meta['ns'] = 'value', { 'config' => 'x' } #assign metadata value and config

Item.meta['ns'] = 'value', {} #assign metadata value with an empty config
Item.meta['ns'] = 'value' #same as above, but more compact

Item.meta['ns'] = nil, { 'config' => 'x' } #assign metadata config with nil value
Item.meta['ns'] = { 'config' => 'x' } #same as above, but more compact
```




> Should we (and does OpenHAB support deleting the value? Do we need `delete_value`, `delete_config` methods?

We could, but essentially delete_value is just assigning the value to nil, and delete_config assigning an empty hash to the config, i.e. `Item.meta['ns'].value = nil`, and delete_config is just `Item.meta['ns'] = Item.meta['ns'].value`

In practice, I believe such operation is very rarely needed.

The OpenHAB Java Metadata class contains 3 elements (instance variable): MetadataKey (item name + namespace), Value, and Configuration. The value cannot be deleted because it's an instance variable of the Metadata class. It can just be set to nil, and the configuration set to an empty hash.






> Would `Item1.meta['namespace2'] = 'qx', {'foo' => 'bar'}` create a new namespace, set the value to 'qx' and config to the supplied hash?

Yes it would. I'll add another example for this to make it clear.



What might be needed more often, is initialising an item with a whole bunch of namespaces / value / config. An example would be in the helper libraries - [Area Trigger and Actions](https://openhab-scripters.github.io/openhab-helper-libraries/Python/Community/Area%20Triggers%20and%20Actions.html#metadata-structure). Example code (not yet implemented) - note the hash is in python syntax just copy pasted from the area triggers example:

```ruby
Item1.meta =
{
    "modes" {
        "Morning": {"low_lux_trigger": 5,  "hue": 100, "saturation": 100, "brightness": 10},
        "Day":     {"low_lux_trigger": 55, "hue": 255, "saturation": 100, "brightness": 10},
        "Evening": {"low_lux_trigger": 90, "hue": 255, "saturation": 100, "brightness": 10},
        "Night":   {"low_lux_trigger": 90, "hue": 240, "saturation": 100, "brightness": 10},
        "Late":    {"low_lux_trigger": 5,  "hue": 0,   "saturation": 100, "brightness": 10},
        "Party":   {"low_lux_trigger": 5,  "hue": 0,   "saturation": 100, "brightness": 10}
    },
    "actions": {
        "light_action": {
            "ON": {
                "delay": 60,
                "recurring": False
            },
            "OFF": {
                "delay": 60,
                "recurring", True
            }
        },
        "notification_action": {
            "limited": False,
            "ON": {
                "delay": 300,
                "message": "The outside doors are not all locked",
                "recurring": True
            },
            "OFF": {
                "delay": 5,
                "recurring": False
            }
        },
        "my_custom_action_function": {
            "limited": True
            "ON": {
                "custom_key_1": "custom value 1",
                "custom_key_2": "custom value 2",
                "delay": 5,
                "recurring": False
            }
        }
    }
}
```

Note also that in the above syntax `Item1.meta = {}` will / should wipe out all other namespaces because it's an assignment to `meta`. Should one wish to **add** to the existing metadata, it should be something like `Item1.meta.merge!` which we should implement also.

jimtng:
Expanding on the [ 'value', { config } ] idea:


I was thinking of adding `Item1.meta.merge!`. The format of the list of metadata:
```
{ 
'namespace1' => [ 'value', { ... } ], # value and config
'namespace2' => 'value', # value only, config assumed to be an empty hash {}
'namespace3' => { ... }, # value = nil, config provided
 }
```

Then one can do:

```ruby
Item1.meta.clear
Item1.meta['ns1'] = 'a'
more = { 'ns1' => 'b', 'ns2' => [ 'c', { 'x' => 'y' } ] }
Item1.meta.merge!(more) # overwrite 'ns1', add 'ns2'
```

Coming up in the next commit! :)

boc-tothefuture:
> I believe it should wipe out the other part that is not in the assigned value.
> 
> When you assign something to a variable, you'd expect that the variable doesn't hold anything else inside, other than what you're assigning. That's why I think it's expected that when you do Item.meta['ns'] = 'value' then 'value' is all that is now inside the 'ns' metadata, and nothing else.
> 

I get what you are saying and it certainly makes sense from a variable standpoint.  I do worry users might inadvertently wipe out parts of their namespace.

Should we do a forum post and get some feedback?  I am happy with whatever the community decides.

jimtng:
> I do worry users might inadvertently wipe out parts of their namespace.

I figured they'll live and learn?



> Should we do a forum post and get some feedback?

I'm not sure, because it would be highly opinion based, but worth a try?

jimtng:
> I get what you are saying and it certainly makes sense from a variable standpoint. I do worry users might inadvertently wipe out parts of their namespace.

While working on Item.meta.merge! implementation, this issue comes up again. It is important that the other part that's not assigned is set to blank / empty, IMO

Consider this code:

```ruby
    TestItem.meta.clear
    config1 = { 'config1' => 'bing', 'config2' => 'bong' }
    config2 = { 'config_new' => 'y' }
    TestItem.meta['namespace3'] = 'ns3', config1
    meta = { "namespace3" => ['overwrite', config2] }
    TestItem.meta.merge!(meta,   { 'ns2' => ['value2', config1 ]}) do |key, old, new_value| 
      if key == 'namespace3'
        'newvalue for namespace3'  
        # the return value here adheres to the same syntax as Item.meta['namespace'] = xxx 
        # it can return a string (which becomes the value), hash (which becomes the config)
        # or an array [ 'value', { 'config' => 'config_value' } ]
        # if this doesn't clear the "existing" config, which config should it have, 
        # config1 (the old one in the existing namespace), or
        # config2 (the one being merged in) ?
      else
        new_value # this could be any of the 3 possible types as above
      end
    end    
```

The merge! method is called with two hashes, one will overwrite the existing `namespace3` and the other is a new namespace `ns2`.

merge! was given a block which decides / returns the new value for the overwritten `namespace3` and the return value of the block is **assigned** to namespace3. This is where what happens when an assignment occurred matters.

Would it have been just as obvious there that the non-specified part will be preserved, or would it be more obvious that it gets cleared and only the returned value/config/value+config becomes the "new" metadata for that namespace and nothing else remains?

I understand that this is still a matter of opinion

jimtng:
I will change the Item1.meta.each implementation

Previously I did it like this:
```ruby
      TestSwitch.meta.each do |namespace, value, config|
      logger.info("TestSwitch namespace #{namespace} is: \"#{value}\" #{config}")
      end
```

I will change it to:
```ruby
      TestSwitch.meta.each do |namespace, metadata|
      logger.info("TestSwitch namespace #{namespace} is: value=\"#{metadata[0]}\" config=#{metadata[1]}")
      end
```


It seems that the previous/first one is easier, but the latter / second one is more consistent with the representation of `[ value, config ]` array.

boc-tothefuture:
> I will change the Item1.meta.each implementation
> 
> Previously I did it like this:
> 
> ```ruby
>       TestSwitch.meta.each do |namespace, value, config|
>       logger.info("TestSwitch namespace #{namespace} is: \"#{value}\" #{config}")
>       end
> ```
> 
> I will change it to:
> 
> ```ruby
>       TestSwitch.meta.each do |namespace, metadata|
>       logger.info("TestSwitch namespace #{namespace} is: value=\"#{metadata[0]}\" config=#{metadata[1]}")
>       end
> ```
> 
> It seems that the previous/first one is easier, but the latter / second one is more consistent with the representation of `[ value, config ]` array.

I think I liked the first one better.  While it is more consistent with the fact we are treating it as an array (the end user doesn't know that because it looks just comma separated and they are arguments.  I think we are better off with the three block for namespace, value and config.

jimtng:
Merged a lot of changes from main branch

jimtng:
This PR seems quite messy because a lot of things have changed since, and now there are many unrelated files included. Should we start with a clean slate? What other reasons are holding it up?

boc-tothefuture:
Hey @jimtng  - My bad..I thought you were still working on it.  Let me take a look, I didn't realize you were finished.

boc-tothefuture:
There are a lot of files in it now.. you are correct.. did you merge or rebase off the main branch to pull in the updates?

boc-tothefuture:
Now that you merged is it too late to rebase instead?

jimtng:
I tried rebasing, but it became even more messy. I'll close this and submit a new PR / branch.

PR #77, opened 2021-01-20
feat: Add SwitchItem.toggle

### LINKED ISSUE #27: SwitchItem.toggle
Is there a reason why having SwitchItem.toggle is a bad idea?

I can do a PR, essentially adding this, unless there's a better way
```
  def toggle
    command(on? ? OnOffType::OFF : OnOffType::ON)
  end
```

jimtng:
resolve #27


boc-tothefuture:
Looks good to me.
Do you want to update the version number?  Should go to 2.7.0.  Then when merged it should be automatically released.

boc-tothefuture:
@jimtng you want to announce this one in the thread?

PR #85, opened 2021-01-25
fix: set the rule name as the logger name

### LINKED ISSUE #84: logger doesn't show the rule name

```ruby
rule 'rule1 ' do
  on_start
  run do
  end
end
```

```
```

What it should say is:
```
```

jimtng:
Fix #84 

Perhaps this fix is a bit crude, but it seems to work. @boc-tothefuture WDYT?

PS The version number is set so the other PR should be merged first.


boc-tothefuture:
You are right this does seem a bit hacky..  We probably need a way to associate an object with a name in the logger method so that during rule construction we set the name.

jimtng:
Could we merge this now and fix it later with a more elegant solution?

boc-tothefuture:
Let me know if #126 captures this?

jimtng:
> Let me know if #126 captures this?

Yes, it's a better solution

PR #87, opened 2021-01-25
fix: inject event.item for group update trigger

### LINKED ISSUE #75: run execution block's event.item for group triggers
It seems that for group triggers, the run execution block event doesn't have the event.item instance variable.

Code:
```
Group MyGroup
Switch Item1 (MyGroup)
```

```ruby
rule 'test' do
  updated MyGroup.items
  run do |event|
    logger.info("TEST: #{event.item}")
  end
end
```

This will cause an exception error 

```
10:00:50.602 [DEBUG] [re.automation.internal.RuleEngineImpl] -
java.lang.RuntimeException: Fail to execute action: 1
	at org.openhab.core.automation.internal.RuleEngineImpl.executeActions(RuleEngineImpl.java:1187) ~[bundleFile:?]
	at org.openhab.core.automation.internal.RuleEngineImpl.runRule(RuleEngineImpl.java:987) [bundleFile:?]
	at org.openhab.core.automation.internal.TriggerHandlerCallbackImpl$TriggerData.run(TriggerHandlerCallbackImpl.java:89) [bundleFile:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) [?:?]
	at java.util.concurrent.FutureTask.run(FutureTask.java:264) [?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) [?:?]
	at java.lang.Thread.run(Thread.java:834) [?:?]
Caused by: org.jruby.exceptions.NoMethodError: (NoMethodError) undefined method `item' for #<Java::OrgOpenhabCoreItemsEvents::ItemStateEvent:0x65207b6c>
Did you mean?  items
	at RUBY.<main>(<script>:7) ~[?:?]
	at RUBY.process_queue(/openhab/conf/automation/lib/ruby/core/openhab/core/dsl/rule/rule.rb:209) ~[?:?]
	at RUBY.execute(/openhab/conf/automation/lib/ruby/core/openhab/core/dsl/rule/rule.rb:183) ~[?:?]
```

Could we make it consistent and add `event.item` even for group triggers?
Using `triggered` works.

jimtng:
Fix #75 

This would be a temporary patch, until openhab core would fix it. 
I will look into the openhab-core code and see if I can submit a fix there too, and if it is accepted, I will revert this patch. I just don't know whether it will be accepted as they're stricter there

WDYT?

### DIFF
--- a/features/patched.feature
+++ b/features/patched.feature
@@ -1,0 +1,32 @@
+Feature:  Temporary patches
+
+  Background:
+    Given Clean OpenHAB with latest Ruby Libraries
+    And groups:
+      | name         | group |
+      | Temperatures |       |
+    And items:
+      | type   | name            | label                   | state | groups       |
+      | Number | Livingroom_Temp | Living Room temperature | 70    | Temperatures |
+      | Number | Bedroom_Temp    | Bedroom temperature     | 50    | Temperatures |
+      | Number | Den_Temp        | Den temperature         | 30    | Temperatures |
+
+
+  Scenario: Group update trigger has event.item in run block
+    Given code in a rules file
+      """
+      rule 'group member updated' do
+        updated Temperatures.items
+        run do |event|
+          logger.info("event.item is #{event.item.name}")
+        end
+      end
+
+      rule 'update a group member' do
+        on_start
+        run { Livingroom_Temp.update 65 }
+      end
+      """
+    When I deploy the rules file
+    Then It should log 'event.item is Livingroom_Temp' within 5 seconds
+
--- a/lib/openhab/core/dsl/rule/item.rb
+++ b/lib/openhab/core/dsl/rule/item.rb
@@ -97,6 +97,7 @@ module OpenHAB
           #
           def updated(*items, to: nil)
             items.flatten.each do |item|
+              item = item.group if item.is_a? Group
               logger.trace("Creating updated trigger for item(#{item}) to(#{to})")
               [to].flatten.each do |to_state|
                 case item
--- a/lib/openhab/core/dsl/rule/rule.rb
+++ b/lib/openhab/core/dsl/rule/rule.rb
@@ -317,7 +317,15 @@ module OpenHAB
               when RuleConfig::Run
 
                 event = inputs&.dig('event')
-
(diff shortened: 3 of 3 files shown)


boc-tothefuture:
Looks good - can you refactor it out to its own private method?

jimtng:
> Looks good - can you refactor it out to its own private method?

I could try moving it out in a private method, but it would mean having to remove two things later on, the call to the priv method and the method itself?

boc-tothefuture:
Sorry.. I think I wasn't being clear. 
I meant just refactoring the code you have that add the item to the event to be in a private method rather than being in the process_queue method.  I need to refactor a bunch of our methods to be smaller and didn't want to add to the technical debt. If not, no big deal, i can hit it when I refactor.

Any chance we could have a test for this one since it looks like we may want to check it against future versions of the code to remove when no longer necessary?

jimtng:
I added a separate test in patched.feature

I also made a small addition to core/dsl/rule/item.rb to make it consistent for `updated GroupItem`

jimtng:
> Sorry.. I think I wasn't being clear.
> I meant just refactoring the code you have that add the item to the event to be in a private method rather than being in the process_queue method. I need to refactor a bunch of our methods to be smaller and didn't want to add to the technical debt. If not, no big deal, i can hit it when I refactor.

I'll leave it as is to be refactored by you. I had a look around openhab-core regarding this issue but, boy, is it complex in there!

boc-tothefuture:
How critical is this?  I would like to test out #91 with this, but don't want to hold something up if we need it out soon.

jimtng:
> How critical is this? I would like to test out #91 with this, but don't want to hold something up if we need it out soon.

It's a bug, but since it exists in openhab-core, and perhaps I'm the only one using jruby affected, and I've already got the patch on my installation, it can wait, especially if #91 is almost ready.

boc-tothefuture:
closed per request by @jimtng for me to refactor

PR #88, opened 2021-01-26
feat: Action Support

boc-tothefuture:
@jimtng  Can you take a look and provide comments.  This is still a WIP as I wait for a response from the Jruby team.

Missing Items:

- [ ] support for JRuby auto created methods `send_email`, etc.
- [x] Working tests
- [x] Documentation around usage from a `thing` object


boc-tothefuture:
@jimtng Do you have a good idea on how we could safely (without exposing credentials) test notifications?

boc-tothefuture:
JRuby doesn't currently support a way to do this with the JRuby auto created methods.  JRuby devs are looking at, so that would be something we can tackle in a future release. 

@jimtng if we can figure out testing for notify and you are good with this I think we can merge it.

jimtng:
I wonder if the `action` method should be called `actions`. The jython version is  `actions.get` and RulesDSL `getActions`

jimtng:
> @jimtng Do you have a good idea on how we could safely (without exposing credentials) test notifications?

Sorry I can't think of any practical way. The best we could do is to see if there's an exception i.e. issue a log.info afterwards and if the execution reaches it, we could assume there's no error.

One would argue that we don't need to test the notification system itself. We just need to test that the call has been made successfully, because the notification system itself should have been tested by the binding / notification service, not by us.

jimtng:
> @jimtng if we can figure out testing for notify and you are good with this I think we can merge it.

We need to install the `openhabcloud` binding/addon for it to work.

boc-tothefuture:
I agree, it is out of our scope to test that notifications work, just that our call to notify works to invoke the action.  

I am fine with renaming action to actions.  I can do that.  I am going to open a separate issue on testing w/ notify since I don't think that should hold up this PR.

boc-tothefuture:
@jimtng  -  If you are happy with this, I will merge it.
Let me know.

jimtng:
On my production environment, Inside actions.rb method actions_for_thing, I changed the logger.trace to logger.info so I could see the logging.

Then just by having one line `things['mail:smtp:local'].sendMail()` it seems that actions_for_thing was called many many times (didn't count, but I'd say at least 30+ times). Far too many times.

Also it would be good to do this as a 'lazy load' so that only getting the action when needed for that particular thinguid, instead of preloading all.

boc-tothefuture:
> It would be good to do this as a 'lazy load' so that only getting the action when needed for that particular thinguid, instead of preloading all.
> 
> Maybe utilising method_missing()?

We could, but it wouldn't be simple because we are already delegating all method_missing calls (via the SimpleDelegator) to the actual Thing object.  I will need to build our own "multi-delegator" class, which I probably need to do for the refactoring away from monkeypatching the items so they are consistent. I would rather wait until then. 

The actual calls should be very quick.  Can we open an issue on this as a future performance enhancement but merge the functionality now?  WDYT?

jimtng:
> Can we open an issue on this as a future performance enhancement but merge the functionality now?

OK

boc-tothefuture:
looks like you already opened one.. great.

Issue #55, opened 2021-01-11
Package and distribute this library as a Ruby Gem


boc-tothefuture:
Distribute this library as a RubyGem

This requires updating the OpenHAB JRuby implementation to accept as a configuration item a list of gems which get installed.

This should make installation, upgrade, etc much easier as user can just specify this gem at a specific version and it would be installed.

Additionally, this means this gem could start relying on other gems as part of its function.  For example, it could rely on the time of day gem rather than repeat some of its functionality.

jimtng:
This is awesome!

PR #97, opened 2021-01-28
Install JRuby Helper Library as a Ruby Gem

boc-tothefuture:


boc-tothefuture:
This addresses #55 

@jimtng - Could you look at this and test it out?

jimtng:
Trying out the gem version now. Question, how can I overwrite / customise / experiment with the library when it's installed as a gem? Can I just change the files inside the gem directory?

boc-tothefuture:
> Trying out the gem version now. Question, how can I overwrite / customise / experiment with the library when it's installed as a gem? Can I just change the files inside the gem directory?

That should work.  You can also do a 'rake build' and then do what 'rake openhab:deploy' does which is sets your GEM_HOME to point to the same gem_home dir and then run gem install ./pkg/<gem-version>.  That will update the gem.

jimtng:
It seems to work. I've just installed it. It needed the newer version of the openhab addon jar.

boc-tothefuture:
Yeah, it requires the new jar.. docs should be pointing it.

jimtng:
How do people update their gem? It would be good to document that for Ruby noobs like me.

boc-tothefuture:
> How do people update their gem - without restarting openhab? It would be good to document that for Ruby noobs like me.

You probably really shouldn't.  At least we shouldn't expose it to those not doing development.  Because of how OpenHAB works these rules are preprocessed and then stored by the OpenHAB rules engine.  So if you don't restart, some rules are going to be processed by one version of the library and updated rules by the new version. 

For development, I think in another message I explained you can just do a gem install.   but same rules would apply to a running system.

PR #98, opened 2021-01-28
docs: timer.execution_time

jimtng:
I've recently submitted a PR to openhab-core to add Timer::getExecutionTime(). It's currently available in openhab 3.1 (snapshot).


boc-tothefuture:
@jimtng 
Should I not have merged that until openhab 3.1 comes out?  We are only building and testing against 3.0.. We don't have 3.1 snapshot listed as a requirement..?

PR #99, opened 2021-01-29
docs: timer execution_time

jimtng:
Specified that execution_time is only available in openhab 3.1+


boc-tothefuture:
Thanks, quick question..  BTW, happy to have this in the docs.. and awesome that you contributed it to OpenHAB!

For most of the other objects we just note that all of the java methods are available in addition to these.  Should we do the same here?

WDYT?

jimtng:
I updated the docs

PR #106, opened 2021-01-31
fix: multiple changed delays

### LINKED ISSUE #105: Duration ignored when there are multiple conditions with duration
I found a bug

Consider this code:

```
Switch TestItem1
Switch TestItem2
```

Initial state: UNDEF

```ruby
rule 'test' do
  changed TestItem1, to: ON, for: 10.seconds
  changed TestItem2, to: ON, for: 10.seconds
  run { logger.info("XXX") }
end
```

**BUG 1**: When I set TestItem1 to ON, the log "XXX" appears immediately, not 10 seconds after.

I expected that the rule would only execute 10 seconds after TestItem1 OR TestItem2 changed to ON. If TestItem1 changed to ON then 3 seconds later TestItem2 changed to ON, the rule should fire twice. The first time it should fire 10 seconds after TestItem1 changed to on, and the second time, 10 seconds after TestItem2 changed to on.

**BUG 2**: Now TestItem1 = ON and TestItem2 = OFF. I then send a command to set TestItem1 to OFF. The log "XXX" would appear again, immediately after TestItem1 changed from ON to OFF. This should not even happen at all.

It would work fine when there's only one `changed` trigger.

Changing TestItem2 to ON yields the desired behaviour, i.e. Log XXX appears after 10 seconds, and setting TestItem2 to OFF does not fire the rule.

It seems to affect the **first** `changed` item.

jimtng:
fix #105 

Should I update the changelog and version ?

### DIFF
--- a/features/changed_duration.feature
+++ b/features/changed_duration.feature
@@ -102,3 +102,27 @@ Feature: changed_duration
       | ON   | OFF | changed Switches, to: ON, for: 10.seconds            | should not |
       | OFF  | ON  | changed Switches, from: OFF, to: ON, for: 10.seconds | should     |
       | OFF  | ON  | changed Switches, from: ON, to: ON, for: 10.seconds  | should not |
+
+  Scenario: Multiple changed with duration triggers
+    Given items:
+      | type   | name       | label      | group | state |
+      | Switch | Switch_One | Switch One |       | OFF   |
+      | Switch | Switch_Two | Switch Two |       | OFF   |
+    And a rule:
+      """
+      rule 'A rule with multiple changed duration triggers' do
+        changed Switch_One, to: ON, for: 4.seconds
+        changed Switch_Two, to: ON, for: 8.seconds
+        triggered do |item|
+          logger.info("#{item.name} changed")
+        end
+      end
+      """
+    When I deploy the rule
+    And item "Switch_One" state is changed to "ON"
+    And item "Switch_Two" state is changed to "ON"
+    Then It should not log "Switch_One changed" within 2 seconds
+    And It should not log "Switch_Two changed" within 5 seconds
+    And It should log "Switch_One changed" within 6 seconds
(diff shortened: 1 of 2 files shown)


jimtng:
changed the commit message to start with a lowercase

boc-tothefuture:
Is something we can test (to prevent regressions)?

jimtng:
How about something like this:
```
start = Time.now

rule 'test' do
  changed Item1, to: ON, for: 2.seconds
  changed Item2, to: ON, for: 4.seconds
  triggered do |item|
    logger.info("#{item.name} triggered in #{Time.now.to_i - start.to_i}")
  end
end

[Item1, Item2].each(&:off).each(&:on)
```

boc-tothefuture:
So the test is that one trigger should fire then later another one, do I have that right?

jimtng:
Not only that there should be two triggers, but trigger for item1 should occur in 2 seconds, and the other in 4 seconds. The timing matters. So we'll need to look for the log line 
* Item1 triggered in 2
* Item2 triggered in 4

The problem before was Item1 got triggered immediately, not in 2 seconds, and in fact it got triggered when it changed to OFF too - which would require a different check

boc-tothefuture:
Ok, I can take a crack at writing the feature test for it tomorrow.

jimtng:
Ahh yes, the check should also include:

```
And It should not log "Item1 triggered in 2" in 1 second
And It should not log "Item2 triggered in 4" in 3 seconds
```

jimtng:
I've written a test to check for this scenario, added to the end of changed_duration.feature. The check will fail with the previous code, and will succeed with the changes included in this PR.

jimtng:
The cucumber test worked on my system. Can you please test @boc-tothefuture ?

jimtng:
@boc-tothefuture this should be ready to merge

PR #107, opened 2021-01-31
fix: items['nonexistent'] now returns nil, add items.include?

### LINKED ISSUE #101: Checking the existence of an item
Is there a more elegant way to check if an item exists, given a name?

Currently, items['ANonExistentItem'] will cause an exception. I'd have to rescue the exception to proceed.

It would be good I think if items['ANonExistentItem'] will simply return nil.

jimtng:
Resolve #101 
Add items.include? to check for items existence
Make items['nonexistent'] to return nil. This will enable such shortcut:

```ruby
basename = 'LivingRoom'
power_item = items["#{basename}_Light_Dimmer"] ||  items["#{basename}_Light_Power"] ||  items["#{basename}_Light_Switch"]
power_item&.on
```

### DIFF
--- a/docs/usage/items.md
+++ b/docs/usage/items.md
@@ -15,6 +15,7 @@ All items can be accessed as an enumerable the `items` method.
 | Method             | Description                                                                    |
 | ------------------ | ------------------------------------------------------------------------------ |
 | []                 | Get a specific item by name, this syntax can be used to dynamically load items |
+| include?           | Check to see if an item with the given name exists                             |
 | enumerable methods | All methods [here](https://ruby-doc.org/core-2.5.0/Enumerable.html)            |
 
 ## Examples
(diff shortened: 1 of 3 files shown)


boc-tothefuture:
Do we want to generate and catch the stack trace or should we just use the include? method you wrote?

```
return nil unless include?(item)
```

jimtng:
> Do we want to generate and catch the stack trace or should we just use the include? method you wrote?
> 
> ```
> return nil unless include?(item)
> ```

Calling `return nil unless include?` will involve two function calls each and every time items[] is called, whereas catching the exception only involves one call when the item exists. It only involves the exception / rescue when it doesn't exist, so to me it's more efficient for the normal case when people are referring to existing items.

boc-tothefuture:
This really should be two commits since we shouldn't combine a fix with a feature in the same commit.  Otherwise we will get the wrong versioning from semver/conventional commits.  
Would you mind rebasing, create two commits and then force pushing with lease?

jimtng:
Shall I keep them in one PR?

boc-tothefuture:
One PR is fine.

jimtng:
I've split them into two commits

boc-tothefuture:
commit-lint is failing https://github.com/boc-tothefuture/openhab-jruby/pull/107/checks?check_run_id=1810741397

I find this super annoying but the commit messages (the part after the colon) cannot start with uppercase.  Not sure why, but that is the standard and all the tooling works off of it being formatted in a specific way.

jimtng:
Re-submitted with lower case

PR #111, opened 2021-02-01
feat: Support Persistence features

### LINKED ISSUE #50: Integrating PersistenceExtensions into items
I feel there is an opportunity to include persistence extensions (if it's available on the openhab environment) into JRuby items, so that instead of:

```ruby
PersistenceExtensions.averageSince(NumberItem1, ZonedDateTime.now().minusHours(1), 'influxdb')
```

we could do something more elegant, e.g.

```ruby
NumberItem1.average_since(......)
```

I'm not yet familiar with what to use for date/time in Ruby though, and it would need to eventually translate into java's ZonedDateTime to pass to the actual PersistenceExtensions.

jimtng:
Resolve #50 
Add Persistence methods to Items. It accepts a relative duration in addition to ZonedDateTime, e.g. 
```
Item1.average_since 12.hours #or
Item1.average_since ZonedDateTime.now.minusHours(12)
```

To specify the persistence service: (I dropped the idea of named parameter)
```
Item1.last_update 'influxdb'
Item1.sum_since 1.hour, :influxdb #it could accept a symbol
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,4 @@
 AllCops:
   TargetRubyVersion: 2.5
   NewCops: enable
+  SuggestExtensions: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    openhab-scripting (2.14.1)
+    openhab-scripting (2.14.3)
       bundler (~> 2.2)
 
 GEM
@@ -247,4 +247,4 @@ DEPENDENCIES
   yard
 
 BUNDLED WITH
-   2.2.4
+   2.2.7
--- a/docs/usage/misc/persistence.md
+++ b/docs/usage/misc/persistence.md
@@ -1,0 +1,52 @@
+---
+layout: default
+title: Persistence
+nav_order: 2
+has_children: false
+parent: Misc
+grand_parent: Usage
+---
+
+# Persistence
+
+[Persistence](https://www.openhab.org/docs/configuration/persistence.html) functions can be accessed through the item object. The following methods related to persistence are available: 
+
+| Method            | Parameters                            | Example                                                         |
+| ----------------- | ------------------------------------- | --------------------------------------------------------------- |
+| `persist`         | service                               | `Item1.persist`                                                 |
+| `last_update`     | service                               | `Item1.last_update`                                             |
+| `previous_state`  | skip_equal: (default: false), service | `Item1.previous_state` `Item1.previous_state(skip_equal: true)` |
+| `average_since`   | timestamp, service                    | `Item1.average_since(-1.hours, :influxdb)`                      |
+| `changed_since`   | timestamp, service                    |                                                                 |
+| `delta_since`     | timestamp, service                    |                                                                 |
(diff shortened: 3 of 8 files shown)


jimtng:
@boc-tothefuture if you're happy with this implementation, would you mind writing the tests? It would need to install a persistence extension (either rrd4j, or influxdb?), setting it as default, generating some data, etc.

boc-tothefuture:
I will take a look at how to test this. 

What happens if you don't specify a persistence service with the calls?

jimtng:
> What happens if you don't specify a persistence service with the calls?

It will use the system default persistence service. It is specified in the gui, or in config file although I'm not sure how exactly.

boc-tothefuture:
Is there any value, in addition to the work you did, add a persistence block to group multiple persistence operations together under a single service?  Much like we did with units?

So, just an example.
```ruby
persistence(:influxdb) do 
  Item1.average_since
  Item2.sum_since
  etc...
end
```

We would keep both styles.

boc-tothefuture:
FYI - build is failing is because a method was added without documentation.

jimtng:
I've added a yard documentation for the `previous_state` method. How do we document the other ones that are defined with define_method ?

boc-tothefuture:
How do you feel about the addition of the block method?

jimtng:
> Is there any value, in addition to the work you did, add a persistence block to group multiple persistence operations together under a single service? Much like we did with units?
> 
> So, just an example.
> 
> ```ruby
> persistence(:influxdb) do 
>   Item1.average_since
>   Item2.sum_since
>   etc...
> end
> ```
> 
> We would keep both styles.

I didn't see this message. This looks good. I wasn't aware of such usage with units until you've pointed it out here.

I'll have a play with this and update this PR.

jimtng:
@boc-tothefuture Please review the persistence block that I've just added. I've briefly tested it.

jimtng:
Note: Once #122 is merged, we'll need to remove this piece of code from persistence.rb since it is no longer relevant.

```
                if timestamp.respond_to? 'to_ms'
                  timestamp = ZonedDateTime.now.plus(Java::JavaTime::Duration.ofMillis(timestamp.to_ms))
```

jimtng:
I've removed the `to_ms` reference and also modified it to accept `TemporalAmount`, which is the parent interface of `Duration`, so technically it can also accept `java.time.Period`.

jimtng:
Please don't merge yet. I think I can write the tests too, just have been a bit busy with life + also working on another PR for doing comparisons.

jimtng:
@boc-tothefuture it's ready for review now

boc-tothefuture:
Awesome!  Thanks.  Lets merge in #130 first.

boc-tothefuture:
Do we think there is no need to test with rrd or something similar?

jimtng:
This is read to merge

> Do we think there is no need to test with rrd or something similar?

Currently it's tested by installing rrd4j and using it using the persistence block. It doesn't really verify the values. It just assumes that if the calls were successful, it will continue to the end, i.e. no errors.

boc-tothefuture:
I totally missed that it installed that the rrd4j.

PR #114, opened 2021-02-01
[Documentation] Change paths to conform with OpenHAB documentation

pacive:
See e.g. http://www.openhab.org/docs/installation/#getting-installation-info

### DIFF
--- a/docs/installation/index.md
+++ b/docs/installation/index.md
@@ -13,15 +13,15 @@ has_children: false
 
 
 ## Installation
-1. Create directory for Ruby Gems `<openhab_base_dir>/conf/automation/lib/ruby/gem_home`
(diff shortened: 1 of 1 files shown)


boc-tothefuture:
Thanks for the PR @pacive ! 

The build here is failing because the PR doesn't conform to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)

We use conventional commits because that give us automatic version bumping, automatic changelog generation, etc. 

Can you update your commit message to be the following:
docs(install): change paths to conform with OpenHAB documentation

That tells the build system it is a doc change (no need to push a new gem).  That it was for the install section of the docs and then what the change is. 

Once you update, just to a force-with-lease push.

Let me know if you need any help with the git commands!

PR #123, opened 2021-02-02
fix(number_item): make comparison with Float work

### LINKED ISSUE #121: Comparison between NumberItem and Float doesn't work
When trying to do a comparison between a number item and a float literal, e.g.
```ruby
if Sun_Position_Azimuth > 212.5
```
the rule fails to load with an error:
```
org.jruby.embed.EvalFailedException: (ArgumentError) can't omit precision for a Float.
```
I tracked this down to `lib/openhab/core/dsl/items/number_item.rb` and the conversion to `BigDecimal` for `Numeric` types, which fails for `Floats`.

This could be solved either by adding
```ruby
            when Float
              @number_item.state.to_big_decimal.to_f <=> other
```
Before the `when Numeric` expression in the `<=>(other)` method, or changing it to:
```ruby
            when Numeric
              @number_item.state.to_big_decimal.to_d <=> BigDecimal(other.to_s)
```
Which solution would be preferable? I'll create a PR with a fix ASAP!

pacive:
Fixes #121

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 JRuby OpenHAB Scripting Change Log
 
+## [2.13.1](https://github.com/boc-tothefuture/openhab-jruby/compare/2.13.0...2.13.1) (2021-02-02)
+
 # [2.13.0](https://github.com/boc-tothefuture/openhab-jruby/compare/2.12.0...2.13.0) (2021-02-02)
 
 
--- a/lib/openhab/version.rb
+++ b/lib/openhab/version.rb
@@ -5,5 +5,5 @@
 #
 module OpenHAB
   # @return [String] Version of OpenHAB helper libraries
-  VERSION = '2.13.0'
+  VERSION = '2.13.1'
 end


boc-tothefuture:
Thanks for the PR @pacive !

Can you add a test for this?  We are using cucumber for testing.  There is an existing file for number items in features/number_item.feature

If you don't know how to do cucumber testing, I would be happy to help show you. 

We have several refactors coming up and don't want to introduce any regressions.  If you want to test it locally, the dev instructions should get you a version setup to test against.  Again, just let me know if you need any help.

Really happy to have another contributor here

pacive:
I haven't worked with cucumber tests before, but should be able to figure it out. But as I mentioned in the issue (#121) this is also a problem with calculations, so would have to make some more changes. Do you have any thoughts on that?

boc-tothefuture:
Awesome, the instructions for the dev environment are here:  https://boc-tothefuture.github.io/openhab-jruby/contributing/

I would be interested in knowing which platform you are using for development and if it all works for you.  It so far has only been tested on Mac.

pacive:
I currently have the repo cloned to the linux (debian) computer I run OH on, and use VSCode SSH remote to connect to it and do the coding. But in order to test the modifications more easily I might move it locally to my windows computer where I do my other OH development and have testing installations set up. But the addon runs without problem on linux at least, I'm in the middle of migrating my rules, and thinking about improvements as I go!

pacive:
I had to close this because I screwed up my git history and had to do a reset. Will create a new one when I have everything up and running correctly.

PR #126, opened 2021-02-03
feat(logging): append rule name to logging class if logging within ru…

boc-tothefuture:
@jimtng  can you take a look at this?  I wrote a test and stole some of our code.  However, this should log the rule name when run within the rule context.


jimtng:
Works for me. Thanks!

PR #127, opened 2021-02-04
fix(number_item): make math operations and comparisons work with Floats

### LINKED ISSUE #121: Comparison between NumberItem and Float doesn't work

pacive:
Fixes #121

### DIFF
--- a/features/number_item.feature
+++ b/features/number_item.feature
@@ -23,6 +23,10 @@ Feature:  number_item
       | 50      | -        | 2       | 48    |
       | 50      | /        | 2       | 25    |
       | 50      | *        | 2       | 100   |
+      | 50      | +        | 2.0     | 52    |
+      | 50      | -        | 2.0     | 48    |
+      | 50      | /        | 2.0     | 25    |
+      | 50      | *        | 2.0     | 100   |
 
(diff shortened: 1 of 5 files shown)


pacive:
Had to give up my attempts to make development work on windows, I managed to adapt the Rakefile so the setup worked, but there was something weird going on with the tests where the commands sent to karaf was run in a separate console window, so the tests continued running before they had time to complete. So I ended up running a WSL-container with debian, and that worked flawlessly!

PR #130, opened 2021-02-04
Rubocop

boc-tothefuture:


boc-tothefuture:
@jimtng and @pacive - 
 This is a big big one..  :( Almost every file was touched.
 
  There was a lot of technical debt/anti-patterns, etc that creeped into the code and kept growing because rubocop was not enforced in the build.  This addresses all 120+ rubocop violations in the code base and documents all private methods. 

  With rubocop now enabled in the build chain, we shouldn't have to deal with this again.

  This addresses #4.

Can you take a look before I merge?  

There should be one more large refactoring (#3) before I am ready to start on new features, but I think that will be less impacting as it should just be moving files around and placing them into the right modules.  Once we have this solid foundation should enable us to not add any more debt.

jimtng:
Wow, great job! Once this is merged, I'll have to adjust my existing PRs to follow suit.

pacive:
Nice work! Will revise my pr against this when it is merged and update accordingly.

boc-tothefuture:
Merging..  hopefully our test coverage is complete.. but let me know if you hit any issues.

PR #134, opened 2021-02-06
feat: add DateTime Item type

### LINKED ISSUE #35: Add DateTime Item type

pacive:
I'm working on making DateTime Items work in the rules. I have made wrappers for both DateTimeItem and DateTimeType. This is what I have completed so far:
* DateTime Items can be updated and commanded with ruby Time objects:
  ```ruby
  Test << Time.now
  ```
  For this I had to monkey patch the Time class' `#to_s` method to make it return a string that OH's DateTimeType can parse. Hope this doesn't cause any problems elsewhere
* Calculations (+, -) work with Integers, Durations and Strings in the form of HH:MM or HH:MM:SS. Subtraction also works with other Time objects, in which case a float is returned, like in the ruby Time class.
  ```ruby
  Test << Test + 600
  Test << Test - '00:05'
  Test << Test + 5.minutes
  ```
* Comparisons between different time objects and Items work, `==` returns true if two objects have the same instant but in different time zones
* Ruby Time instance methods like `#monday?` are supported

TODO:
* ~~Make calculations work with the added Integer methods:~~
  ```ruby
  Test << Test + 5.hours
  ```
* ~~Integrate with Time of day, e.g. Ranges~~
* ~~Write more tests~~
* ~~Documentation~~

I noticed there is a name collision since the DateTime class from ruby stdlib is included. I haven't had any issues except having to refer to the class explicitly in some cases. Should I rename it to something else? Suggestions?

Suggestions for more functionality or comments on my code are very welcome!

Closes #35


pacive:
Seems it somehow messed with the `every` trigger, will look into it.

lib/openhab/core/dsl/items/datetime_item.rb
```diff
@@ -0,0 +1,78 @@
...
+            logger.trace("Comparing self (#{self.class}) to #{other} (#{other.class})")
+            case other
+            when DateTimeItem
+              to_dt <=> other.to_dt
+            else
+              to_dt <=> other
+            end
+          end
+
+          def <(other)
```
boc-tothefuture, line 39: Does implementing comparable not work here for some reason?

lib/openhab/core/dsl/items/datetime_item.rb
```diff
@@ -0,0 +1,78 @@
...
+
+          def ==(other)
+            (self <=> other).zero?
+          end
+
+          def zone
+            to_dt.zone
+          end
+
+          def method_missing(meth, *args, &block)
```
boc-tothefuture, line 63: For datetime can we SimpleDelegator rather than method missing? I don't see any special logic in here beyond what SimpleDelegator gives us.

lib/openhab/core/dsl/items/datetime_item.rb
pacive: It probably does, forgot about that, thanks!

lib/openhab/core/dsl/items/datetime_item.rb
pacive: Haven't used it much before, but you're probably right, will read up a bit on it. But we want it to delegate to DateTimeItem first, what happens if there are name conflicts? Doesn't the delegator take precedence?

pacive: I tried to use SimpleDelegator, but still had to implement `method_missing` to be able to delegate to the DateTimeType object kept as the state, and the code looked uglier:
```ruby
def method_missing(meth, *args, &block)
  super
rescue NoMethodError
  if state?
    to_dt.send(meth, *args, &block)
  end
end
```

lib/openhab/core/dsl/items/datetime_item.rb
```diff
@@ -0,0 +1,97 @@
...
+          # @param [Object] other Other objet to compare against
+          #
+          # @return [Integer] -1, 0 or 1 depending on the result of the comparison
+          #
+          def <=>(other)
+            return unless state?
+
+            logger.trace("Comparing self (#{self.class}) to #{other} (#{other.class})")
+            other_value = other.is_a?(DateTimeItem) ? other.to_dt : other
+            to_dt <=> other_value
```
jimtng, line 52:
```suggestion
            other =  other.to_dt if other.is_a? DateTimeItem
            to_dt <=> other
```

jimtng:
> * DateTime Items can be updated and commanded with ruby Time objects:
>   ```ruby
>   Test << Time.now
>   ```

Can it do this too?
```ruby
Test << ZonedDateTime.now
```

Should it be able to do this, by using the current date?
```ruby
Test << TimeOfDay.now
Test << '7pm' # use TimeOfDay.parse
```

and
```ruby
Test << 'yyyy/mm/dd HH:MM:ss' 
#or variations of it but we'll come across locale differences with mm/dd vs dd/mm, dd-mm, etc.
```

pacive:
The `<<` method is defined in `lib/openhab/core/dsl/monkey_patch/items/items.rb` and basically just calls `#to_s`on the passed object and the passes it to the openhab event bus that manages the command to the item. So basically:
```
Test << ZonedDateTime.now
```
Should work, since the string representation can be parsed by OH into a DateTimeType, but the others would not. You can see the formats supported in the [OH Core repo](https://github.com/openhab/openhab-core/blob/7907f487357f8fa731b18ca104720fa90bb0c84c/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java#L45-L57)

If the others are to work, I would have to override the `<<` and `update` methods. Should I do that?

lib/openhab/core/dsl/items/datetime_item.rb
```diff
@@ -0,0 +1,97 @@
...
+    module DSL
+      module Items
+        #
+        # Delegation to OpenHAB DateTime Item
+        #
+        class DateTimeItem
+          extend Forwardable
+          include Comparable
+
+          def_delegator :@datetime_item, :to_s
```
jimtng, line 19: I'd like to suggest making `state` available, i.e.:

```suggestion
          def_delegators :@datetime_item, :to_s, :state
```

It will be useful when doing comparisons from other objects. I had to make this change on NumberItem and StringItem in #135

@boc-tothefuture WDYT?

lib/openhab/core/dsl/items/datetime_item.rb
```diff
@@ -0,0 +1,97 @@
...
+
+          #
+          # Compare the Item's state to another Item or object that can be compared
+          #
+          # @param [Object] other Other objet to compare against
+          #
+          # @return [Integer] -1, 0 or 1 depending on the result of the comparison
+          #
+          def <=>(other)
+            return unless state?
```
jimtng, line 48: Just a question (I'm still learning). Where does this `state?` come from?

jimtng:
> The `<<` method is defined in `lib/openhab/core/dsl/monkey_patch/items/items.rb` and basically just calls `#to_s`on the passed object and the passes it to the openhab event bus that manages the command to the item. So basically:
> 
> ```
> Test << ZonedDateTime.now
> ```
> 
> Should work, since the string representation can be parsed by OH into a DateTimeType

That's great! Thanks for the explanation.



> the others would not. You can see the formats supported in the [OH Core repo](https://github.com/openhab/openhab-core/blob/7907f487357f8fa731b18ca104720fa90bb0c84c/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java#L45-L57)

OK, so in my suggestion with string, it can be done provided that it adheres to OH Core's formatting. That's fine by me.



> If the others are to work, I would have to override the `<<` and `update` methods. Should I do that?

No, I'd say leave it for now. I wasn't even sure if we should support `Test << TimeOfDay.now`. I'm happy to drop it.

jimtng:
One thing missing, if I may suggest it, is the documentation.

lib/openhab/core/dsl/items/datetime_item.rb
pacive: `state` is already delegated to the item via `method_missing`, the reason `to_s` need to be explicitly delegated is because all ruby objects have that method (since they inherit from Object), so `method_missing` never gets called, and we need to make sure it's the Item's `toString()` method that gets called instead (jruby handles the name translation).

lib/openhab/core/dsl/items/datetime_item.rb
pacive: `lib/openhab/core/dsl/monkey_patch/items/items.rb` monkeypatches the openhab Item classes with some additional methods that can be used. The reason it works here is because it gets delegated to the DateTimeItem via `method_missing`

pacive:
> One thing missing, if I may suggest it, is the documentation.

Of course, that needs to be updated as well. On it!

jimtng:
> * DateTime Items can be updated and commanded with ruby Time objects:
>   ```ruby
>   Test << Time.now
>   ```

Perhaps this need to be added to the cucumber test.


Will this work?

> ```ruby
> Test << 600 + Test
> Test << '00:05' - Test
> Test << 5.minutes + Test
> ```

It might be do-able by overriding DateTimeItem.coerce


Will this work?
```ruby
Test << DateTimeType.new(xxxx) #or
Test << DateTime.new(xxx)

#currently event.state is not wrapped into our type, so it's Java::OrgOpenhabCoreLibraryTypes::DateTimeType
rule 'test' do
  changed DateTimeItem1
  run do |event|
    Test << event.state
  end
end

#currently event.item is not wrapped into our type, so it's Java::OrgOpenhabCoreLibraryItems::DateTimeType
rule 'test' do
  changed DateTimeItem1
  run do |event|
    Test << event.item
  end
end

```

lib/openhab/core/dsl/items/datetime_item.rb
jimtng: > `state` is already delegated to the item via `method_missing`

You're correct. I re-tested my own changes without delegating `state` and they still work. I've updated my own PR. Thanks! Feel free to mark this as resolved.

pacive:
> > * DateTime Items can be updated and commanded with ruby Time objects:
> >   ```ruby
> >   Test << Time.now
> >   ```
> 
> Perhaps this need to be added to the cucumber test.
> 
Good catch, will do

> Will this work?
> 
> > ```ruby
> > Test << 600 + Test
> > Test << '00:05' - Test
> > Test << 5.minutes + Test
> > ```
> 
> It might be do-able by overriding DateTimeItem.coerce
> 
Not sure how coerce works, but I think only the ruby Numeric classes use it. String e.g. doesn't have a `#-`-method, so wouldn't work, and I don't think the java Duration class returned by `#minutes` can use it either. So it would only work in the first case, which would confuse users as to why it works both ways sometimes but not others.

> Will this work?
> 
> ```ruby
> Test << DateTimeType.new(xxxx) #or
> Test << DateTime.new(xxx)
> 
> #currently event.state is not wrapped into our type, so it's Java::OrgOpenhabCoreLibraryTypes::DateTimeType
> rule 'test' do
>   changed DateTimeItem1
>   run do |event|
>     Test << event.state
>     Test << event.state + 1.hour
>   end
> end
> 
> #currently event.item is not wrapped into our type, so it's Java::OrgOpenhabCoreLibraryItems::DateTimeType
> rule 'test' do
>   changed DateTimeItem1
>   run do |event|
>     Test << event.item
>     Test << event.item + 1.hour
>   end
> end
> ```
The first command in the rules should work, since it's the string representation that gets sent as the command by `<<` and the DateTimeType's `to_s` method returns a date string that can be converted back. In the other cases the event object would need to be intercepted and the Items wrapped before being passed to the rule. Otherwise doing this should work: 
```ruby
Test << DateTimeType.new(xxxx) #or
Test << DateTime.new(xxx)

#currently event.state is not wrapped into our type, so it's Java::OrgOpenhabCoreLibraryTypes::DateTimeType
rule 'test' do
  changed DateTimeItem1
  run do |event|
    Test << DateTime.new(event.state)
    Test << DateTime.new(event.state) + 1.hour
  end
end


#currently event.item is not wrapped into our type, so it's Java::OrgOpenhabCoreLibraryItems::DateTimeType
rule 'test' do
  changed DateTimeItem1
  run do |event|
    Test << DateTime.new(event.item)
    Test << DateTime.new(event.item) + 1.hour
  end
end
```
There might however be issues due to a name collision. The TimeOfDay class currently `require 'date'` which includes a DateTime class from ruby stdlib. I don't however see it being used anywhere, and tests run fine even with it removed. @boc-tothefuture is that import needed or is it just a remnant from a previous version of the class?

lib/openhab/core/dsl/items/datetime_item.rb
boc-tothefuture: Ok, so it is really doing ordered multi-delegation?  Delegates to A, if A doesn't have the method, delegates to B, etc?

pacive: Exactly, first it checks if the real OpenHAB item has the method, otherwise it delegates to a DateTime object that wraps the state, so it can be accessed without having to write `Item.state` everywhere.

boc-tothefuture:
> 
> There might however be issues due to a name collision. The TimeOfDay class currently `require 'date'` which includes a DateTime class from ruby stdlib. I don't however see it being used anywhere, and tests run fine even with it removed. @boc-tothefuture is that import needed or is it just a remnant from a previous version of the class?

You can remove it, I a guessing that it was a mistake.

This is looking really good BTW..

I tried playing around with a multi-delegator..  But it is more complicated than I would like so we won't wait for that.

pacive:
If all checks pass I think I'm done with this PR, unless you notice something that's missing?

pacive:
I believe the 'every' cucumber test might sometimes fail if the rule is created very close to when the minute changes, and there's a slight delay before the rule gets triggered. Should be able to fix it by changing the timeout to 61 seconds to have a small buffer.

jimtng:
> notice something that's missing?

docs :)

lib/openhab/core/dsl/types/datetime.rb
```diff
@@ -0,0 +1,342 @@
...
+
+          #
+          # The ZonedDateTime representing the state
+          #
+          # @return [Java::java::time::ZonedDateTime] ZonedDateTime representing the state
+          #
+          def zoned_date_time
+            @datetime.zonedDateTime
+          end
+
```
jimtng, line 179:
```suggestion
          end
          alias to_zdt zoned_date_time
```

pacive: Can be nice to have!

pacive:
Just added another nifty feature to make it possible to create TimeOfDay ranges directly from DateTimeItems:
```ruby
if between(DateOne...DateTwo).cover? '05:00'
  logger.info('05:00 is between DateOne..DateTwo')
end
```
Just going to make sure it passes the tests and then push it

jimtng:
> Just added another nifty feature to make it possible to create TimeOfDay ranges directly from DateTimeItems:
> 
> ```ruby
> if between(DateOne...DateTwo).cover? '05:00'
>   logger.info('05:00 is between DateOne..DateTwo')
> end
> ```
> 
> Just going to make sure it passes the tests and then push it

```ruby
TimeOfDay.now.between? DateOne..DateTwo
```
Sorry I keep throwing more things in 😄

pacive:
> > Just added another nifty feature to make it possible to create TimeOfDay ranges directly from DateTimeItems:
> > ```ruby
> > if between(DateOne...DateTwo).cover? '05:00'
> >   logger.info('05:00 is between DateOne..DateTwo')
> > end
> > ```
> > 
> > 
> > Just going to make sure it passes the tests and then push it
> 
> ```ruby
> TimeOfDay.now.between? DateOne..DateTwo
> ```
> 
> Sorry I keep throwing more things in 😄

That should work as well, since the `#between?` method on TimeOfDay calls the global `between` method.

boc-tothefuture:
Is this ready for review/merge or does it still have ongoing work?

pacive:
> Is this ready for review/merge or does it still have ongoing work?

I believe it's ready now!

pacive:
There's a conflict with @jimtng's recent commit, but it's an easy one. @boc-tothefuture do you fix it when you merge or should I rebase?

boc-tothefuture:
if you could rebase that would be great.  Now that it is finalized, I will take a look.  Thanks for the PR!

docs/usage/items/datetime.md
```diff
@@ -0,0 +1,55 @@
...
+
+Example_DateTimeItem - Example_DateTimeItem2 # Calculate the time difference, in seconds
+Example_DateTimeItem - '2021-01-01 15:40' # Calculates time difference
+```
+
+Comparisons between different time objects can be performed
+
+```ruby
+Example_DateTimeItem == Example_DateTimeItem2 # Equality, works across time zones
+Example_DateTimeItem > '2021-01-31' # After midnight jan 31st 2021
```
boc-tothefuture, line 39: Does this work across different locales?

lib/openhab/dsl/monkey_patch/ruby/time.rb
```diff
@@ -0,0 +1,34 @@
...
+        #
+        module TimeExtensions
+          include OpenHAB::Core
+
+          #
+          # Convert to ISO 8601 format
+          #
+          # @return [Java::JavaTime::Duration] Duration with number of units from self
+          #
+          def inspect
```
boc-tothefuture, line 18: I think this should be to_s and not inspect.

boc-tothefuture: Inspect should be deeper than just the openhab time representation.

lib/openhab/dsl/types/datetime.rb
```diff
@@ -0,0 +1,339 @@
...
+        include OpenHAB::Log
+
+        def_delegator :datetime, :to_s
+        def_delegator :zoned_date_time, :month_value, :month
+        def_delegator :zoned_date_time, :minute, :min
+        def_delegator :zoned_date_time, :second, :sec
+        def_delegator :zoned_date_time, :nano, :nsec
+        alias inspect to_s
+
+        java_import org.openhab.core.library.types.DateTimeType
```
boc-tothefuture, line 30: Can you switch this to the format of he other imports so it is consistent? 
Java::OrgOpenhabCore....

lib/openhab/dsl/types/datetime.rb
```diff
@@ -0,0 +1,339 @@
...
+
+        #
+        # Adds another object to this DateTime
+        #
+        # @param [Object] other Object to add to this. Can be a Numeric, another DateTime/Time/DateTimeType, a
+        #   Duration or a String that can be parsed into a DateTimeType or Time object
+        #
+        # @return [DateTime] A new DateTime object representing the result of the calculation
+        #
+        # rubocop: disable Metrics/MethodLength
```
boc-tothefuture, line 88: remove this disable please.

lib/openhab/dsl/types/datetime.rb
```diff
@@ -0,0 +1,339 @@
...
+        # Disabled line length - case dispatch pattern
+        def +(other)
+          logger.trace("Adding #{other} (#{other.class}) to #{self}")
+          case other
+          when Numeric
```
boc-tothefuture, line 93: convert to the one line form of the when statement.

when Number then ...
when DateTime, time then...

Then you don't need to disable the rubocop check above.

lib/openhab/dsl/types/datetime.rb
```diff
@@ -0,0 +1,339 @@
...
+        #
+        # @param [Object] other Object to subtract fom this. Can be a Numeric, another DateTime/Time/DateTimeType, a
+        #   Duration or a String that can be parsed into a DateTimeType or Time object
+        #
+        # @return [DateTime, Float] A new DateTime object representing the result of the calculation, or a Float
+        #   representing the time difference in seconds if the subtraction is between two time objects
+        #
+        def -(other)
+          logger.trace("Subtracting #{other} (#{other.class}) from self")
+          case other
```
boc-tothefuture, line 115: same here.. one line check and probably need to break them out into methods in the cases.

boc-tothefuture: Can + and - share the same private methods with just the operator passed in?

lib/openhab/dsl/types/datetime.rb
```diff
@@ -0,0 +1,339 @@
...
+
+        alias to_tod to_time_of_day
+
+        #
+        # Returns the value of time as an integer number of seconds since the Epoch
+        #
+        # @return [Integer] Number of seconds since the Epoch
+        #
+        def to_i
+          zoned_date_time.to_epoch_second
```
boc-tothefuture, line 155: should this be a def_delegator rather than a method?

lib/openhab/dsl/types/datetime.rb
```diff
@@ -0,0 +1,339 @@
...
+        end
+
+        #
+        # Converts a Numeric into a DateTimeType
+        #
+        # @param [Numeric] numeric A Integer or Float representing the number of seconds since the epoch
+        #
+        # @return [Java::org::openhab::core::library::types::DateTimeType] Object representing the same time
+        #
+        def datetime_type_from_numeric(numeric)
```
boc-tothefuture, line 270: is this coerce?  (asking, not sure).

lib/openhab/dsl/types/datetime.rb
```diff
@@ -0,0 +1,339 @@
...
+
+        #
+        # Converts other objects to a DateTimeType
+        #
+        # @param [String, Numeric, Time] datetime an object that can be parsed or converted into
+        #   a DateTimeType
+        #
+        # @return [Java::org::openhab::core::library::types::DateTimeType] Object representing the same time
+        #
+        def datetime_type_from(datetime)
```
boc-tothefuture, line 250: Should these be static methods?

lib/openhab/dsl/types/datetime.rb
```diff
@@ -0,0 +1,339 @@
...
+        end
+
+        #
+        # Converts a ruby Time object to an OpenHAB DateTimeType
+        #
+        # @param [Time] time The Time object to be converted
+        #
+        # @return [Java::org::openhab::core::library::types::DateTimeType] Object representing the same time
+        #
+        def datetime_type_from_time(time)
```
boc-tothefuture, line 288: static instead?

lib/openhab/dsl/types/datetime.rb
```diff
@@ -0,0 +1,339 @@
...
+
+        #
+        # Parses a string representing a time into an OpenHAB DateTimeType. First tries to parse it
+        #   using the DateTimeType's parser, then falls back to the ruby Time.parse
+        #
+        # @param [String] time_string The string to be parsed
+        #
+        # @return [Java::org::openhab::core::library::types::DateTimeType] Object representing the same time
+        #
+        def parse(time_string)
```
boc-tothefuture, line 302: same question on static

boc-tothefuture:
Ton of awesome work in here @pacive !  

I had a few questions/suggestions.. let me know what you think.

pacive:
Thanks for the review, will get right on it!

docs/usage/items/datetime.md
pacive: It uses the OH DateTimeType parser, which by uses the system default timezone. But it might be a good idea to move some of the logic to ruby so it can be set to the same timezone as the first operand.

lib/openhab/dsl/monkey_patch/ruby/time.rb
pacive: Had the same thought, will change to only override to_s an leave inspect to the default.

lib/openhab/dsl/types/datetime.rb
pacive: In ruby Time they work somewhat differently: + only accepts Numerics and always return a new Time object. - Does the same for Numerics, but also accepts another Time object and returns the time difference as a Float. I was trying to mimic that as much as possible, but with som extended functionality. That's why I had do write them as separate methods.

lib/openhab/dsl/types/datetime.rb
pacive: This is used by the initialize method to create a new DateTimeType from a numeric passed to initialize. See below for further discussion.

lib/openhab/dsl/types/datetime.rb
pacive: As above, these methods are used by the initialize method to construct a DateTimeType from different objects passed to initialize. But I might be able to make them static and call them directly from the <=>, + and - methods which are the only ones that use this I believe.

lib/openhab/dsl/types/datetime.rb
boc-tothefuture: ok, no need to overcomplicate it then..

lib/openhab/dsl/types/datetime.rb
boc-tothefuture: Ok..  I thought they had external usage.

Should they all be private then?

pacive: I have changed them to static methods, and it feels better actually :)

docs/usage/items/datetime.md
pacive: I rewrote the logic for this so it uses the timezone from the first operand. Should work more as expected now.

boc-tothefuture:
This ready?

pacive:
> This ready?

Unless you have any other questions i feel it is!

PR #135, opened 2021-02-07
feat: support comparisons between numbers, NumberItem, DimmerItem, String, DecimalType, PercentType, Quantity and QuantityType

### LINKED ISSUE #30: DecimalType comparison against RubyFixNum
I noticed that currently DecimalType has been extended with == operator to compare against Integer.

It would be nice to implement a similar solution for other comparison operators <, <=, >, >=. It would also be great to be able to compare against floating point.

jimtng:
Resolve #30 

support comparisons between numbers, NumberItem, StringItem, DecimalType, PercentType and QuantityType

The tests are done in code instead of using cucumber scenarios. This makes the tests run in an instant instead of hours.


boc-tothefuture:
Awesome.. let me look at the testing, there is probably a way to accomplish this (at roughly the same speed) without having one large test.  

We would just need to have a different background (that doesn't reset OpenHAB every time).  Let me see what I can do here.

boc-tothefuture:
FYI - not going to hold up merging this for the test change I am talking about..  I will open an issue after merge to look at it.

But I do want to get #130 merged first.

jimtng:
rebased to the main branch

lib/openhab/core/dsl/monkey_patch/ruby/string.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module OpenHAB
+  module Core
+    module DSL
+      module MonkeyPatch
+        module Ruby
+          #
+          # Extend integer to create duration object
```
boc-tothefuture, line 9: Can you fix the comment?

lib/openhab/core/dsl/monkey_patch/types/decimal_type.rb
```diff
@@ -10,15 +10,51 @@ class Java::OrgOpenhabCoreLibraryTypes::DecimalType
...
   #
-  # @return [Boolean] True if have the same BigDecimal representation, false otherwise
+  # @return [Integer] -1,0,1 or nil depending on value supplied, nil comparison to supplied object is not possible.
   #
-  def ==(other)
-    return equals(other) unless other.is_a? Integer
+  def <=>(other)
+    logger.trace("#{self.class} #{self} <=> #{other} (#{other.class})")
+    case other
+    when Integer, Float
```
boc-tothefuture, line 22: why not Numeric instead of these specific types?

lib/openhab/core/dsl/monkey_patch/types/decimal_type.rb
```diff
@@ -10,15 +10,51 @@ class Java::OrgOpenhabCoreLibraryTypes::DecimalType
...
+      to_big_decimal.compare_to(Java::JavaMath::BigDecimal.new(other))
+    when Java::OrgOpenhabCoreTypes::UnDefType
+      1
```
boc-tothefuture, line 25: I don't know the answer here.. is 1 right right answer or is nil here?

lib/openhab/core/dsl/monkey_patch/types/decimal_type.rb
```diff
@@ -10,15 +10,51 @@ class Java::OrgOpenhabCoreLibraryTypes::DecimalType
...
+  # Coerce objects into a DecimalType
+  #
+  # @param [Object] other object to coerce to a DecimalType if possible
+  #
+  # @return [Object] Numeric when applicable
+  #
+  def coerce(other)
+    logger.trace("Coercing #{self} as a request from  #{other.class}")
+    case other
+    when Integer, Float
```
boc-tothefuture, line 42: Same question on Numeric vs the concrete types.

lib/openhab/core/dsl/monkey_patch/ruby/string.rb
jimtng: Oops, I've been caught copy pasting!

lib/openhab/core/dsl/monkey_patch/types/decimal_type.rb
jimtng: Ruby Numeric includes complex number but I suppose it won't hurt. However I need to convert the 'other' to string to pass onto Java BigDecimal's constructor.

lib/openhab/core/dsl/monkey_patch/types/decimal_type.rb
jimtng: If it returns nil, the equality comparison will raise an error because nil class doesn't have <=> operator. I picked 1 so anything is greater than undef/null.

lib/openhab/core/dsl/monkey_patch/types/decimal_type.rb
jimtng: see above.

lib/openhab/core/dsl/monkey_patch/types/decimal_type.rb
jimtng: According to https://www.rubyguides.com/2016/07/numbers-in-ruby/ Numeric has several subclasses. Out of those, Java's BigDecimal.new(String) will fail on Rational (because I'm guessing that it won't know "1/2" as a string) and Complex. It would only succeed on Integer, Float and BigDecimal, although I haven't tested this.

Do you think we should stick to Numeric or list the three supported types specifically?

jimtng: Changed to Numeric

PR #147, opened 2021-02-10
feat(units): import OpenHAB common units for UoM

jimtng:
I noticed that the common units from OpenHAB core isn't imported. This PR will allow the use of various units available without having to import them in the user scripts
* WATT
* AMPERE
* VOLT
* ONE
* PERCENT
* NEWTON
* HERTZ
* SECOND
* MINUTE
* HOUR
* DAY
* WEEK
* YEAR
* LITRE
* etc

With this, user scripts can now refer to, e.g.  `Units::VOLT` without having to add `java_import org.openhab.core.library.unit.Units` at the top of the scripts.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    openhab-scripting (2.16.2)
+    openhab-scripting (2.16.3)
       bundler (~> 2.2)
 
 GEM
--- a/docs/usage/items/number.md
+++ b/docs/usage/items/number.md
@@ -8,23 +8,23 @@ grand_parent: Usage
 ---
 
 
-#### Number Item
+# Number Item
 
 | Method          | Parameters | Description                                                                                                                                              | Example                                                                      |
 | --------------- | ---------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------- |
 | truthy?         |            | Item state not UNDEF, not NULL and is not Zero                                                                                                           | `puts "#{item.name} is truthy" if item.truthy?`                              |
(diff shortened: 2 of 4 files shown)


boc-tothefuture:
Do we need to document this somewhere?

jimtng:
> Do we need to document this somewhere?

I'm glad you asked. I believe we should move the UoM / Quantity section that's currently inside number.md into a separate section inside Misc. It would make it easier to find.

boc-tothefuture:
Go ahead and reorganize it. I think we want to link form number to the quantity stuff since it is related.

jimtng:
> I believe we should move the UoM / Quantity section that's currently inside number.md into a separate section inside Misc.

Come to think of it, Quantity and UoM are mainly used in conjunction with NumberItems, and there is a lot of overlap in the examples of Quantity and NumberItem which makes sense to keep them together.

On the other hand, I think the use of unit classes such as Units::VOLT would be rare. I would just rather use the string representation of the Unit. Besides, I don't think we should keep a list of all the units in our documentation as it could get out of sync with the openhab core, so perhaps just leave it undocumented? I also noticed that the Units class isn't a comprehensive list of available units. For example, there needs to be the metric multipliers like Kilo, Deca, etc which are defined yet elsewhere. In short, using the string unit is easier.

I did fix a few typos in number.md

I also added NumberItem to NumberItem in comparisons.feature - should I submit that as a separate PR?

boc-tothefuture:
I don't think we need to list all them out.  Just a link on the docs to the openhab UoM page saying that all of these are available as a string and providing one example?

jimtng:
> Just a link on the docs to the openhab UoM page saying that all of these are available as a string and providing one example?

That's already covered in https://boc-tothefuture.github.io/openhab-jruby/usage/items/number/

PR #148, opened 2021-02-10
fix(items): items[itemname] didn't wrap the object

jimtng:


jimtng:
Will combine this with #149

PR #149, opened 2021-02-10
fix: decorate items[itemname], event.item, and triggered item

### LINKED ISSUE #58: Convert all items to delegate rather than monkeypatching
Discussed in #47 and #48 
Also need to wrap when coming back from framework on rule triggers.

jimtng:
This task is mentioned in #58

### DIFF
--- a/lib/openhab/core/dsl/rule/automation_rule.rb
+++ b/lib/openhab/core/dsl/rule/automation_rule.rb
@@ -221,20 +221,18 @@ module OpenHAB
           end
 
           #
-          # Patch event to include event.item when it doesn't exist
-          # This is to patch a bug see https://github.com/boc-tothefuture/openhab-jruby/issues/75
-          # It may be fixed in the openhab core in the future, in which case, this patch will no longer be necessary
+          # Patch event to decorate event.item with our item wrapper
           #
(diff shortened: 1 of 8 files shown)


lib/openhab/core/dsl/rule/automation_rule.rb
```diff
@@ -237,6 +237,19 @@ class << event
...
 
+          #
+          # Patch event to decorate event.item with our item wrapper
+          #
+          # @param [OpenHAB Event] event patch
+          #
+          def decorate_event_item(event)
+            class << event
+              def item
+                EntityLookup.decorate_items(super).first
```
pacive, line 248: Would it make sense to split the `EntityLookup.decorate_items` method into two methods: `::decorate_items` that take an array of items, and `::decorate_item` that takes a single item (and is called by `::decorate_items` for each item)? Then you could call the second method directly here.

jimtng: Good idea!

jimtng:
I modified the way event.item is wrapped by calling items[item_name]. However, this depends on the fix in #148 which should be merged first. Also this now eliminates the need for #87

PR #151, opened 2021-02-10
docs: document UI created rules

### LINKED ISSUE #13: documentation needs to clarity between rule processing and rule execution phase.
See [this](https://community.openhab.org/t/jruby-openhab-rules-system/110598/38?u=broconne) for more details.

pacive:
Just a quick write-up on how to create rules in the UI, to avoid confusion from users taking that path. Note that one of the examples requires that #134 is merged, so might want to hold off this until then. Feel free to add more examples!

### DIFF
--- a/docs/usage/ui-rules.md
+++ b/docs/usage/ui-rules.md
@@ -1,0 +1,58 @@
+---
+layout: default
+title: Creating rules in the UI
+nav_order: 2
+has_children: false
+parent: Usage
+---
+
+## Creating rules in Main UI ##
+
+Rules can be created in the UI as well as in rules files, but some things are a bit different.
+First of all only the execution blocks need to be created in the script. All triggers and conditions
+are created directly in the UI instead.
+
+To create a rule:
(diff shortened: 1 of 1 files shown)


boc-tothefuture:
Awesome!  Thanks for the doc. 
I don't think this actually closes #13 however. 

#13 needs to describe the fact that there are really "two passes" in this system.  There is a pass that creates the rules when the rules file is read by openhab.  Then later the rules get executed by the system and they are distinct.  So anything outside of rule block gets executed when the file is parsed by OpenHAB.

It is somewhat analogous to the [chef two pass model](https://coderanger.net/two-pass/)

pacive:
Ah, can remove the link to the issue so it won't be closed. Do you think that should be explained in more detail here or elsewhere in the docs? Should perhaps go in the usage/index.md?

boc-tothefuture:
I think it should go in usage/index.md.  

This is a welcome addition as it is for those using the UI..  Thanks for the PR!

PR #152, opened 2021-02-10
feat: add RollershutterItem

### LINKED ISSUE #39: Add RollerShutter Item Type

pacive:
This was actually a rather simple one since it basically just holds a number and can receive various commands, but let me know if I missed something!

Closes #39 


jimtng:
Great work! 
Considering that SwitchItem's commands are on, off, toggle, should this one be up, down, move instead of up!, down! and move! ?

pacive:
> Great work!
> Considering that SwitchItem's commands are on, off, toggle, should this one be up, down, move instead of up!, down! and move! ?

I felt that ruby's convention of using bang-methods to show that you're manipulating the object would be a good way to signal that the real-world device connected to OH will be affected, but can change to be consistent.

Also now remembered that I forgot to write the docs, that will be a task for tomorrow.

jimtng:
I updated my post above to add more requests :)

> I felt that ruby's convention of using bang-methods to show that you're manipulating the object would be a good way to signal that the real-world device connected to OH will be affected

I see your point and somewhat agree with it. Perhaps we should change, or at least alias the SwitchItem's method to on!, off! toggle! to be consistent?

jimtng:
Also, just wondering - so feel free to reject this idea:
* `RollerItem << '50%'`
* `if RollerItem > '20%'`

boc-tothefuture:
The ! mark has a second indication of 'danger' in ruby, that sometimes goes beyond mutation.  

Technically, we aren't changing the object state. We are sending a command that may eventually change the object the state. 

You could argue that that on! and off! should map to update of on! and off! however. For the ones that send a command I would probably leave off the exclamation point.  However, I am open to discussion.

lib/openhab/core/dsl/monkey_patch/ruby/number.rb
```diff
@@ -26,6 +26,15 @@ module IntegerExtensions
             alias ms millis
             alias minute minutes
             alias hour hours
+
+            #
+            # Create an OpenHAB PercentType
+            #
+            # @return [Java::OrgOpenhabCoreLibraryTypes::PercentType] PercentType corresponding to the value
+            #
+            def percent
```
boc-tothefuture, line 35: Do we want to define % as well? 

Not sure we do, but what do you think?

pacive: Wouldn't that mess with the modulus operator?

pacive:
> The ! mark has a second indication of 'danger' in ruby, that sometimes goes beyond mutation.
> 
> Technically, we aren't changing the object state. We are sending a command that may eventually change the object the state.
> 
> You could argue that that on! and off! should map to update of on! and off! however. For the ones that send a command I would probably leave off the exclamation point. However, I am open to discussion.

I'm fine with leaving it out, will remove it!

pacive:
> * Can I do this? `Rolleritem << 50`, and `RollerItem << 50.percent` (or the equivalent .command(50))?
> * `RollerItem << DecimalType.new(50)` and `Item << NumberItem`

Will add some more tests to verify, but shouldn't be a problem!

lib/openhab/core/dsl/monkey_patch/ruby/number.rb
boc-tothefuture: Not sure I do either...  Something about 'percent' however feels off from the rest of the DSL. 
Should we think of this like a quantity type? 

50 | '%' 

?

pacive: Not sure, we would have to override the command method to translate the quantitytype to a percenttype before passing it to the eventbus, otherwise the command won't be accepted, even when converted to string.

pacive: Another solution would be to skip this entirely, since sending regular Integers as command works as well.

boc-tothefuture: Let's remove percent for now?  Since we aren't sure and I would rahter not remove it as a breaking change.

features/rollershutter_item.feature
```diff
@@ -0,0 +1,71 @@
...
+      | type          | name      | label        | group          |
+      | Rollershutter | RollerOne | Roller One   | Rollershutters |
+      | Rollershutter | RollerTwo | Roller Two   | Rollershutters |
+
+  Scenario: Sending command using PercentType work
+    Given item "RollerOne" state is changed to "<initial>"
+    And item "RollerTwo" state is changed to "0"
+    And code in a rules file
+      """
+      java_import Java::OrgOpenhabCoreLibraryTypes::DecimalType
```
boc-tothefuture, line 17: Should we be testing / supporting this?  Trying to figure out why the user would want to create a decimal type in here and ever use it?

lib/openhab/dsl/items/rollershutter_item.rb
```diff
@@ -0,0 +1,140 @@
...
+require 'forwardable'
+require 'java'
+
+module OpenHAB
+  module DSL
+    module Items
+      #
+      # Delegator to OpenHAB Rollershutter Item
+      #
+      class RollershutterItem
```
boc-tothefuture, line 12: This should extend numeric?

lib/openhab/dsl/items/rollershutter_item.rb
```diff
@@ -0,0 +1,140 @@
...
+
+        #
+        # Returns the rollershutter's position
+        #
+        # @return [Java::OrgOpenhabCoreLibraryTypes::PercentType] the position of the rollershutter
+        #
+        def position
+          state.as(PercentType)
+        end
+
```
boc-tothefuture, line 59: This should support coerce too, right?
So you can do:
```RollerShutter << 10 + OtherShutter```

File: features/rollershutter_item.feature [hunk 5]
```diff
@@ -0,0 +1,71 @@
...
+    When I deploy the rules file
+    Then It should log "<result>" within 5 seconds
+      Examples:
+          | initial | method | result |
+          | 0       | up?    | true   |
+          | 0       | down?  | false  |
+          | 100     | up?    | false  |
+          | 100     | down?  | true   |
+          | 50      | up?    | false  |
+          | 50      | down?  | false  |
```
boc-tothefuture, line 71: Once it extends numeric we should ensure that case statements work..
```
when RollerShutter
case UP:
case Down:
case 50:
case 0..25: 
```

boc-tothefuture:
@pacive - 
  Great work!  This looks good..  Main question I have is around extending numeric and also probably need to monkeypath UP/Down type so it works with case statements like we did for the others.

features/rollershutter_item.feature
pacive: I added it because of a question from @jimtng, wouldn't expect it to be used in practice. It might be useful to pass the state from a regular Number Item though, but that conversion is perhaps handled already?

lib/openhab/dsl/items/rollershutter_item.rb
pacive: Not sure what the benefit would be? But wouldn't be a problem to add

features/rollershutter_item.feature
pacive: That's a great idea, would just need to ovverride `===` for that, right?

features/rollershutter_item.feature
boc-tothefuture: I think it should be handled automatically.. its ok to leave the test.

lib/openhab/dsl/items/rollershutter_item.rb
boc-tothefuture: See below, it is what should make it automatically work in a case statement.

features/rollershutter_item.feature
boc-tothefuture: Need to make it implement numeric and coerce so that '0..25' or `50` work.   Need to override === on UP and DOWN for the other to work.   Look at the monkeypatch on on_off_type.rb.

features/rollershutter_item.feature
jimtng: All persistence functions return a decimaltype

lib/openhab/dsl/items/rollershutter_item.rb
pacive: Not sure how to implement this actually, since this just wraps the item, and I can't coerce a number into that. Would have to create a wrapper class for PercentType and coerce the number to that?

boc-tothefuture: Is that different than what we do in NumberItem?  I think on coere we just return a Quantity or even a BigDecimal?  Something that can then be added or subtracted, etc?

pacive: Haven't really got the grasps on coerce, but if I just need to return two numbers that can be added could I convert self to an Integer? That wouldn't be a problem

jimtng: I've seen this elsewhere in our library too - we seem to be treating PercentType as Integer, but it is actually a DecimalType. Should we treat it as a Float?

features/rollershutter_item.feature
pacive: Added a monkey-patch to UpDownType so it supports case equality.

lib/openhab/dsl/items/rollershutter_item.rb
```diff
@@ -0,0 +1,179 @@
...
+          when UpDownType then state.as(UpDownType) == other
+          end
+        end
+
+        #
+        # Coerce self into other to enable calculations
+        #
+        # @param [Numeric] other Other numeric to coerce into
+        #
+        # @return [Array<Numberic>] an array of other and self coerced into other's type
```
boc-tothefuture, line 83: Minor typo on 'Numeric'

boc-tothefuture:
Looks good to me.  Want to fix the minor typo and the merge conflict and then its ready.

PR #159, opened 2021-02-12
feat: add dig-method to top level metadata

### LINKED ISSUE #158: Metadata syntax - safe navigation to access the configuration
Currently when the metadata namespace doesn't exist, `Item.meta['nonexistent']` would return nil. Is there a way to do some sort of safe navigation to access the metadata config on a non existent metadata, i.e. (I know this syntax is wrong but it should demonstrate the intention) `Item.meta['nonexistent']&['namespace_config']`

Currently it can be done like this `Item.meta['nonexistent']['namespace_config'] unless Item.meta['nonexistent'].nil?` which is not very nice.

Maybe we can do it like this (which is yet to be implemented): `Item.meta['nonexistent']&.config['namespace_config']`

What's the Ruby way of doing this?

pacive:
Added a dig method to the top level metadata namespace, so you can include all levels in a single query and not worry about if one level is missing.

Fixes #158 

### DIFF
--- a/docs/usage/misc/metadata.md
+++ b/docs/usage/misc/metadata.md
@@ -13,13 +13,14 @@ Item metadata can be accessed through `Item.meta` using the hash syntax. The `It
 
 In addition to the Enumerable methods, the following methods are available for `Item.meta`:
 
-| Method     | Parameters               | Description                                | Example                          |
-| ---------- | ------------------------ | ------------------------------------------ | -------------------------------- |
-| clear      |                          | Deletes all metadata namespaces            | `Item1.meta.clear`               |
-| delete     | namespace                | Delete the given namespace                 | `Item1.meta.delete 'namespace1'` |
-| each       | namespace, value, config | Loops through all the item's namespaces    |                                  |
-| merge!     | **other                  | Merge a hash or other item's metadata      |                                  |
-| namespace? | namespace                | Returns true if the given namespace exists |                                  |
+| Method     | Parameters               | Description                                           | Example                                     |
+| ---------- | ------------------------ | ----------------------------------------------------- | ------------------------------------------- |
+| clear      |                          | Deletes all metadata namespaces                       | `Item1.meta.clear`                          |
+| delete     | namespace                | Delete the given namespace                            | `Item1.meta.delete 'namespace1'`            |
+| dig        | key, *keys               | Dig through the namespace to find the specified value | `Item1.meta.dig('namespace1', 'foo', 'bar') |
+| each       | namespace, value, config | Loops through all the item's namespaces               |                                             |
+| merge!     | **other                  | Merge a hash or other item's metadata                 |                                             |
+| namespace? | namespace                | Returns true if the given namespace exists            |                                             |
 
 Metadata configuration is a hash and can be accessed using a subscript of `Item.meta['namespace']`. For example, the following Item metadata
 
@@ -60,6 +61,11 @@ Item1.meta['namespace1'].value
 
 # Access namespace1's configuration
 Item1.meta['namespace1']['config1']
+
+# Safely search for the specified value - no errors are raised, only nil returned if a key
+# in the chain doesn't exist
+Item1.meta.dig('namespace1', 'config1') #=> 'foo'
+Item1.meta.dig('namespace2', 'config1') #=> nil
 
 # Set item's metadata value, preserving its config
 # Item1's metadata before: { namespace1="value" [ config1="foo", config2="bar" ] }
--- a/lib/openhab/dsl/monkey_patch/items/metadata.rb
+++ b/lib/openhab/dsl/monkey_patch/items/metadata.rb
@@ -135,6 +135,18 @@ module OpenHAB
             end
 
             #
+            # Implements Hash#dig-like functionaity to metadata
+            #
+            # @param [String] key The first key
+            # @param [Array<String, Symbol>] keys More keys to dig deeper
+            #
+            # @return [OpenHAB::DSL::MonkeyPatch::Items::MetadataItem], or nil if the namespace doesn't exist
+            #
+            def dig(key, *keys)
+              keys.empty? ? self[key]&.value : self[key]&.dig(*keys)
+            end
+
+            #
             # Enumerates through all the namespaces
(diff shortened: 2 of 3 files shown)


boc-tothefuture:
@jimtng you wrote the original code, are you good with this?

The docs for metadata probably need to be updated too?

lib/openhab/dsl/monkey_patch/items/metadata.rb
```diff
@@ -134,6 +134,20 @@ def []=(namespace, value)
...
+            #
+            # Implements Hash#dig-like functionaity to metadata
+            #
+            # @param [String, Symbol] key The first key
+            # @param [Array<String, Symbol>] keys More keys to dig deeper
+            #
+            # @return [OpenHAB::DSL::MonkeyPatch::Items::MetadataItem], or nil if the namespace doesn't exist
+            #
+            def dig(key, *keys)
+              return self[key.to_s] if keys.empty?
```
jimtng, line 146: If we want to support key as symbol, perhaps implement it in the parent NamespaceAccessor::[] and remove key.to_s from here. However, to be consistent, we'd need to support it all the way into the configuration, and for that the MetadataItem needs to implement hash with indifference access, something that I didn't tackle in the original implementation.

If you could make the nested configuration hash support symbols too, that'd be great. Otherwise, for the purpose of this PR, perhaps remove the symbol support for now.

lib/openhab/dsl/monkey_patch/items/metadata.rb
```diff
@@ -134,6 +134,20 @@ def []=(namespace, value)
...
+
+              key?(key) ? self[key.to_s].dig(*keys) : nil
```
jimtng, line 148: Stylistic preference, I suppose? Your version makes it more obvious that it would return nil if key is not defined. Below is shorter. Which one is more common in Ruby?

```suggestion
              self[key.to_s].dig(*keys) if key?(key)
```

lib/openhab/dsl/monkey_patch/items/metadata.rb
jimtng: I am also wondering whether we should return the .value here or return the MetadataItem

pacive: I could drop the symbol, no problem

lib/openhab/dsl/monkey_patch/items/metadata.rb
pacive: I think your version looks more ruby actually :)

lib/openhab/dsl/monkey_patch/items/metadata.rb
pacive: > I am also wondering whether we should return the .value here or return the MetadataItem

Not sure either, returning the MetadataItem would be more consistent with the [] method.

jimtng: > Not sure either, returning the MetadataItem would be more consistent with the [] method.

If it returns the value, it would make dig always return a value wether it's the namespace's value or the config value. OTOH, it would be inconsistent because Item.meta.dig('namespace') should technically be the same as Item.meta['namespace'], right?

@boc-tothefuture WDYT?

jimtng: The more I think about it, the more I think that dig('namespace') should return the value, not the MetadataItem. If the user wants to get the MetadataItem, they could just use Item.meta['namespace'].

Another reason: consider this metadata definition in an item:
```
Switch Switch_Item { namespace1="ns_value" [ config1="conf1_value", config2="conf2_value" ] }
```

It seems clear to me that dig('namespace1') should return 'ns_value', whereas dig('namespace1', 'config1') should return 'conf1_value'.

In any case I don't think it matters that much because the main reason for using dig would be to get into the configuration values. However, we do need to make a decision here.

So my vote, is for dig('namespace') to return the value, i.e.
```ruby
            def dig(key, *keys)
              return self[key]&.value if keys.empty?
```

pacive: I can agree with that, will change so it returns the value

lib/openhab/dsl/monkey_patch/items/metadata.rb
```diff
@@ -134,6 +134,20 @@ def []=(namespace, value)
...
+            #
+            # @param [String] key The first key
+            # @param [Array<String, Symbol>] keys More keys to dig deeper
+            #
+            # @return [OpenHAB::DSL::MonkeyPatch::Items::MetadataItem], or nil if the namespace doesn't exist
+            #
+            def dig(key, *keys)
+              return self[key]&.value if keys.empty?
+
+              self[key].dig(*keys) if key? key
```
jimtng, line 148:
```suggestion
              keys.empty? self[key]&.value : self[key]&.dig(*keys)
```

jimtng:
@pacive could you also add this to the docs please? Thank you

docs/usage/misc/metadata.md
```diff
@@ -13,13 +13,14 @@ Item metadata can be accessed through `Item.meta` using the hash syntax. The `It
...
+| Method     | Parameters               | Description                                           | Example                          |
+| ---------- | ------------------------ | ----------------------------------------------------- | -------------------------------- |
+| clear      |                          | Deletes all metadata namespaces                       | `Item1.meta.clear`               |
+| delete     | namespace                | Delete the given namespace                            | `Item1.meta.delete 'namespace1'` |
+| each       | namespace, value, config | Loops through all the item's namespaces               |                                  |
+| merge!     | **other                  | Merge a hash or other item's metadata                 |                                  |
+| namespace? | namespace                | Returns true if the given namespace exists            |                                  |
+| dig        | key, *keys               | Dig through the namespace to find the specified value | `Item1.meta.dig('namespace1', 'foo', 'bar')                                 |
```
jimtng, line 23: Thanks! I'd like to see it in alphabetical order for easy scanning through the list.

pacive:
@jimtng Does everything look ok now?

jimtng:
Yes LGTM

PR #170, opened 2021-02-16
build(Gemfile): Removed Gemfile.lock from source control

boc-tothefuture:
Now that we have moved to a Gem, I *think* we are supposed to remove Gemfile.lock from source control.

https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

Has some discussion on it.


boc-tothefuture:
@jimtng and @pacive can you take a look at the doc and let me know what you think?

pacive:
The scripting addon uses `gem install` to install the library, right? In that case I interpret the post you linked to as meaning we should remove it, since the dependencies installed by `gem` is not necessarily the same version as we have, so if one of them is updated and it breaks something we wouldn't notice, but it would break for the users.

boc-tothefuture:
@jimtng - any issue if I remove the Gemfile.lock?

jimtng:
> @jimtng - any issue if I remove the Gemfile.lock?

No issue with me. I don't fully understand the Gemfile.lock yet.

Issue #168, opened 2021-02-16
DRY Items


boc-tothefuture:
There is a lot of repeated code in the items and we should DRY them up a bit before.  These needs to be available as mixins so they can decorate any classes regardless of superclass hierarchy:

1. Multiple ordered and conditional delegation.  We need to delegate to multiple objects and sometimes only delegate if a condition is try (state exists for example).
2. Generate command methods based on Java Object Enums
3. Generate commands/checks based on java Object Enumbs

Potential syntax:

(1) 
item_delegator :@string_item
item_delegator:  {@string_item.state.to_full_string.to_s}, {@string_item.state?}

(2)
item_command StopMoveType

should generate:
```
def stop
def move
```

(3) 
item_command_state  UpDownType
should generate
```
def up
def up?
def down
def down?
```

boc-tothefuture:
I started doing number one on this list a while ago.. it is hard..

I will work on 2 and 3 which should be much easier.

pacive:
Since we're trying to move away from monkey-patching would it make sense to have a module with common patterns that can be included in each item class? (a base class is unfortunately not possible since some item classes need to subclass Numeric)
Something like
```ruby
module Item
  include Forwardable

  def_delegator :@item, :to_s

  def initialize(item)
    @item = item
  end

  def method_missing(method, *args, &block)
    return @item.__send__(method, *args, &block) if @item.respond_to(method)
    return @item.state.__send__(method, *args, &block) if @item.state.respond_to(method)
  end
```
Also the methods currently in `lib/openhab/dsl/monkey_patch/items/items.rb` would also be in this module.
Other item classes can then just include this and override if necessary.

pacive:
For 2 and 3 it should be possible to check the getAcceptedCommandTypes and getAcceptedDataTypes on the OH item instance to handle this automatically I think?

jimtng:
Most (all?) items can accept REFRESH (RefreshType) command too.

PR #171, opened 2021-02-16
refactor(items):  item_command and item_state for creating methods to…

boc-tothefuture:
… sending commands, checking state

Partially addressed #168 

Implemented against RollerShutter to act as an RFC.

### DIFF
--- a/lib/openhab/dsl/items/item_command.rb
+++ b/lib/openhab/dsl/items/item_command.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+
+require 'java'
+
+module OpenHAB
+  module DSL
+    module Items
+      #
+      # Holds methods to automatically generate commands and
+      # accessors for items
+      module ItemCommand
+        #
+        # For every value in the supplied enumeration create a corresponding method mapped to the lowercase
+        # string representation of the enum value For example, an enum with values of STOP and START
+        # would create methods stop() and start() that send the corresponding STOP and START commands to the item
+        #
+        # @param [Java::JavaLang::Enum] command_enum Enumeration to create commands for
+        #
+        #
+        def item_command(command_enum)
+          # rubocop:disable Style/HashEachMethods
+          # Disable rule because Java enum does not support each_value
+          command_enum.values.each do |command|
+            command_method = command.to_s.downcase
+            define_method(command_method) do
+              self.command(command)
+            end
+          end
+          # rubocop:enable Style/HashEachMethods
+        end
+
+        #
+        # For every value in the supplied enumeration create a corresponding method mapped to the lowercase
+        # string representation appended with a question mark '?' of the enum value For example,
+        # an enum with values of UP and DOWN would create methods up? and down? that check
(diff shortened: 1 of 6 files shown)


lib/openhab/dsl/items/item_command.rb
```diff
@@ -0,0 +1,61 @@
...
+        #
+        #
+        def item_state(command_enum)
+          # rubocop:disable Style/HashEachMethods
+          # Disable rule because Java enum does not support each_value
+          command_enum.values.each do |command|
+            status_method = "#{command.to_s.downcase}?"
+            define_method(status_method) do
+              if state?
+                current_state = state
```
pacive, line 50: Would it be possible to use `current_state = state.as(command_enum)`? Then we could have a method where you can pass all supported types as an array.

boc-tothefuture: Would that always work for all cases to do it as default? Right now I have the ability to pass in a block to do that, but if that isn't needed then it would be better.  I didn't realize we wanted to do that on all types since the others (besides rollershutter) seemed to just work.

pacive: Just thinking that if we should make things as generic as possible it would be easier. Or are there any possible cases where the block would contain anything else? And even if the state already is a e.g PercentType, it won't cause any problems.

pacive: The other item types we have only supports one type of state (except UndDefType), so they wouldn't have the same problem. But should be useable to add OnOffType to DimmerItems for example

boc-tothefuture: Looks like it does work everywhere..  code updated.

lib/openhab/dsl/items/item_command.rb
```diff
@@ -0,0 +1,57 @@
...
+        # @yield [state] Optional block that can be used to transform state prior to comparison
+        #
+        #
+        def item_state(command_enum)
+          # rubocop:disable Style/HashEachMethods
+          # Disable rule because Java enum does not support each_value
+          command_enum.values.each do |command|
+            status_method = "#{command.to_s.downcase}?"
+            define_method(status_method) do
+              state.as(command_enum) == command if state?
```
pacive, line 49: It probably shouldn't matter in practice, but should it return an explicit `false` instead of `nil` when there's no state? Will things like `!Item.on?` work when nil is returned?

```suggestion
              state? && state.as(command_enum) == command
```

pacive: Just tested, and `!nil` returns true, so that wouldn't be a problem at least.

jimtng: The problem will be when someone did this:
```
if Item.on? == false
```

pacive: And that is almost guaranteed to happen ;)

pacive (approved):
Great work! Will remove a lot of boilerplate code!

PR #178, opened 2021-02-18
feat(persistence): automatically convert to quantity for dimensioned items

jimtng:
This PR makes return value of the persistence functions converted to Quantity with the same unit as the attached NumberItem

Somewhat related to #153 which will apply more generically to any DecimalType data, whereas this PR applies specifically to persistence without having to specify the unit.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    openhab-scripting (2.19.2)
+    openhab-scripting (2.20.1)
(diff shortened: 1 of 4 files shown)


boc-tothefuture:
Is it possible to create a test for this?  It feels like the sort of thing that could get broken in a refactor.

jimtng:
> Is it possible to create a test for this? It feels like the sort of thing that could get broken in a refactor.

Test added

PR #179, opened 2021-02-18
refactor: automatically infer accepted types from OH item

pacive:
Added the ability to automatically detect accepted states/command from the OH item. Or is that too much?

Had to create a dummy item, because the `getAccepted*Types`-methods are instance methods. Or is there a way to read private static fields from jruby?

### DIFF
--- a/lib/openhab/dsl/items/rollershutter_item.rb
+++ b/lib/openhab/dsl/items/rollershutter_item.rb
@@ -19,13 +19,7 @@ module OpenHAB
 
         def_item_delegator :@rollershutter_item
 
-        java_import Java::OrgOpenhabCoreLibraryTypes::PercentType
-        java_import Java::OrgOpenhabCoreLibraryTypes::UpDownType
-        java_import Java::OrgOpenhabCoreLibraryTypes::StopMoveType
-
-        item_command Java::OrgOpenhabCoreLibraryTypes::StopMoveType
-        item_command Java::OrgOpenhabCoreLibraryTypes::UpDownType
-        item_state   Java::OrgOpenhabCoreLibraryTypes::UpDownType
+        item_type Java::OrgOpenhabCoreLibraryItems::RollershutterItem
 
         #
         # Creates a new RollershutterItem
--- a/lib/openhab/dsl/items/datetime_item.rb
+++ b/lib/openhab/dsl/items/datetime_item.rb
@@ -17,10 +17,12 @@ module OpenHAB
       class DateTimeItem
         extend Forwardable
         extend OpenHAB::DSL::Items::ItemDelegate
-
+        extend OpenHAB::DSL::Items::ItemCommand
         include Comparable
 
         def_item_delegator :@datetime_item
+
+        item_type Java::OrgOpenhabCoreLibraryItems::DateTimeItem
 
         #
         # Create a new DateTimeItem
--- a/lib/openhab/dsl/items/item_command.rb
+++ b/lib/openhab/dsl/items/item_command.rb
@@ -51,6 +51,21 @@ module OpenHAB
           end
           # rubocop:enable Style/HashEachMethods
         end
+
+        #
+        # Extract the accepted state and command types from the specified OpenHAB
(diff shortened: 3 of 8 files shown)


jimtng:
> Had to create a dummy item, because the `getAccepted*Types`-methods are instance methods.

I'm wondering if you could do this inside #initialize so you have access to an instance of @rollershutter_item

pacive:
> 
> I'm wondering if you could do this inside #initialize so you have access to an instance of @rollershutter_item

Then the method would be called every time an instance is created instead of when the class is created (i.e when the library is loaded) which is what I was hoping to avoid.

jimtng:
> Then the method would be called every time an instance is created

Good point - your current method is better.

boc-tothefuture:
> Had to create a dummy item, because the getAccepted*Types-methods are instance methods. Or is there a way to read private static fields from jruby?

You can read private static fields from Ruby.  Look at 'script_action_things.rb'

pacive:
> > Had to create a dummy item, because the getAccepted*Types-methods are instance methods. Or is there a way to read private static fields from jruby?
> 
> You can read private static fields from Ruby. Look at 'script_action_things.rb'

That worked, thanks. But that is dependant on that all Item types define these fields (and use the same name). The methods OTOH are defined in the Item interface, so all Item types are required to have them. Currently, all Item types have those fields, and I believe it's very unlikely that it would change, but still something one has to be aware of.

lib/openhab/dsl/items/rollershutter_item.rb
```diff
@@ -17,13 +17,7 @@ class RollershutterItem < Numeric
...
         def_delegator :@rollershutter_item, :to_s
 
-        java_import Java::OrgOpenhabCoreLibraryTypes::PercentType
-        java_import Java::OrgOpenhabCoreLibraryTypes::UpDownType
-        java_import Java::OrgOpenhabCoreLibraryTypes::StopMoveType
-
-        item_command Java::OrgOpenhabCoreLibraryTypes::StopMoveType
-        item_command Java::OrgOpenhabCoreLibraryTypes::UpDownType
-        item_state   Java::OrgOpenhabCoreLibraryTypes::UpDownType
+        oh_item_type Java::OrgOpenhabCoreLibraryItems::RollershutterItem
```
jimtng, line 20: Should we just call this `item_type` instead of `oh_item_type`?

boc-tothefuture:
Can you change this from a feat() to a refactor() as there is no external visible change? Also, can you apply to any other items that need it after we agree and merge in #183 ?

lib/openhab/dsl/items/rollershutter_item.rb
boc-tothefuture: I agree.. nothing else has oh_ in front of it. 

item_type doesn't really make it clear to me based on its name what it is doing either. 
to be in line with other ruby stuff they should all probably be renamed
def_item_command
def_item_state
def_item_from_types  (?maybe?)

pacive:
> Can you change this from a feat() to a refactor() as there is no external visible change? Also, can you apply to any other items that need it after we agree and merge in #183 ?

Absolutely!

lib/openhab/dsl/items/rollershutter_item.rb
jimtng: def_item_from_type ?

pacive: Was a bit unsure what to name it, but item_type feels best IMO. Am a bit busy this weekend, so won't have much time to make adjustments, but hopefully on Monday.

pacive:
There was some problems when adding this to the items under `monkey_patch`, probably because it also adds the `undef?` and `null?` methods, which in turn is called by the `state?` method, throwing it into some kind of endless loop. It does seem to work fine for the delegated item types though.

pacive:
Removed UnDefType from the list of types to add methods for, now it works for all items.

boc-tothefuture:
Awesome.. Great work.

PR #183, opened 2021-02-18
Item delegatation

boc-tothefuture:
Completes #168

### DIFF
--- a/lib/openhab/dsl/items/item_delegate.rb
+++ b/lib/openhab/dsl/items/item_delegate.rb
@@ -1,0 +1,91 @@
+# frozen_string_literal: true
+
+require 'java'
+require 'openhab/log/logger'
+
+module OpenHAB
+  module DSL
+    module Items
+      #
+      # Holds methods to delegate to items
+      #
+      module ItemDelegate
+        include OpenHAB::Log
+
+        #
+        # Delegates methods to the object returned from the supplied block if
+        #   they don't exist in the object this is included in.  If the supplied block returns nil, no delegation occurs
+        # If this item is called more than once delegation occurs in the order of invocation, i.e. the object returned
+        #   by the first block is delegated to if it responds to the missing method,
+        #   then the second block is processed, etc.
+        #
+        # @param [Proc] &delegate delgegate block
+        #
+        #
+        def item_delegate(&delegate)
+          @delegates ||= []
+          @delegates << delegate
+        end
+
+        #
+        # Delegate missing method calls to delegates supplied to item_delgate method
+        #  if no delegates exist or respond to missing method, super is invoked which will
+        #  throw the appropriate method missing error
+        #
+        # @param [String] meth misisng method
+        # @param [Array] *args Arguments to the missing method
+        # @param [Proc] &block supplied to the missing method
+        #
+        # @return [Object] Result of missing method invocation
+        #
+        def method_missing(meth, *args, &block)
+          logger.trace("Method (#{meth}) missing for item #{self.class}")
+          delegate = delegate_for(meth)
+          if delegate
+            logger.trace("Delegating #{meth} to #{delegate.class}")
+            delegate.__send__(meth, *args, &block)
+          else
+            super
+          end
+        end
+
+        #
+        # Checks if any of the supplied delgates respond to a specific method
+        #
+        # @param [String] meth method to check for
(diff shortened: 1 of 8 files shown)


boc-tothefuture:
Take a look at the item delegation here and let me know what you think.  Will use this as the basis for moving from monkeypatching items to delegating and fixing group, etc.

lib/openhab/dsl/items/item_delegate.rb
```diff
@@ -0,0 +1,84 @@
...
+        #
+        # Find a delegate for the supplied method
+        #
+        # @param [String] meth method to find delegate for
+        #
+        # @return [Boolean] True if any method responds to the supplied delegate, false otherwise
+        #
+        def delegate_for(meth)
+          (@delegates || []).each do |delegate_block|
+            delegate = delegate_block.call
```
pacive, line 77: Should we pass the method as an argument to the block? Then it would be possible to explicitly allow or deny certain methods:
```ruby
item_delegate { |method| @item if [:state, :to_s].include? method }
```
or
```ruby
item_delegate { |method| @item unless method == :to_s }
```
Can't think of a case where it would be useful right now, but it would make it more configurable.

boc-tothefuture: done.

lib/openhab/dsl/rules/triggers/changed.rb
```diff
@@ -113,6 +113,7 @@ def create_thing_changed_trigger(thing, from, to)
         #  second element is a Hash configuring trigger
         #
         def create_item_changed_trigger(item, from, to)
+          pp item
```
pacive, line 116: Is this supposed to be here?

pacive (commented):
Great work, will make things much easier going forward!

lib/openhab/dsl/rules/triggers/changed.rb
boc-tothefuture: nope.  thanks!

boc-tothefuture:
This look good @jimtng ?

jimtng:
LGTM

lib/openhab/dsl/items/item_delegate.rb
```diff
@@ -0,0 +1,91 @@
...
+        #
+        # Find a delegate for the supplied method
+        #
+        # @param [String] meth method to find delegate for
+        #
+        # @return [Boolean] True if any method responds to the supplied delegate, false otherwise
+        #
+        def delegate_for(meth)
+          (@delegates || []).each do |delegate_block|
+            delegate = delegate_block.call(meth)
```
jimtng, line 78: Could we do this in the constructor/initialize, and store the result of delegate_block.call inside @delegates array?

lib/openhab/dsl/items/item_delegate.rb
```diff
@@ -0,0 +1,91 @@
...
+            logger.trace("Checking if delegate (#{delegate.class}) responds to (#{meth})")
+            if delegate.respond_to? meth
```
jimtng, line 80: `delegate` could potentially be nil?

lib/openhab/dsl/items/item_delegate.rb
boc-tothefuture: I don't think we can because the object changes over time in some cases and if we 'cache' the object at construction time it won't work.  Things that work off the state for example.

lib/openhab/dsl/items/item_delegate.rb
boc-tothefuture: I think it's ok if it is nil as nil is still an object that has a class and a respond_to?  unless I am missing something else.

jimtng: It's to handle a case that will probably never be encountered. 
`item_delegate {}`
and the only bad side effect is it would log:
`Delegate (nil) found for method (#{meth})"`

It isn't going to cause a problem otherwise. I'm happy for it to be left as is, just wanted to point it out.

lib/openhab/dsl/items/item_delegate.rb
jimtng: I see, because it's a block, it could contain some logic which could change the returned object to be delegated to. In that case, the current implementation is fine.

PR #187, opened 2021-02-19
fix(persistence): selective conversion to Quantity

jimtng:
I've just realised that some Persistence functions aren't supposed to be converted to Quantity. For example, `evolution_rate`. The returned value is the percentage of change.

This PR will apply the Quantity selectively.

### DIFF
--- a/lib/openhab/dsl/monkey_patch/items/persistence.rb
+++ b/lib/openhab/dsl/monkey_patch/items/persistence.rb
@@ -8,7 +8,23 @@ module OpenHAB
         # Persistence extension for Items
         #
         module Persistence
-          %w[persist last_update].each do |method|
+          # All persistence methods that could return a Quantity
+          QUANTITY_METHODS = %i[average_since
+                                delta_since
+                                deviation_since
+                                sum_since
+                                variance_since].freeze
+
+          # All persistence methods that require a timestamp
+          PERSISTENCE_METHODS = (QUANTITY_METHODS +
+                                %i[changed_since
+                                   evolution_rate
+                                   historic_state
+                                   maximum_since
+                                   minimum_since
+                                   updated_since]).freeze
+
+          %i[persist last_update].each do |method|
             define_method(method) do |service = nil|
               service ||= persistence_service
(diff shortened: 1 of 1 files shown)


lib/openhab/dsl/monkey_patch/items/persistence.rb
```diff
@@ -46,12 +46,9 @@ def previous_state(service = nil, skip_equal: false)
...
             define_method(method) do |timestamp, service = nil|
               service ||= persistence_service
-              if timestamp.is_a? Java::JavaTimeTemporal::TemporalAmount
-                timestamp = Java::JavaTime::ZonedDateTime.now.minus(timestamp)
-              end
-              result = PersistenceExtensions.public_send(method, self, timestamp, service&.to_s)
-              if result.is_a?(Java::OrgOpenhabCoreLibraryTypes::DecimalType) && respond_to?(:unit) && unit
-                Quantity.new(Java::OrgOpenhabCoreLibraryTypes::QuantityType.new(result.to_big_decimal, unit))
+              result = PersistenceExtensions.public_send(method, self, to_zdt(timestamp), service&.to_s)
+              if %w[average_since delta_since deviation_since sum_since variance_since].include? method
```
pacive, line 50: Could you make this a constant instead?

jimtng: > Could you make this a constant instead?

Are you referring to %w? I've replaced them with %i.

lib/openhab/dsl/monkey_patch/items/persistence.rb
```diff
@@ -31,35 +47,48 @@ def previous_state(service = nil, skip_equal: false)
...
+          end
+
+          #
+          # Convert value to Quantity if it is a DecimalType and a unit is defined
+          #
+          # @param [Object] value The value to convert
+          #
+          # @return [Object] Quantity or the original value
+          #
+          def quantify(value)
```
boc-tothefuture, line 83: That is a better method name than we use elsewhere.. We should convert the others I will open an issue.

PR #188, opened 2021-02-19
feat: add stack trace to errors

### LINKED ISSUE #17: Log Script/JRuby errors
Right now they go to stdout.. investigate if they can go to the log instead.

pacive:
Added some rescue clauses to catch errors in scripts and log them along with stack traces. If the error is in the parsing of the file the error is re-raised to the ScriptEngineManager, but with the stack trace added. If the error is raised when the rule is executed it is handled internally and logged together with the stack trace. The errors are rescued for a single execution block, which allows other blocks in the same rule to run as usual.

In both cases the stack frames pertaining to internal classes/methods (script-library and jruby) are removed, so only the lines relevant to the end user are shown.

WDYT?

Fixes #17

### DIFF
--- a/lib/openhab/dsl/rules/automation_rule.rb
+++ b/lib/openhab/dsl/rules/automation_rule.rb
@@ -255,12 +255,27 @@ module OpenHAB
           event = inputs&.dig('event')
 
           while (task = run_queue.shift)
-            case task
-            when RuleConfig::Run then process_run_task(event, task)
-            when RuleConfig::Trigger then process_trigger_task(event, task)
-            when RuleConfig::Delay then process_delay_task(inputs, mod, run_queue, task)
-            when RuleConfig::Otherwise then process_otherwise_task(event, task)
+            if task.is_a? RuleConfig::Delay
+              process_delay_task(inputs, mod, run_queue, task)
+            else
+              process_task(event, task)
             end
+          end
+        rescue StandardError => e
+          print_backtrace(e)
+        end
+
+        #
+        # Dispatch execution block tasks to different methods
+        #
+        # @param [OpenHab Event] event that triggered the rule
+        # @param [Task] task task containing otherwise block to execute
+        #
+        def process_task(event, task)
+          case task
+          when RuleConfig::Run then process_run_task(event, task)
+          when RuleConfig::Trigger then process_trigger_task(event, task)
+          when RuleConfig::Otherwise then process_otherwise_task(event, task)
           end
         end
 
@@ -314,6 +329,16 @@ module OpenHAB
         def process_run_task(event, task)
           logger.trace { "Executing rule '#{name}' run block with event(#{event})" }
           task.block.call(event)
+        end
+
+        #
+        # Print error and stack trace without calls to internal classes
+        #
+        # @param [Exception] error A rescued error
+        #
+        def print_backtrace(error)
+          error = logger.clean_backtrace(error)
+          logger.error { "#{error.message} (#{error.class})\nIn rule: #{name}\n#{error.backtrace&.join("\n")}" }
         end
 
         #
--- a/lib/openhab/log/logger.rb
+++ b/lib/openhab/log/logger.rb
@@ -16,7 +16,14 @@ module OpenHAB
       java_import org.slf4j.LoggerFactory
 
       # @return [Array] Supported logging levels
-      LEVELS = %i[TRACE DEBUG WARN INFO ERROR].freeze
+      LEVELS = %i[trace debug warn info error].freeze
+      private_constant :LEVELS
+
+      #
+      # Regex for matching internal calls in a stack trace
+      #
+      INTERNAL_CALL_REGEX = %r{(openhab-scripting-.*/lib)|(org/jruby/)}.freeze
+      private_constant :INTERNAL_CALL_REGEX
 
       #
       # Create a new logger
@@ -32,11 +39,30 @@ module OpenHAB
       # def <level>(msg=nil, &block)
       #   log(severity: <level>, msg: msg, &block)
       # end
+      #
+      # Also creates methods to check if the different logging levels are enabled
+      #
       LEVELS.each do |level|
-        method = level.to_s.downcase
-        define_method(method.to_s) do |msg = nil, &block|
+        define_method(level) do |msg = nil, &block|
           log(severity: level, msg: msg, &block)
         end
+        define_method("#{level}_enabled?") { @sl4fj_logger.send("is_#{level}_enabled") }
+      end
+
+      #
+      # Cleans the backtrace of an error to remove internal calls. If logging is set
+      # to debug or lower, the full backtrace is kept
+      #
+      # @param [Exception] error An exception to be cleaned
+      #
+      # @return [Exception] the exception, potentially with a cleaned backtrace.
+      #
+      def clean_backtrace(error)
+        return error if debug_enabled?
+
+        backtrace = error.backtrace_locations.reject { |line| INTERNAL_CALL_REGEX.match? line.to_s }
+        error.set_backtrace(backtrace.map(&:to_s))
+        error
       end
 
       private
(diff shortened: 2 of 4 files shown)


pacive:
Didn't catch all errors, will have to rework a bit

pacive:
It seems like different oh-classes handles the exceptions differently: If the ScriptEngineManager catches the exception, it is logged without stack trace at ERROR level, but if it's caught by the Scheduler it logs at WARN level with full stack trace. Should it only be logged by the library and not re-raised? Could that cause any issues on the OH-side?

pacive:
Managed it by rescuing errors in different places. This has the added benefit of letting execution blocks run even if one of them raises an error, so you can segment your code if you think something might fail. If the error is on rule parsing it is re-raised to the ScriptEngineManager, with added stack trace to the exception.

lib/openhab/dsl/rules/automation_rule.rb
```diff
@@ -60,6 +65,17 @@ def execute(mod = nil, inputs = nil)
...
 
+        #
+        # Print error and stack trace without calls to internal classes
+        #
+        # @param [Exception] error A rescued error
+        #
+        def print_backtrace(error)
+          # Remove internal calls
+          stack_trace = error.backtrace_locations.reject { |line| INTERNAL_CALL_REGEX.match? line.to_s }
+          logger.warn("#{error.class}: #{error.message}\nIn rule: #{get_name}\n#{stack_trace.join("\n")}")
```
jimtng, line 76: Should this be `logger.error`?

pacive: Of course, that would be better

boc-tothefuture:
The current conversation in jruby is that they will fix the stack trace thing.  Do we still want to do this?  It also won't work outside of a rule?

pacive:
> The current conversation in jruby is that they will fix the stack trace thing. Do we still want to do this? It also won't work outside of a rule?

No, it would only work within a rule context (including methods defined outside a rule and called from the rule), but in both when creating and executing the rule. 

Not sure how the jruby guys plan to implement it, but it would still be the case that the ScriptEngineManager doesn't output the stack trace, so it would have to be caught somewhere before that, either in the scripting library (this is apparently how the jython helper libraries does it) or in the ScriptEngine.

boc-tothefuture:
Do we need the full stack trace? I thought the goal (to be on parity with Jython) was just to output the line and file where the error occurred.

pacive:
If there's an error within the rule body that's just about all that's output (as well as the rule name, which I think could be useful) but if the user have defined other methods in the rule file (or in another file that's required, with methods shared between different rules files), wouldn't it be useful to have those stack frames listed as well?

pacive:
Reworked a bit to put some common logic in the Logging class, and retain the full backtrace (including internal calls) if debug logging is enabled.

jimtng:
> wouldn't it be useful to have those stack frames listed as well?

I agree

> The current conversation in jruby is that they will fix the stack trace thing. Do we still want to do this? It also won't work outside of a rule?

I don't quite understand what the jruby people were planning to do, but it sounds like they are just going to add jython's filename and line number to the existing error message. The jython helper libraries added the stack trace on top of the standard jython error messages too, which seems like what this PR is trying to do as well.

jimtng:
> The errors are rescued for a single execution block, which allows other blocks in the same rule to run as usual.

I wonder if this would create unexpected side effects, e.g. if the subsequent blocks expect a certain state that was supposed to be changed by the prior blocks in which the error occurred. IMO it would be better / more predictable to stop the entire chain of execution should there be an error.

```ruby
rule 'test' do
  on_start
  run { error here }
  run { this shouldn't execute either }
end
```

pacive:
> > The errors are rescued for a single execution block, which allows other blocks in the same rule to run as usual.
> 
> I wonder if this would create unexpected side effects, e.g. if the subsequent blocks expect a certain state that was supposed to be changed by the prior blocks in which the error occurred. IMO it would be better / more predictable to stop the entire chain of execution should there be an error.
> 
> ```ruby
> rule 'test' do
>   on_start
>   run { error here }
>   run { this shouldn't execute either }
> end
> ```

Fair point, one could make an argument for both cases. IMO, if something is that dependant on earlier code execution it shouldn't be split into separate blocks. But one could also make the case that blocks that are independent from each other should be in different rules.

We need to make a decision here.

pacive:
After thinking about it a bit, I'm inclined to agree with @jimtng that the rule should exit completely. If a user desires a different outcome, they could add rescue clauses to the rule, but changing the behaviour in the other direction would not be possible (or at least harder).

lib/openhab/log/logger.rb
```diff
@@ -18,6 +18,11 @@ class Logger
       # @return [Array] Supported logging levels
       LEVELS = %i[TRACE DEBUG WARN INFO ERROR].freeze
```
jimtng, line 19: It seems that we could just use lowercase here, and avoid having to call `to_s` everywhere. LEVELS can be made private so as not to pollute the global namespace if necessary.

boc-tothefuture:
> After thinking about it a bit, I'm inclined to agree with @jimtng that the rule should exit completely. If a user desires a different outcome, they could add rescue clauses to the rule, but changing the behaviour in the other direction would not be possible (or at least harder).

Failing fast is usually better, so I think we are all in agreement.

boc-tothefuture:
Did someone look at the jython source code to see how they added the stack trace?  Do they add the stack trace if it fails outside of a method with a rule decorator?

The other option would be to open a PR on openhab-core or at least an issue on OH core to discuss if the rule engine should provide a stack trace rather than each language having to implement it.

pacive:
> Did someone look at the jython source code to see how they added the stack trace? Do they add the stack trace if it fails outside of a method with a rule decorator?

See the link in @jimtng's comment here: https://github.com/boc-tothefuture/openhab-jruby/issues/17#issuecomment-781825572

From the looks of it, it would only work in a decorated rule.

> The other option would be to open a PR on openhab-core or at least an issue on OH core to discuss if the rule engine should provide a stack trace rather than each language having to implement it.

One of the biggest problems with OH core that I noticed when working on this, is that the error might be caught in different places depending on how/when the error occurs:
* If it occurs when the script is parsed it is caught by the ScritpEngineManagerImpl, that prints the error message, but not the stack trace. This includes when the rule is executed by the on_start trigger.
* If the error occurs within the body of a Timer (including when using delay and/or changed ... for: ...) it is caught by the SchedulerImpl, which outputs the full stack trace.
* If it occurs during regular rule execution it is caught by the RuleEngineImpl, which outputs the less than helpful message:
  ```
  Failed to execute rule '25b796b7-63dc-4bc3-b656-9f81afe942d4': Fail to execute action: 1
  ```
  It also outputs the stack trace if debug or trace logging is enabled (which regular users shouldn´t have to worry about IMO).

When the stack trace is written to the log however, it includes a lot of stack frames from OpenHAB and Jruby as well as from the scripting library and the rules. What I have tried to do here is to only write the lines relevant to the end user (i.e those within the user's own files). If we need additional information to see what happens in the scripting library, these can be included by enabling debug logging.

lib/openhab/log/logger.rb
pacive: This wasn't written by me, but can make that change as well, it doesn't seem to be used outside the Logger module.

jimtng: I understand it was written by @boc-tothefuture originally. It is something I was going to suggest changing, so why not just do it now :) WDYT @boc-tothefuture ?

boc-tothefuture:
@pacive is this one ready?

pacive:
I do have a commit that I haven't pushed yet, that fixes @jimtng's [comment](https://github.com/boc-tothefuture/openhab-jruby/pull/188#discussion_r580939680). Should I include that as well?

lib/openhab/log/logger.rb
boc-tothefuture: I missed this ping.. yep.. no problem, lets downcase and make it a private_constant

lib/openhab/log/logger.rb
```diff
@@ -18,6 +18,11 @@ class Logger
...
 
+      #
+      # Regex for matching internal calls in a stack trace
+      #
+      INTERNAL_CALL_REGEX = %r{(openhab-scripting-.*/lib)|(org/jruby/)}.freeze
```
boc-tothefuture, line 24: This should probably be a private_constant too.

pacive:
Now it's finished!

PR #191, opened 2021-02-19
refactor(event.item): monkey patch ItemEvent to provide decorated event.item

### LINKED ISSUE #189: Singleton on non-persistent Java type
A warning has started to appear recently while using the library:
```
openhab-scripting-2.21.0/lib/openhab/dsl/rules/automation_rule.rb:229: warning: singleton on non-persistent Java type Java::OrgOpenhabCoreItemsEvents::ItemStateChangedEvent (http://wiki.jruby.org/Persistence)
```

jimtng:
Resolve #189 

Instead of looking up and decorating items every time a run / triggered block is encountered, which can occur multiple times when there are multiple run/trigger blocks, the lookup and decoration is done only once in the monkey patched event.

### DIFF
--- a/features/run.feature
+++ b/features/run.feature
@@ -72,18 +72,49 @@ Feature:  run
     Then It should log 'from OFF' within 5 seconds
     Then It should log 'to ON' within 5 seconds
 
-  Scenario: Verify decoration of event.item
+  Scenario Outline: Verify decoration of event.item
     Given items:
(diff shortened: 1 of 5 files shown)


pacive (approved):

jimtng:
One less item lookup in process_trigger_task, because it can use the one already done in event.item

PR #192, opened 2021-02-19
feat(event): add event.state for update trigger

jimtng:
This will add event.item to the run block for `updated` trigger


pacive (approved):
Looking good!

PR #198, opened 2021-02-22
perf(datetime): delegate more methods directly to ZonedDateTime

pacive:
Have defined the methods to be compatible with ruby's Time class, but removed the need to create a new Time instance for most methods.


boc-tothefuture:
Looks like a wday method was added, does anything in the external does need to be updated for any of this?

pacive:
> Looks like a wday method was added, does anything in the external does need to be updated for any of this?

This should work the same as in the ruby Time class, which are linked to in the docs. The reason I couldn't just delegate it is that java reports the days 1-7 (mon-sun) whereas ruby uses 0-6 (sun-sat), so wanted to make it ruby compatible.

PR #200, opened 2021-02-23
feat(persistence): ability to set a default persistence service for the script

jimtng:
This PR adds `def_default_persistence` function as an alternative to using a persistence block.
It can still be overridden by a persistence block should it be required.

Example:

```ruby
require 'openhab'

rule 'Test rule' do
  on_start
  run do
    def_default_persistence :influxdb
    logger.info("Item1's last_update: #{Item1.last_update.is_after(Item2.last_update}") } # Both calls will use :influxdb
  end
end
```

### DIFF
--- a/docs/usage/misc/persistence.md
+++ b/docs/usage/misc/persistence.md
@@ -47,10 +47,10 @@ Number:Power  Power_Usage "Power Usage [%.2f W]"
 
 ```ruby
 # UV_Index average will return a DecimalType
-logger.info("UV_Index Average: #{UV_Index.average_since(12.hours)}") 
+logger.info("UV_Index Average: #{UV_Index.average_since(12.hours, :influxdb)}") 
 # Power_Usage has a Unit of Measurement, so 
 # Power_Usage.average_since will return a Quantity with the same unit
-logger.info("Power_Usage Average: #{Power_Usage.average_since(12.hours)}") 
+logger.info("Power_Usage Average: #{Power_Usage.average_since(12.hours, :influxdb)}") 
 ```
 
 Comparison using Quantity
@@ -81,5 +81,27 @@ persistence(:influxdb) do
   Item1.persist
   Item1.changed_since(1.hour)
   Item1.average_since(12.hours)
+end
+```
+
+## Setting The Default Persistence Service
+
+The default persistence service can be set at the beginning of the execution block (run, triggered, otherwise), and it will affect all persistence operations within that block, unless a specific service is specified in the argument or within a persistence block.
+
+Note that this does not alter the system-wide default persistence service that is configured 
+on the OpenHAB installation. This simply affects the current execution block.
+
+To specify the default persistence service, call the `def_default_persistence` function at the beginning
+of the execution block. Example:
+
+```ruby
+require 'openhab'
+
+rule 'Test rule' do
+  on_start
+  run do
+    def_default_persistence :influxdb
+    logger.info("Item1's last_update: #{Item1.last_update}") 
+  end
 end
 ```
--- a/features/persistence.feature
+++ b/features/persistence.feature
@@ -78,3 +78,26 @@ Feature: persistence
       """
     When I deploy the rule
     Then It should log 'Average: 3 kW' within 10 seconds
+
+  Scenario: Set default persistence
+    Given items:
+      | type         | name         | label | pattern | state |
+      | Number:Power | Number_Power | Power | %.1f kW | 2 kW  |
+    And code in a rules file:
+      """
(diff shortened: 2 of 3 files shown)


boc-tothefuture:
This shouldn't actually work.  These variables are tied to the thread and every single rule in openhab gets its own thread and the thread that parses the rules is in fact a different thread from all of them.

jimtng:
> This shouldn't actually work. These variables are tied to the thread and every single rule in openhab gets its own thread and the thread that parses the rules is in fact a different thread from all of them.

This currently works, both in my installation and on the cucumber test.

Even if it doesn't work, at least the call to `def_default_persistence` can be made at the top of inside the first run block, and it removes the need to have yet another level of block.

I have been thinking about the name, `def_default_persistence` perhaps should be renamed to `use_persistence_service` or just `default_persistence`?

boc-tothefuture:
I really don't know how it would be working... 

Looking at your code, what is supposed to happen if that first rules runs again?  What persistence service is used?  What happens if the first rules runs 5 seconds after the setting the persistence?

What happens if there are multiple rule files?  How is this scoped?

jimtng:
You're correct, it only works because on_start must've executed the run blocks in the same thread as the script loader/parser. When I used a cron trigger, it doesn't work.

So it goes back to just using it inside the run block, ie.

```ruby
rule 'x' do
  changed Something
  run do
    def_default_persistence :influxdb
    Something.last_update
    SomethingElse.last_update
    ...
  end
end
```

pacive:
PersistenceExtensions have methods that don't take a persistence service as argument and uses the default one instead. Could this behavior be implemented instead? 
Something like:
```ruby
          PERSISTENCE_METHODS.each do |method|
            define_method(method) do |timestamp, service = nil|
              if service.nil?
                result = PersistenceExtensions.public_send(method, self, to_zdt(timestamp))
              else
                result = PersistenceExtensions.public_send(method, self, to_zdt(timestamp), service&.to_s)
              end
              QUANTITY_METHODS.include?(method) ? quantify(result) : result
            end
          end
```
That would also make it more consistent with other rule languages.

jimtng:
> PersistenceExtensions have methods that don't take a persistence service as argument and uses the default one instead. Could this behavior be implemented instead?

It is already implemented by passing nil as the service argument. Because Java does not have default arguments, they have to use multiple functions with different signatures, so the one that doesn't need the service name, will call the one with the service name and passing null as the service name, which is exactly what we're doing already.

The problem is that a) I don't know of a programmatic way to change openhab's default, and b) I don't necessarily want to change openhab's default

The idea of this PR is when you don't have a system default service set in openhab, or if it's set to something different, e.g. mapdb, and you want to use influxdb, you can still get the same effect as if influxdb was the default, i.e. not having to specify the service name.

In other words, this is the scenario I encounter and I want to make it easier:
* OpenHAB System Default: None or Mapdb
* Some of my items are persisted to Influxdb

In this situation, I must usually always specify the service name (influxdb) on every function call, because it's not the system default.

Before this PR, we have the "persistence block" so we can just specify it in a block:
```
persistence :influxdb do
  Item1.last_update # no need to specify influxdb here, it will pick it up from the persistence block above
end
```

I am trying to avoid having to have yet another level of a block, so this PR enables this syntax as well:
```
def_default_persistence :influxdb
Item1.last_update
```

pacive:
> It is already implemented by passing nil as the service argument. Because Java does not have default arguments, they have to use multiple functions with different signatures, so the one that doesn't need the service name, will call the one with the service name and passing null as the service name, which is exactly what we're doing already.

Not sure if jruby does some magic here, but wouldnt that be like calling e.g. `public static ZonedDateTime lastUpdate(Item item, String serviceId)` with `null` as the last argument? If it works it's all good! And I get your point about not wanting to change the default service.

jimtng:
Come to think of it, if the thread gets reused to execute another rule, the thread variable could still apply to that other rule, creating an unpredictable situation. So this is a bad idea. Closing this PR.

boc-tothefuture:
@jimtng I do think its valuable problem you are solving.  You could create a 'persistence' parameter to a rule and then from within the rule wrap all executions in the persistence block which takes care of setting/unsetting the variable.

jimtng:
> @jimtng I do think its valuable problem you are solving. You could create a 'persistence' parameter to a rule and then from within the rule wrap all executions in the persistence block which takes care of setting/unsetting the variable.

What a good idea. I hadn't thought of that. To clarify, is this is what you're proposing:

```ruby
rule 'x' do
  on_start
  persistence :influxdb
  run { Item1.last_update }
end
```

Issue #202, opened 2021-02-24
Group Redesign


boc-tothefuture:
Opening up this issue to discuss redesign of groups and to consolidate other threads

boc-tothefuture:
Also it should support:

on|off|on?|off? (if it's based on Switch type)
open?|closed? (if based on Contact type)
.truthy? so that it can be used with only_if

From @jimtng in #180

boc-tothefuture:
#185 makes item.groups to return the list of groups that the item is a member of, i.e. its ancestors.
However, Group.groups currently would return the list of its members (which is of group type), i.e. its descendants

From #201

boc-tothefuture:
Let's use this issue to layout the group redesign.

boc-tothefuture:
> Also it should support:
> 
> on|off|on?|off? (if it's based on Switch type)
> open?|closed? (if based on Contact type)
> .truthy? so that it can be used with only_if
> 
> From @jimtng in #180

Good ideas..  this should be much easier to support now that we item_command

truthy? would be supported based on the underlying items truthy correct? 
What do we do if only some of the items support truthy?

pacive:
I have made a PR that I think can be used as a base for discussion.

> How do we get the type?

The accepted states and command can be gotten from the OH GroupItem and patched on the Item's singleton class, see my PR. If you know a better way I'm all ears. But would be even easier if #179 gets merged.

boc-tothefuture:
#179 has conflicts

pacive:
> #179 has conflicts and I thought also outstanding items to address?

Didn't notice there had been conflicts, fixed now, as well as the other items!

jimtng:
> truthy? would be supported based on the underlying items truthy correct?
> What do we do if only some of the items support truthy?

We'll only support truthy? if the Group was defined like this:

```
Group:Switch:OR(ON,OFF)
```

It doesn't matter if some/all the members do not support truthy? It's the Group's state that matters.

I haven't checked, but my guess is that when it was defined like above, getAcceptedCommand Types will return OnOffType. So when it does return OnOffType, we can define truthy?

jimtng:
We also need to be able to do this:

```
if Group == OPEN
or
if Group == ON
```
Which are the equivalent of Group.open? and Group.on?

jimtng:
Ref: https://www.openhab.org/docs/configuration/items.html#derive-group-state-from-member-items
[...]

Issue #180, opened 2021-02-18
Group should support << 


boc-tothefuture:
Details in this post: https://community.openhab.org/t/jruby-openhab-rules-system/110598/126

jimtng:
Also it should support .truthy? if not already, so that it can be use in only_if

PR #203, opened 2021-02-24
feat(groups): support command and <<

boc-tothefuture:
This addresses #180 

We have a broader #202 to address other group issues, I would like to have this one reviewed and released since a user was asking for it while we work through the broader strategic group issues.

### DIFF
--- a/features/groups.feature
+++ b/features/groups.feature
@@ -93,3 +93,17 @@ Feature:  groups
       """
     When I deploy the rules file
     Then It should log 'event.item is Livingroom_Temp' within 5 seconds
+
+  Scenario Outline: Commands sent to a group are propagated to the group items
+    Given code in a rules file
+      """
+        GroundFloor.<method> <value>
+      """
+    When I deploy the rules file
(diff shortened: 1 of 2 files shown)


features/groups.feature
```diff
@@ -93,3 +93,17 @@ Feature:  groups
...
+      """
+        GroundFloor.<method> <value>
+      """
+    When I deploy the rules file
+    Then "Bedroom_Temp" should be in state "<value>" within 5 seconds
+    And "Den_Temp" should be in state "<value>" within 5 seconds
+    Examples:
+      | method  | value |
+      | <<      | 55    |
+      | command | 55    |
```
jimtng, line 108:
```suggestion
      | command | 60    |
```
So that we know it wasn't the value from before

boc-tothefuture: Can do..  but the cucumber framework wipes out all items and rules between runs.  

But doenst' hurt either way.

jimtng:
LGTM

pacive (approved):
Looks good

Issue #185, opened 2021-02-19
add groups method to Item


boc-tothefuture:
An item should be able to lookup which groups it belongs to.

Proposed syntax:
```
Item.groups
```
Returns an array (potentially empty) of any groups that the items belongs to.

boc-tothefuture:
closed by #199

PR #204, opened 2021-02-24
refactor(group_item)!: add GroupItem delegation

### LINKED ISSUE #202: Group Redesign

pacive:
I have been experimenting with the group implementation for the last few days, and thought it might be a good starting point for discussing #202. This would be a breaking change, since some functionality behaves differently, but all tests in `groups.feature` pass with these small modifications.

* In most cases the group behaves just like a regular Item, and it can be commanded and updated.
* Iteration through the direct members can be done with `GroupItem.each` and through this, all Enumerable methods are supported. To get all descendant recursively you can use `all_members`.
* `each` and `all_members` decorates the items.
* If the group have a base Item type, it automatically adds methods from that type (e.g `on?`, `on`, `off?`, `off` for Group:Switch)

Summary of points raised in #202 
- [x] Also it should support:
       on|off|on?|off? (if it's based on Switch type)
       open?|closed? (if based on Contact type)
       .truthy? so that it can be used with only_if **(works by delegating to baseItem, but test should be added)**
- [x] Cannot get group state
- [x] #185 makes item.groups to return the list of groups that the item is a member of, i.e. its ancestors.
       However, Group.groups currently would return the list of its members (which is of group type), i.e. its descendants
- [x] We also need to be able to do this:
        ```
        if Group == OPEN
        or
        if Group == ON
        ```
       Which are the equivalent of Group.open? and Group.on?
- [x] Groups can also act like DateTime and Number, so they should also be comparable to DateTime, TimeOfDay, 
       ZonedDateTime, etc. when it's a DateTime Group, and comparable to numbers, or even quantity when it's a number.
       **(For quantity both group and items need to have the same type - OpenHAB dependant)**
- [x] This also means that all of our other types (OnOffType,OpenClosedType) need to be group aware in their equals and case operators.
- [x] Make sure it also has access to metadata and persistence, i.e. Group.meta, Group.last_update, etc. 
- [x] Update documentation
      
Closes #202 


boc-tothefuture:
It is more than just each, I want all enumerable methods (max, min, etc)..
The delegates have a priority order.  Let me put a comment in the code.

lib/openhab/dsl/group.rb
```diff
@@ -38,63 +34,11 @@ class Groups < SimpleDelegator
       #
       def groups
         # rubocop: disable Style/GlobalVars
-        Groups.new(OpenHAB::Core::EntityLookup.decorate_items($ir.items.select { |item| item.is_a? GroupItem }))
+        Groups.new(OpenHAB::Core::EntityLookup.decorate_items(
+                     $ir.items.select { |item| item.is_a? Java::OrgOpenhabCoreItems::GroupItem }
```
boc-tothefuture, line 38: Should be able to just grep

lib/openhab/dsl/items/group_item.rb
```diff
@@ -0,0 +1,74 @@
...
+
+          item_missing_delegate { @group_item }
+          add_state_methods
+          add_command_methods
+        end
+
+        #
+        # Iterates through the direct members of the Group
+        #
+        def each(&block)
```
boc-tothefuture, line 38: can remove this method

lib/openhab/dsl/items/group_item.rb
```diff
@@ -0,0 +1,74 @@
...
+        #
+        # Create a new GroupItem
+        #
+        # @param [Java::Org::openhab::core::items::GroupItem] group_item OpenHAB GroupItem to delegate to
+        #
+        def initialize(group_item)
+          @group_item = group_item
+
+          item_missing_delegate { @group_item }
+          add_state_methods
```
boc-tothefuture, line 31: item_missing_delegate { OpenHAB::Core::EntityLookup.decorate_items(@group_item.members.to_a) }

lib/openhab/dsl/items/items.rb
```diff
@@ -17,7 +17,8 @@ class Items < SimpleDelegator
         # @param [String] name
         # @return Item from registry, nil if item missing or requested item is a Group Type
         def[](name)
-          OpenHAB::Core::EntityLookup.lookup_item(name)
+          item = OpenHAB::Core::EntityLookup.lookup_item(name)
+          item.is_a? OpenHAB::DSL::Items::GroupItem ? nil : item
```
boc-tothefuture, line 21: Need parens around the is_a? parameter when using a ternary.

pacive:
> It is more than just each, I want all enumerable methods (max, min, etc)..

Aren't those added when the Enumerable module is included, as long as the members define <=>?

lib/openhab/dsl/items/group_item.rb
```diff
@@ -0,0 +1,74 @@
...
+          OpenHAB::Core::EntityLookup.decorate_items(@group_item.all_members.to_a)
+        end
+
+        private
+
+        #
+        # Add state methods for the Groups base item type
+        #
+        def add_state_methods
+          data_enums = get_accepted_data_types.select(&:is_enum)
```
boc-tothefuture, line 56: Do we want to support that face that group type can change distinct from the rule runtime?

lib/openhab/dsl/items/items.rb
```diff
@@ -33,12 +34,13 @@ def include?(name)
...
       end
 
-      java_import org.openhab.core.items.GroupItem
       # Fetches all non-group items from the item registry
       # @return [OpenHAB::DSL::Items::Items]
       def items
         # rubocop: disable Style/GlobalVars
-        Items.new(OpenHAB::Core::EntityLookup.decorate_items($ir.items.reject { |item| item.is_a? GroupItem }))
+        Items.new(OpenHAB::Core::EntityLookup.decorate_items(
+                    $ir.items.reject { |item| item.is_a? Java::OrgOpenhabCoreItems::GroupItem }
```
boc-tothefuture, line 42: grep_v instead

boc-tothefuture (changes requested):

boc-tothefuture:
> > It is more than just each, I want all enumerable methods (max, min, etc)..
> 
> Aren't those added when the Enumerable module is included, as long as the members define <=>?

Oh, I totally missed you included enumerable!  

My review may be totally invalid.

If you include enumerable do you only have to implement each?

pacive:
> > > It is more than just each, I want all enumerable methods (max, min, etc)..
> > 
> > 
> > Aren't those added when the Enumerable module is included, as long as the members define <=>?
> 
> Oh, I totally missed you included enumerable!
> 
> My review may be totally invalid.
> 
> If you include enumerable do you only have to implement each?

Yes, it's all that should be required
https://ruby-doc.org/core-2.5.8/Enumerable.html

Thanks for the comments, will get on it. One thing I'm struggling with right now, and which need to be discussed is how to deal with Groups in triggers (changed, updated, received_command). The way it it right now it creates a groupStateChanged trigger, which triggers on group members. To trigger on the group itself you have to use Group.name. Would it be best to try to make it work like before?

lib/openhab/dsl/items/group_item.rb
pacive: Wouldn't that mess with things like `Group.state` and other methods delegated to the OH GroupItem?

boc-tothefuture:
i deleted my requests that were wrong because I missed that it was including enumerable..  great idea.

boc-tothefuture:
> Wouldn't that mess with things like Group.state and other methods delegated to the OH GroupItem?

The way did enumerable was better.. so it doesn't matter.  But the answer is no.  The design of the item_missing_delegate is order dependant.  In other words if the first thing you delegate to supports .state it will call .state on that, then it will check the next delegate and so on.

lib/openhab/dsl/items/group_item.rb
pacive: Not sure I follow. Isn't the object recreated on every call to const_missing? Or is the result of those calls saved in memory when the rule is parsed and reused on every rule execution? In the latter case this would probably need to be reworked somehow.

lib/openhab/dsl/items/group_item.rb
```diff
@@ -0,0 +1,83 @@
...
+
+        #
+        # Get all members of the group recursively
+        #
+        # @return [Array] An Array containing all descendants of the Group
+        def all_members
+          OpenHAB::Core::EntityLookup.decorate_items(@group_item.all_members.to_a)
+        end
+
+        def items
```
boc-tothefuture, line 55: We need to rethink this.

pacive:
> One thing I'm struggling with right now, and which need to be discussed is how to deal with Groups in triggers (changed, updated, received_command). The way it it right now it creates a groupStateChanged trigger, which triggers on group members. To trigger on the group itself is currently not possible. Would it be best to try to make it work like before?

Reverted to using the GroupItems struct and Group.items for groupChangedTrigger.

lib/openhab/dsl/items/group_item.rb
```diff
@@ -0,0 +1,83 @@
...
+          GroupItems.new(group: self)
```
jimtng, line 56: While we're at it, should we change this to `members`, and alias `items` to it, for backwards compatibility?

`members` should also be an array of direct descendants, in contrast to all_members.

It would read better in the rule as `changed Switches.members` rather than Switches.items.

lib/openhab/dsl/items/group_item.rb
pacive: Agreed. Tried without it (see my comment above) but reverted to see if other issues need to be worked out first.

lib/openhab/dsl/items/group_item.rb
pacive: The problem is that `members` is used by the each-method to get the group members. The items is (i believe?) only ment to be used in triggers to differentiate between member of-triggers and item triggers when passing a group item.

jimtng: Is `members` already delegated to group_item? In any case, can we somehow also use that for member-of triggers thus making it dual purpose?

So that `changed Switches` is for a trigger on the GroupItem itself, and `changed Switches.members` is for a trigger on the members of the group (ie member-of trigger)

jimtng: PS I should be commenting on the `def items` line instead.

pacive: If `members` is used in the trigger I think it would expand to a list of items, which would create a separate trigger for all the items. But this wouldn't include any new items added to the group later. So we need some way to differentiate between:

1. Changes/updates to the actual group item's state (itemState*Trigger with the group as item)
2. Changes to one of the group's members (groupState*Trigger)
3. Expanding to a list of items for which to create individual triggers (multiple itemState*Trigger)

Not sure if 3 is needed, but that is the current consequence of using `members` in the trigger.

boc-tothefuture: What @pacive outlines is the core problem..  Differentiating between when it should be a group trigger or an item trigger.  I considered a few possibilities and all of them have drawbacks.
One option is to have member_updated, member_changed, as well as updated and changed, instead but I wasn't particularly fond of that and it gets messy with received_command too.

another options is
changed Group, :members

lib/openhab/dsl/items/group_item.rb
pacive: Made a test rule (where Switches is a group item):
```ruby
rule 'test' do
  every 5.seconds
  run do
    logger.info(Switches.__id__)
    logger.info(Switches.__id__)
  end
end
```
Which confirms that the object gets recreated on every call, so even if the underlying item should change between calls, the object gets re-instantiated and the correct methods should get created.

lib/openhab/dsl/items/group_item.rb
jimtng: https://github.com/pacive/openhab-jruby/pull/1

lib/openhab/dsl/items/group_item.rb
```diff
@@ -0,0 +1,90 @@
...
+          item_missing_delegate { @group_item }
+          add_state_methods
+          add_command_methods
+        end
+
+        #
+        # Iterates through the direct members of the Group
+        #
+        def each(&block)
+          OpenHAB::Core::EntityLookup.decorate_items(@group_item.members.to_a).each(&block)
```
jimtng, line 44: This is not a suggestion, more of a question for me to learn: Why not do item_missing_delegate { @group_item.members } instead of implementing each?

lib/openhab/dsl/items/group_item.rb
pacive: Created a trial with using `changed Group, :members` (only for the changed trigger so far). Let me know what you think.

jimtng: I think the syntax `changed Group.members` looks better / less confusing than `changed Group, :members`.

Consider:
```
changed Group, :members, Item1, Item2
vs
changed Group.members, Item1, Item2
```
The :members syntax looks a bit odd because it's not a part of Group in the definition.

pacive: I can agree with that, but can't see how we can work around the issue of using this both for signaling a group trigger and using it for getting an Array of the members. (Without resorting to the same technique of using a Struct or mediating class as the previous implementation)

jimtng: I've just added MembersOfGroup.group to refer back to the group_item

jimtng: @pacive Thanks for exploring this further. I see the problem with Group.members, and not sure how to get around the flatten issue.

How about `changed Item2, Item3, member_of: Group, to: XX`
`changed Item2, Item3, member_of: [Group1, Group2]`

It's not ideal either, but an alternative to `Group, :members` syntax to consider.

jimtng: Or we can stick to Group.items like before - less breaking change

pacive: That would work, but could only be used once. `changed Item1, member_of: Group1, member_of: Group2` wouldn't work. Another approach would be to make it a method: `changed Item1, member_of(Group1), member_of(Group2)`, but it doesn't read as nice.

jimtng: > That would work, but could only be used once. `changed Item1, member_of: Group1, member_of: Group2` wouldn't work. Another approach would be to make it a method: `changed Item1, member_of(Group1), member_of(Group2)`, but it doesn't read as nice.

`member_of: [Group1, Group2]` ?

pacive: That would be the way to do it, yes, but could it be confusing for users?

jimtng: Another alternative: `changed Group.member_of` - basically renaming it from `Group.items`

boc-tothefuture (commented):
Is there a reasonable use case in which we would expect a user to do:
`changed group.members` and not want it to mean all members of the group whose membership may change over time?

I think it is pretty reasonable that we don't support the use case where they want individual triggers on the items in a group, possible but more difficult to do.  

If so, there might be a way to make that work.

pacive:
> Is there a reasonable use case in which we would expect a user to do:
> `changed group.members` and not want it to mean all members of the group whose membership may change over time?
> 
> I think it is pretty reasonable that we don't support the use case where they want individual triggers on the items in a group, possible but more difficult to do.
> 
> If so, there might be a way to make that work.

I think we all agree on that, but the question is what the `members` method should return to be able to identify this   and still be able to use it for getting an array of all members.

boc-tothefuture:
> > Is there a reasonable use case in which we would expect a user to do:
> > `changed group.members` and not want it to mean all members of the group whose membership may change over time?
> > I think it is pretty reasonable that we don't support the use case where they want individual triggers on the items in a group, possible but more difficult to do.
> > If so, there might be a way to make that work.
> 
> I think we all agree on that, but the question is what the `members` method should return to be able to identify this and still be able to use it for getting an array of all members.

yeah.. let me try something.. not sure it will work.

pacive:
If everything works as I think it does, comparisons was actually rather easy. The OH GroupItem holds a baseItem as it's state, so it should be possible to just decorate that and let it handle the comparison. Will add some tests to verify.

boc-tothefuture:
Take a look at this branch which seems to do what we want for group.members
https://github.com/boc-tothefuture/openhab-jruby/tree/group_item

jimtng:
> Take a look at this branch which seems to do what we want for group.members
> https://github.com/boc-tothefuture/openhab-jruby/tree/group_item

Noob git/github question: Is there a way to see a diff against the main branch?

> Take a look at this branch which seems to do what we want for group.members
> https://github.com/boc-tothefuture/openhab-jruby/tree/group_item

@pacive does this not have the same issue with flatten?

pacive:
> > Take a look at this branch which seems to do what we want for group.members
> > https://github.com/boc-tothefuture/openhab-jruby/tree/group_item
> 
> @pacive does this not have the same issue with flatten?

I think it might, unfortunately. But it gave me an idea, if you just pass `group_item.each` (without a block) it creates an Enumerator that gets delegated to instead, which have all the methods of enumerable, but not `to_ary` or `flatten`.

Another way would be to override `respond_to?` to exclude `to_ary`:
```ruby
def respond_to?(method)
  return false if method.to_sym == :to_ary

  super
end
```
Will experiment a bit.
@boc-tothefuture WDYT? (See [here](https://github.com/pacive/openhab-jruby/pull/1#discussion_r582736003) for our discussion on the `flatten` issue)

boc-tothefuture:
Why don't we just not call flatten on group items?
```
groups, other = items.partition { |item| item.is_a? OpenHAB::DSL::Items::GroupItem::GroupMembers }
items = groups + other.flatten
items.each do |item|
```
Seeing the other thread, you can also prevent recursion of flatten by passing it a depth.

`items.flatten(1).each`

pacive:
> Why don't we just not call flatten on group items?
> 
> ```
> groups, other = items.partition { |item| item.is_a? OpenHAB::DSL::Items::GroupItem::GroupMembers }
> items = groups + other.flatten
> items.each do |item|
> ```
> 
> Seeing the other thread, you can also prevent recursion of flatten by passing it a depth.
> 
> `items.flatten(1).each`

If it doesn't break any existing intended use case, that would probably be the best solution!

boc-tothefuture:
We need to add a test case of the group membership changing..  I can do it, but no idea which branch to do it in.
Are you going to pull in my changes into your branch and we work on it there?

pacive:
> We need to add a test case of the group membership changing.. I can do it, but no idea which branch to do it in.
> Are you going to pull in my changes into your branch and we work on it there?

Reverted my previous commit and merged yours, had to do some rebasing, so might be easier to delete your branch and recreate it.

pacive:
Removing `flatten` entirely, and just using `items.each do |item|` works, but that would make the [third example on thins page](https://boc-tothefuture.github.io/openhab-jruby/examples/) fail unless it's changed to
```ruby
virtual_switches = items.select { |item| item.is_a? Switch }
                        .select { |item| item.label&.include? 'Virtual' }

rule 'Log whenever a Virtual Switch Changes 2' do
  changed *virtual_switches
  run { |event| logger.info "#{event.item.id} changed from #{event.last} to #{event.state} 2" }
end
```
I.e adding the splat operator to the array passed to `changed`.

boc-tothefuture:
Any reason not to do the partition and flatten? I can just see some use cases were we get an array.

pacive:
> Any reason not to do the partition and flatten? I can just see some use cases were we get an array.

No, that would work fine I think. Would just need to restructure the methods a bit to make rubocop happy.

boc-tothefuture:
Yeah, I think we need it multiple places.. probably best to move it to trigger.rb in the module so we can reuse it across the triggers?

pacive:
Good idea, it works now, but would need to make it recursive in order for it to work if someone adds an array of Group.members to the trigger.

pacive:
The groups.feature tests got really heavy with all the comparisons, so moved them to comparisons.feature to speed it up.

lib/openhab/dsl/items/group_item.rb
pacive: Added a test to verify that this works as intended (if a group changes type that is reflected even in running rules)

features/changed.feature
```diff
@@ -134,4 +134,32 @@ Feature:  changed
       | OFF           | OFF   | should not |
       | OFF           | ON    | should     |
       | ON            | ON    | should not |
-      | ON            | OFF   | should     |
\ No newline at end of file
+      | ON            | OFF   | should     |
+
+  Scenario: GroupMembers are separated from items in triggers
+    Given group "Switches"
```
boc-tothefuture, line 140: you can do the groups: syntax here

features/changed.feature
```diff
@@ -134,4 +134,32 @@ Feature:  changed
...
+    And a deployed rule:
+      """
+      rule 'Nested groups' do
+        changed Switches.members, [[Contacts.members], Switch3]
+        triggered { |item| logger.info("#{item.id} triggered the rule") }
+      end
+      """
+    When I add items:
+      | type    | name     | group    |
+      | Contact | Contact3 | Contacts |
```
boc-tothefuture, line 158: You can set the item state here too.

features/comparisons.feature
```diff
@@ -318,6 +341,26 @@ Feature: comparisons
         [ 10                         , '>'  , DecimalType.new(5)         , true  ]  ,
         [ 10                         , '>'  , DecimalType.new(10)        , false ]  ,
 
+        # Groups
+        [ Temperatures               , '<'  , Temperatures.max           , true  ]  ,
```
boc-tothefuture, line 345: These tests should be bidirectional, right?

boc-tothefuture:
This is really coming along.

boc-tothefuture:
I think we are going to have to make a special GroupItem for numbers and dimmers that extends from Numeric.

Without that I don't think ranges and case statements will work correctly.

lib/openhab/dsl/items/group_item.rb
pacive: Ah, now I understand, good point. But wouldn't this be a problem with regular items as well? If it's saved to an array as a SwitchItem and then changed to a DimmerItem the object in the array would still be a SwitchItem, right?

Either case since it delegates to the base item (which is decorated on per-call basis) it should work even with those methods are removed.

But OH also lets you e.g. send a ON command to a group of Switches and Dimmers even if the group hasn't got a type, which wouldn't work with the`on` method.

If I would add `method_missing` would this shadow the one defined in ItemDelegate? Which one would get called first?

features/comparisons.feature
pacive: Yes, most of them at least, do we e.g. support String <=> Quantity type?

pacive:
> I think we are going to have to make a special GroupItem for numbers and dimmers that extends from Numeric.
> 
> Without that I don't think ranges and case statements will work correctly.

Was hoping to avoid making different group item types, but for this you might be right. 

But it would only work for groups with a declared type. Even if you have a group with only Number items (but not defined as Group:Number:..) there would be no way of knowing how to wrap it, unless you make it generic for all items that have DecimalType as accepted data type.

OTOH for those groups that do have a type (and consequently a base item) this would handle all comparisons, the only downside being the same we had with TimeOfDay before, that it have to compare against every single item in the range until it finds a match. Not sure how that works for non-integer values though.

Any ideas how this could work that I have missed?

features/changed.feature
pacive: Should I remove that from the Examples section below then, and just have the state set there? Then I would have to add a separate 'It should log ...' statement for that item?

features/comparisons.feature
pacive: I will create an issue for this, I would especially expect SwitchItem == OtherSwitchItem to work, and the other ones need to be discussed and tests created. Feels like this is outside the scope of this PR though.

pacive:
> I think we are going to have to make a special GroupItem for numbers and dimmers that extends from Numeric.
> 
> Without that I don't think ranges and case statements will work correctly.

Update: Ranges actually work well, but only if the group is defined as Group:Number of Group:Dimmer, since it has a NumberItem/DimmerItem as delegate, which handles the comparisons. Doesn't look like it has to make a lot of comparisons (not that shows in the log at least).

How important is it to support these functions for groups that doesn't have a type defined? The state of a generic group is not predictable anyway

boc-tothefuture:
> > I think we are going to have to make a special GroupItem for numbers and dimmers that extends from Numeric.
> > Without that I don't think ranges and case statements will work correctly.
> 
> Update: Ranges actually work well, but only if the group is defined as Group:Number or Group:Dimmer, since it has a NumberItem/DimmerItem as delegate, which handles the comparisons. Doesn't look like it has to make a lot of comparisons (not that shows in the log at least).
> 
> How important is it to support these functions for groups that doesn't have a type defined? The state of a generic group is not predictable anyway

Do things like:
 5 + GroupItemThatIsNumeric

work because of the delegation?

pacive:
> Do things like:
> 5 + GroupItemThatIsNumeric
> 
> work because of the delegation?

Yes, everything that is supported by the base item should work.

lib/openhab/dsl/items/group_item.rb
pacive: Seems like the state methods aren't reliable for type-less groups anyway, since the group doesn't update its state if no aggregation function is defined, will remove them so we don't confuse users.

features/changed.feature
boc-tothefuture: This is fine, I misunderstood what was happening here.

boc-tothefuture:
> However, Group.groups currently would return the list of its members (which is of group type), i.e. its descendants
> (Not tested, but think it should work like other items)

Since we are going to lose this functionality, anyone opposed to providing an optional symbol to the members and all_members function.  symbols would ':items' and ':groups' which would filter appropriately?

pacive:
> > However, Group.groups currently would return the list of its members (which is of group type), i.e. its descendants
> > (Not tested, but think it should work like other items)
> 
> Since we are going to lose this functionality, anyone opposed to providing an optional symbol to the members and all_members function. symbols would ':items' and ':groups' which would filter appropriately?

That's a good idea! Would be easy to implement.

pacive:
Wasn't as easy to implement as I first thought, the `getAllMembers` filters out the groups. But it's possible to use the `getMembers(Predicate<item> filterItem)` method, even from ruby. Should we use the default behaviour (exclude group items) if called without filter arguments and use `:groups` to get all groups and `:all` to get everything instead?

jimtng:
The docs need to be updated as well:
- Document Group methods
  - Remove `items`
  - Add `groups`, `members`, `all_members`, `command`, `<<`, `update`, `on`, `off`, `toggle` (is it available?), `open`, etc
- Group member-of trigger syntax (with Group.member)
- Update examples
etc.

pacive:
> * Add `groups`, `members`, `all_members`, `command`, `<<`, `update`, `on`, `off`, `toggle` (is it available?), `open`, etc

`on`, `off`, `open`, `closed`, `toggle` etc are available if the group has a type that supports them, just as if it was a regular item. But yeah, this needs to be documented. Will hold it off until we have decided on the last issues though!

pacive:
> Wasn't as easy to implement as I first thought, the `getAllMembers` filters out the groups. But it's possible to use the `getMembers(Predicate<item> filterItem)` method, even from ruby. Should we use the default behaviour (exclude group items) if called without filter arguments and use `:groups` to get all groups and `:all` to get everything instead? And should this be possible on both the `members` and `all_members` method or just the latter?
> 
> Edit: another possibility would be to allow a block to be passed to the members/all_members method, to allow for more flexibility. Or should we just leave this to `select`?

I decided to only implement this on the `all_members` method, I think that would cause the least confusion. By default it returns only the non-group Items like the OH GroupItem, but it allows to pass either `:groups` to get all subgroups (recursively) or `:all` to get all items (including Groups). Alternatively you can pass a block that basically works like `Enumerable#select`, and let's you have more advances filters (see groups.feature for an example). In this case it also includes all items and groups. WDYT?

pacive:
Ran into issues with the persistence as well. The problem is that the methods in the `Persistence` module operates on the monkey patched OH item, so the delegation to the base item isn't available. So when the `Persistence#quantify` method tries to get the unit, there is no such method available (since the OH GroupItem doesn't have it). It does however respond to `unit` but that is the unit method in `Openhab::DSL::Units` that gets added to all objects. It requires an argument however, which causes an ArgumentError to be raised. 

I managed to work around it by using `get_unit` instead, but this means that units won't be available for persisted groups.

Would it be possible to have the `Persistence` module `include`d in the item classes instead of monkey patched? I think that would solve the problem.

jimtng:
> Ran into issues with the persistence as well. The problem is that the methods in the `Persistence` module operates on the monkey patched OH item, so the delegation to the base item isn't available. So when the `Persistence#quantify` method tries to get the unit, there is no such method available (since the OH GroupItem doesn't have it). It does however respond to `unit` but that is the unit method in `Openhab::DSL::Units` that gets added to all objects. It requires an argument however, which causes an ArgumentError to be raised.
> 
> I managed to work around it by using `get_unit` instead, but this means that units won't be available for persisted groups.
> 
> Would it be possible to have the `Persistence` module `include`d in the item classes instead of monkey patched? I think that would solve the problem.

If that would solve the issue without breaking anything, I have no objections.

pacive:
Seems like it doesn't work out of the box, since the `PersistenceExtensions` methods expect a Object that implements the `Item` interface. We would need to have a common method to get access to the delegate item which would require more refactoring. So I'll just leave this with the workaround in the meantime.

pacive:
I believe this is ready now, have updated the docs and think I found everything that needs to be changed. Will be quite a big job reviewing it, but if you have the time please take a good look to see if I missed something or if anything else should be added.

jimtng (commented):
I just have some minor suggestions.I am very excited about this PR. Thank you!

docs/usage/items/groups.md
```diff
@@ -9,19 +9,34 @@ grand_parent: Usage
...
 ## Group Methods
 
 | Method             | Description                                                                                     |
 | ------------------ | ----------------------------------------------------------------------------------------------- |
-| group              | Access Group Item                                                                               |
-| items              | Used to inform a rule that you want it to operate on the items in the group (see example below) |
-| groups             | Direct subgroups of this group                                                                  |
-| set methods        | All methods [here](https://ruby-doc.org/stdlib-2.5.0/libdoc/set/rdoc/Set.html)                  |
+| members            | Used to inform a rule that you want it to operate on the items in the group (see example below) |
+| all_members        | Gets all descendants of the group recursively excluding groups. Pass :all as argument to include groups as well, or :groups to only get the sub groups |
```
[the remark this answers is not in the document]
boc-tothefuture: Doesn't members all do dual duty now?  It also just provides access to the direct members of the of the group?

docs/usage/items/groups.md
```diff
@@ -9,19 +9,34 @@ grand_parent: Usage
 
 # Groups
 
-A group can be accessed directly by name, to access all groups use the `groups` method. 
+A group can be accessed directly by name, to access all groups use the `groups` method. Groups behave like a regular
+Item, but also lets you iterate through it's members and use all the available methods of 
+[Enumerable](https://ruby-doc.org/core-2.5.0/Enumerable.html). If the group have a type all methods of that type is
+directly available.
```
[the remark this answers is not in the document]
pacive: The `groups` method works like for other items now, it returns the parent groups of the item

docs/usage/items/groups.md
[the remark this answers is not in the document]
pacive: `members` work just like using the enumerable methods directly on the item, that's why I only mention that it's used in triggers.

A mention that `all_members` accept a block can be added though.

features/groups.feature
```diff
@@ -10,52 +10,68 @@ Feature:  groups
...
       | Number | Bedroom_Temp    | Bedroom temperature     | 50    | GroundFloor, Temperatures |
       | Number | Den_Temp        | Den temperature         | 30    | GroundFloor, Temperatures |
 
-
   Scenario: Ability to operate on the items in a group using enumerable methods
     Given code in a rules file
       """
       logger.info("Total Temperatures: #{Temperatures.count}")
-      logger.info("Temperatures: #{House.sort_by(&:label).map(&:label).join(', ')}")
+      logger.info("Temperatures: #{House.all_members.sort_by(&:label).map(&:label).join(', ')}")
```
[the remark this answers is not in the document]
pacive: I only changed the syntax here, but I think it was meant to show that you can get members of subgroups (excluding the groups) with this method

docs/usage/items/groups.md
```diff
@@ -9,19 +9,34 @@ grand_parent: Usage
...
+
+Groups can be used in triggers in two different ways:
+
+```ruby
+changed Switches # Executes the rule when the state of the group item changes
+```
+or
+
+```ruby
+changed Switches.members # Executes the rule when any of the groups members changes it's state
```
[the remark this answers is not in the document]
pacive: One of the hardest things about the English language 😉

docs/examples/conversions.md
```diff
@@ -41,15 +41,15 @@ end
...
 ```ruby
 rule 'Snap Fan to preset percentages' do
-  changed(*CeilingFans)
+  changed CeilingFans.members
   triggered do |item|
     snapped = case item
               when 0...25 then 25
               when 26...66 then 66
-              when 67...100 then 100
+              when 67..100 then 100
```
boc-tothefuture, line 49: The triple dot was correct here. Can you put it back? 
You want to exclude the ending value.  If it is already at 100 it shouldn't take any action.

boc-tothefuture:
Great PR.  I did find one regression in the docs which I highlighted above.  
I think there is a couple of open conversations then this one is ready.  
@pacive - let me know when you think it is ready for merge.

docs/examples/conversions.md
pacive: Ah, didn't read the rest of the rule carefully enough, Will revert it!

docs/usage/items/groups.md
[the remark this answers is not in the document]
pacive: Should I add that `members` return a collection of the member items as well?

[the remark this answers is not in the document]
boc-tothefuture: I guess it isn't needed and perhaps we don't want to highlight that in case we need to change it for some reason in the future.

boc-tothefuture:
@pacive Can you rebase as a single commit?

pacive:
Absolutely! Should I tag it as a `feat!:` or `refactor!:`?

jimtng:
> Absolutely! Should I tag it as a `feat!:` or `refactor!:`?

I'd say this is more of a refactor than a new feature

jimtng (approved):
Thanks again for this great PR

boc-tothefuture (approved):

pacive:
If you don't have any other comments, I feel this is ready to be merged now!

boc-tothefuture:
Great work @pacive

PR #207, opened 2021-03-01
perf(logging): use block syntax to log method calls

### LINKED ISSUE #194: change all logger.trace to use block syntax
When doing logger.trace("xxx #{method.call}"), the string will get interpolated and evaluated, hence method.call executed, regardless of the enabled logging level. We should use the block syntax logger.trace { "xxx #{method.call}" } so method.call wouldn't execute when trace level is not enabled. It may only be an infinitesimal benefit, but it is the right thing to do, IMO. This makes our logging similar to or even better than Java's string expansion syntax of `logger.trace("xxx {}", var)`

I can do a PR for this if y'all agreed

jimtng:
I went through the library and only found these three instances where logging could be further optimised
Resolve #194


pacive (approved):
Looks good!

PR #209, opened 2021-03-02
feat(persistence): convert HistoricItem methods to directly return its state

### LINKED ISSUE #208: Persistence methods that return a HistoricItem should return its state
Most persistence methods return a DecimalType, but `maximumSince`, `minimumSince` and `historicState` returns a `HistoricItem` instead, so for these methods we need to add `.state` to the result in order for them to work correctly.

jimtng:
This resolves #208 

```ruby
midnight = ZonedDateTime.now.withHour(0).withMinute(0).withSecond(0)
max = Solar_AC_Power.maximum_since(midnight, :influxdb)
logger.error("Max since: #{max}, occurred at: #{max.timestamp}") # here we can use #{max} instead of #{max.state}
```

### DIFF
--- a/docs/usage/misc/persistence.md
+++ b/docs/usage/misc/persistence.md
@@ -11,31 +11,31 @@ grand_parent: Usage
 
 [Persistence](https://www.openhab.org/docs/configuration/persistence.html) functions can be accessed through the item object. The following methods related to persistence are available: 
 
-| Method            | Parameters                            | Example                                                         |
-| ----------------- | ------------------------------------- | --------------------------------------------------------------- |
-| `persist`         | service                               | `Item1.persist`                                                 |
-| `last_update`     | service                               | `Item1.last_update`                                             |
-| `previous_state`  | skip_equal: (default: false), service | `Item1.previous_state` `Item1.previous_state(skip_equal: true)` |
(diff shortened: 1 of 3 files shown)


pacive (commented):
Looks like it would do the job. Not sure how to make accurate tests though..

boc-tothefuture:
What is the testing issue?

jimtng:
> Looks like it would do the job. Not sure how to make accurate tests though..

We are not concerned about whether PersistenceExtensions do its job properly, i.e. return the correct average data. That responsibility lies within openhab-core. We are only concerned here as far as our cucumber tests go, that our api calls work without throwing an error.

jimtng:
Since I am leaving the `.state` accessible, how about comparing the two in our test, i.e.

```
max = item.maximum_since(midnight)
logger.info("Success: #{max == max.state}")
```

boc-tothefuture:
This looks good to me and the test you proposed looks fine too. 
@jimtng do you want to add the test and take it out of draft status?

jimtng:
@boc-tothefuture test and doc added. I'm not good with writing docs, so don't hesitate to suggest any improvements.

pacive (approved):

PR #214, opened 2021-03-03
fix: triggering on multiple items caused a stack overflow

### LINKED ISSUE #213: ver 3.0 Error: stack level too deep
When a rule specifies multiple trigger items, an error "stack level too deep" will occur.
```ruby
rule 'x' do
  changed Item1, Item2
  run { logger.info('x') }
end
```

jimtng:
Fix #213

### DIFF
--- a/features/changed.feature
+++ b/features/changed.feature
@@ -136,6 +136,46 @@ Feature:  changed
       | ON            | ON    | should not |
       | ON            | OFF   | should     |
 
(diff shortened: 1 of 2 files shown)


boc-tothefuture:
Do we need a test case for when it is explicitly an array [Item1, Item1]

?

jimtng:
> Do we need a test case for when it is explicitly an array [Item1, Item2]
> 
> ?

Good idea.

PR #218, opened 2021-03-03
feat(persistence): support TimeOfDay as the timestamp for persistence methods

jimtng:
As mentioned in #208 this PR supports this code:

```ruby
Solar_AC_Power.maximum_since(MIDNIGHT) #this uses our MIDNIGHT constant from time_of_day.rb
Solar_AC_Power.maximum_since('7am')
```

### DIFF
--- a/lib/openhab/dsl/monkey_patch/items/persistence.rb
+++ b/lib/openhab/dsl/monkey_patch/items/persistence.rb
@@ -66,11 +66,14 @@ module OpenHAB
           # @return [ZonedDateTime]
           #
           def to_zdt(timestamp)
-            if timestamp.is_a? Java::JavaTimeTemporal::TemporalAmount
-              logger.trace("Converting #{timestamp} (#{timestamp.class}) to ZonedDateTime")
-              Java::JavaTime::ZonedDateTime.now.minus(timestamp)
-            else
-              timestamp
+            logger.trace("Converting #{timestamp} (#{timestamp.class}) to ZonedDateTime")
+            return timestamp.to_zdt if timestamp.respond_to? :to_zdt
+
+            case timestamp
+            when ZonedDateTime then timestamp
+            when Java::JavaTimeTemporal::TemporalAmount then ZonedDateTime.now.minus(timestamp)
+            when String then OpenHAB::DSL::TimeOfDay::TimeOfDay.parse(timestamp).to_zdt
+            else raise ArgumentError, "Invalid timestamp: #{timestamp} (#{timestamp.class})"
             end
           end
 
--- a/docs/usage/misc/persistence.md
+++ b/docs/usage/misc/persistence.md
@@ -28,7 +28,7 @@ grand_parent: Usage
 | `updated_since`   | timestamp, service                    |                                                                 |
 | `variance_since`  | timestamp, service                    |                                                                 |
 
-* The `timestamp` parameter accepts a java ZonedDateTime or a [Duration](../duration/) object that specifies how far back in time.
+* The `timestamp` parameter accepts a java ZonedDateTime, a [Duration](../duration/) object that specifies how far back in time, or a [TimeOfDay](../time_of_day/) string or object that represents the time of day today. `TimeOfDay` constants such as `MIDNIGHT` and `NOON` can be used.
 * The `service` optional parameter accepts the name of the persistence service to use, as a String or Symbol. When not specified, the system's default persistence service will be used.
 * The return value of the following persistence methods is a [Quantity](../../items/number/#quantities) when the corresponding item is a dimensioned NumberItem:
   * `average_since`
@@ -50,7 +50,8 @@ Number:Power  Power_Usage "Power Usage [%.2f W]"
 logger.info("UV_Index Average: #{UV_Index.average_since(12.hours)}") 
 # Power_Usage has a Unit of Measurement, so 
 # Power_Usage.average_since will return a Quantity with the same unit
-logger.info("Power_Usage Average: #{Power_Usage.average_since(12.hours)}") 
+logger.info("Power_Usage Average: #{Power_Usage.average_since(MIDNIGHT)}") 
+logger.info("Power_Usage Average: #{Power_Usage.average_since('7am')}") 
 ```
 
 Comparison using Quantity
--- a/features/persistence.feature
+++ b/features/persistence.feature
@@ -40,7 +40,7 @@ Feature: persistence
         run do
           Number1.update 10
           sleep 1
-          persistence(:rrd4j) do
+          persistence(:mapdb) do
             Number1.persist
             Test_Group.persist
             logger.info("Last update: #{Number1.last_update}")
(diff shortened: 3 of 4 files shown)


pacive (commented):
It might make sense to add a to_zdt method to TimeOfDay, then you could change to:
```
when OpenHAB::DSL::TimeOfDay::TimeOfDay then datetime.to_zdt
            when String then OpenHAB::DSL::TimeOfDay::TimeOfDay.parse(timestamp).to_zdt
```
It would then be easy to also support passing a DateTimeItem as parameter by adding

```
return datetime.to_zdt if datetime.respond_to?(:to_zdt)
```
As a guard clause at the top (and remove `when TimeOfDay)

Edit: should have had this comment in the `to_zdt` method, hard to do this on the phone!

lib/openhab/dsl/monkey_patch/items/persistence.rb
```diff
@@ -65,14 +65,26 @@ def previous_state(service = nil, skip_equal: false)
...
+            case timestamp
+            when ZonedDateTime then timestamp
+            when Java::JavaTimeTemporal::TemporalAmount then ZonedDateTime.now.minus(timestamp)
+            when OpenHAB::DSL::TimeOfDay::TimeOfDay
+              ZonedDateTime.now
+                           .withHour(timestamp.hour)
+                           .withMinute(timestamp.minute)
+                           .withSecond(timestamp.second)
+                           .withNano(0)
+            when String then to_zdt(OpenHAB::DSL::TimeOfDay::TimeOfDay.parse(timestamp))
```
pacive, line 82: It might make sense to add a to_zdt method to TimeOfDay, then you could change to:
```ruby
when OpenHAB::DSL::TimeOfDay::TimeOfDay then datetime.to_zdt
            when String then OpenHAB::DSL::TimeOfDay::TimeOfDay.parse(timestamp).to_zdt
```
It would then be easy to also support passing a DateTimeItem as parameter by adding
```ruby
return datetime.to_zdt if datetime.respond_to?(:to_zdt)
```
As a guard clause at the top (and remove `when TimeOfDay)

jimtng: Great idea!

jimtng:
Should we:
- Document that persistence methods can also accept a DatetimeItem as the "timestamp" argument?
- Document TimeOfDay.to_zdt on the TimeOfDay's doc page?

pacive (approved):
Nice work!

boc-tothefuture:
Hey, good work in here. .
Before we merge, I wanted to discuss something I was thinking about which was (now that we are gem) bringing in some of the activesupport extensions (this is what drives ruby on rails).

In particular, would we want to bring in the time extensions from active support?
https://api.rubyonrails.org/classes/Time.html

That would change the above to be:
Time.midnight() or maybe Time.current.midnight()

I haven't played with the API to be sure. 

One concern with the above is that we are mixing TimeOfDay with an instant.  

What do you think?

pacive:
Haven't worked much with ActiveSupport, is it possible to just `require` some parts? Seems rather heavy to load the whole library just for some time extensions otherwise?

boc-tothefuture:
> Haven't worked much with ActiveSupport, is it possible to just `require` some parts? Seems rather heavy to load the whole library just for some time extensions otherwise?

Yeah, you can just pull in the parts you need.

boc-tothefuture:
@jimtng do you want to weigh in here? 
Trying to figure out if we want to continue doing stuff with zoneddatetime or we say ruby time is what we use with the extensions?  

This also probably impact #222

jimtng:
> Trying to figure out if we want to continue doing stuff with zoneddatetime or we say ruby time is what we use with the extensions?

Since ZonedDateTime is used almost everywhere inside openhab-core, it should still get first class support by our library, yet Ruby Time is sometimes more compact and easier to deal with, and I use it in some of my rules. I don't have enough knowledge / information about the ActiveSupport (time) class at the moment to make a useful suggestion either way. However, it does seem that I keep stumbling upon ActiveSupport stuff and it seems to have a lot of very handy stuff.

So if it's not too 'expensive', adding the support is fine. Are you proposing to replace/remove our TimeOfDay completely and use the ActiveSupport time class?

jimtng:
> One concern with the above is that we are mixing TimeOfDay with an instant.

Could you explain what you meant with this please?

boc-tothefuture:
> > One concern with the above is that we are mixing TimeOfDay with an instant.
> 
> Could you explain what you meant with this please?

TimeOfDay is "Local Time" in the java.time parlance.  The methods are expecting an instant (a specific time measured along some timeline that is unique).  We are mixing the two here and using them interchangeably.  

Which maybe is ok, I just wanted to point it out. If we went the active support route here instead it would be:
```
Solar_AC_Power.maximum_since(time.now.midnight)
Solar_AC_Power.maximum_since(time.yesterday.midnight)
```
So starting and staying with an instant.

If we don't see a problem with that, I am ok.  We can always add active support time modifiers later.

jimtng:
> > > One concern with the above is that we are mixing TimeOfDay with an instant.
> > 
> > 
> > Could you explain what you meant with this please?
> 
> TimeOfDay is "Local Time" in the java.time parlance. The methods are expecting an instant (a specific time measured along some timeline that is unique). We are mixing the two here and using them interchangeably.
> 
> Which maybe is ok, I just wanted to point it out. 

Thanks for the clarifications. Yes, I understand the difference, and also agree that it's a bit icky to mix the two.

> If we went the active support route here instead it would be:
> 
> ```
> Solar_AC_Power.maximum_since(time.now.midnight)
> Solar_AC_Power.maximum_since(time.yesterday.midnight)
> ```
> 
> So starting and staying with an instant.
> 
> If we don't see a problem with that, I am ok. We can always add active support time modifiers later.

I would much prefer to drop the support for TimeOfDay (hence this PR) and add the Activesupport time modifiers as above. It is clear which "midnight" we're referring to, not to mention having more options such as yesterday.midnight.

I noticed in the examples, the method zone= and [find_zone!](https://api.rubyonrails.org/classes/Time.html#method-c-find_zone-21) accept an argument of `-5.hours` in the example. Do they override the Numeric/Integer class, and will this conflict with our (java) duration?

pacive:
I think the problem that this should solve was to allow other better ways to specify the timestamp to the persistence methods, but IMO it doesn't have to be TimeOfDay.

jimtng:
> I am a bit confused, do we want to close this PR and open another one to investigate using activesupport time? Or do we want to continue to merge this one?

Yes, let's close this PR and open another one to add activesupport time. Could you tackle it at some point in the future, @boc-tothefuture ? I'm interested in exploring what other neat things are available too. 



> I think the problem that this should solve was to allow other better ways to specify the timestamp to the persistence methods, but IMO it doesn't have to be TimeOfDay.

Do you agree that we should just close this PR and add activesupport time?

pacive:
I'm fine with this being closed.

boc-tothefuture:
Closed per discussion

PR #219, opened 2021-03-03
feat: support more comparisons

### LINKED ISSUE #205: Some comparisons not working
When testing out the group item PR I stumbled across some comparisons that isn't currently supported:
```
Numeric <=> QuantityType/Quantity | Needs to be fixed in the Quantity class (Error about incompatible units)

String <=> DateTimeItem/DateTime | I can add this to the monkey patched String#<=> in a separate PR
Time <=> DateTimeItem/DateTime   | Can add a <=> method to the monkey patched Time class (same as above)

OnOffType == Item        |
OpenClosedType == Item   | These only support case equality, should == be added as well?
UpdownType == Item       |

SwitchItem == SwitchItem   | Only supports comparison with OnOffType - not other items
ContactItem == ContactItem | Only supports comparison with OpenClosedType - not other items
```

IMO, the last two is the most urgent to fix, I can see many use cases where you would want to compare the state of two different Switches/Contacts. The others are easier to work around by just changing the order of the operands.

We should also add more tests to verify any other comparisons wee need to support.

pacive:
Made some additions to make more comparisons work bi-directionally. Also moved some comparison tests into `comparisons.feature` to speed up the test runs a little.

Most notably made Switch and Contact Items comparable to each other.

**Question:** Do we want to support everything bi-directionally, or should we discourage Yoda-syntax such as
```ruby
CLOSED == ContactItem
```

Fixes #205 

### DIFF
--- a/lib/openhab/dsl/types/datetime.rb
+++ b/lib/openhab/dsl/types/datetime.rb
@@ -75,13 +75,13 @@ module OpenHAB
         # @return [Integer] -1, 0 or 1 depending on the outcome
         #
         def <=>(other)
+          if other.respond_to?(:zoned_date_time)
+            return zoned_date_time.to_instant.compare_to(other.zoned_date_time.to_instant)
+          end
+
           case other
-          when DateTime, DateTimeType, DateTimeItem
-            zoned_date_time.to_instant.compare_to(other.zoned_date_time.to_instant)
-          when TimeOfDay::TimeOfDay, TimeOfDay::TimeOfDayRangeElement
-            to_tod <=> other
-          when String
-            self <=> DateTime.parse(DATE_ONLY_REGEX =~ other ? "#{other}'T'00:00:00#{zone}" : other)
+          when TimeOfDay::TimeOfDay, TimeOfDay::TimeOfDayRangeElement then to_tod <=> other
+          when String then self <=> DateTime.parse(DATE_ONLY_REGEX =~ other ? "#{other}'T'00:00:00#{zone}" : other)
           else
             self <=> DateTime.from(other)
           end
--- a/features/comparisons.feature
+++ b/features/comparisons.feature
@@ -13,31 +13,25 @@ Feature: comparisons
       | ShuttersPos  | Rollershutter      | MAX      |              |
 
     And items:
-      | type               | name       | label         | group | state |
-      | Number             | Number5    | Number Five   |       | 5     |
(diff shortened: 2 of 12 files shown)


lib/openhab/dsl/types/datetime.rb
```diff
@@ -75,13 +75,13 @@ def initialize(datetime)
         # @return [Integer] -1, 0 or 1 depending on the outcome
         #
         def <=>(other)
+          if other.respond_to?(:zoned_date_time)
```
jimtng, line 78: If we changed this to `to_zdt`, it will include TimeOfDay by way of its to_zdt that is implemented in #218 - WDYT?

pacive: This was mostly to enable comparisons with DateTime, DateTimeItem and GroupItem (if it's a DateTime group), but can see if it can work with TimeOfDay this way as well!

jimtng: It depends on whether we want to ignore the date portion, so today at 7:00 is equal to ToD 7:00 and tomorrow at 7:00 will equal to ToD 7:00 too. If that's the case, then we don't want to use TimeOfDay.to_zdt because that will fill it with today's date.

jimtng:
> **Question:** Do we want to support everything bi-directionally, or should we discourage Yoda-syntax such as
> 
> ```ruby
> CLOSED == ContactItem
> ```

IMO we do need to support bi-directional comparisons to avoid surprises.

boc-tothefuture:
I agree that they should be bidirectional.  

FYI - IMO, this a feature not a fix for the commit.

lib/openhab/dsl/types/datetime.rb
pacive: I would think TimeOfDay would mostly (only?) be used with the intent that the represent a time in the current day, so don't think it would be a problem for most users. But might want to inform about that in the docs for the TimeOfDay#to_zdt-method.

pacive: Marking this as resolved since the other PR for implementing this is to be closed.

boc-tothefuture:
@pacive this looks good to me.  Could you update the commit message with more details about what is now supported?  Think about what you would want communicated in the change log.

pacive:
Added some lines to the commit message to describe the new features.

PR #221, opened 2021-03-03
feat(image): support for image items

boc-tothefuture:
Initial image item support. Support for updating images from a URL, a
Base 64 encoded image, and a byte array/mime type pairing.

Support is added to extract the mime type and image bytes from the image
item.


boc-tothefuture:
Closes #36

boc-tothefuture:
@pacive and @jimtng please take a look..  
This one override update (and none of the other items do).. But the image handling is weird (only command it accepts is refresh).  

I would be open to alternate structures/designs if you see something more fitting.

jimtng:
One thing came to mind, how about adding `file:` keyword?

boc-tothefuture:
> One thing came to mind, how about adding `file:` keyword?

Sure.  That would be easy. 


No issues on the different use of update here compared to the rest of the items?

jimtng:
If it feels awkward to call update with keyword arguments, perhaps add a load_from method?

boc-tothefuture:
Added the file option. 
@jimtng 
are you think
load_from_file
load_from_url
load_from_bytes

jimtng:
> Added the file option. 
> @jimtng 
> are you thinking?
> ```
> load_from_file
> load_from_url
> load_from_bytes
> ```

Are those better than load_file, load_url, load_ hmm not sure if bytes is a good name.

pacive:
I thing the reason it doesn't accept commands is that commands are supposed to indicate to OH/bindings that some action should be taken (like changing an external actuator) which wouldn't make sense for image items. Update OTOH just updates the internal state of the item, which is all that's needed here.

I haven't used any image items, so not really sure how they work, but this PR looks like it could be useful!

lib/openhab/dsl/items/image_item.rb
```diff
@@ -0,0 +1,157 @@
...
+
+        #
+        # Update image from file
+        #
+        # @param [String] mime_type of image
+        # @param [String] file location
+        #
+        #
+        def update_from_file(mime_type:, file:)
+          update_from_parts(mime_type: mime_type, bytes: IO.binread(file))
```
jimtng, line 105: Can the mime type be read/derived from the file itself, so that we only need to provide the file name? It would involve utilising some sort of image parsing library.

boc-tothefuture: Yes.  https://github.com/minad/mimemagic

boc-tothefuture:
I went with 
```
update_from_byte
update_from_file
update_from_url
```

Felt important to keep 'update' in the name so the user knows it is an update operation.

boc-tothefuture:
@pacive and @jimtng can you review this with changes?

lib/openhab/dsl/items/image_item.rb
```diff
@@ -0,0 +1,163 @@
...
+
+        #
+        # Update image from image at URL
+        #
+        # @param [String] uri location of image
+        #
+        #
+        def update_from_url(uri)
+          logger.debug("Downloading image from #{uri}")
+          response = Net::HTTP.get_response(URI.parse(uri))
```
pacive, line 71: Is the to_hash necessary here? Headers should be directly accessible on the response object: `response['content-type']`. Also `URI(uri)` should be enough.

lib/openhab/dsl/items/image_item.rb
```diff
@@ -0,0 +1,163 @@
...
+          mime_type = response.to_hash['content-type'].first
```
pacive, line 72: Is the to_hash necessary here? Headers should be directly accessible on the response object: `response['content-type']`. Also `URI(uri)` should be enough.

pacive (approved):
Looks good, great work!

lib/openhab/dsl/items/image_item.rb
```diff
@@ -0,0 +1,163 @@
...
+        end
+
+        #
+        # Get the bytes of the image
+        #
+        # @return [Array] Bytes that comprise the image
+        #
+        def bytes
+          state&.get_bytes
+        end
```
jimtng, line 107: Can these two use `def_delegators @state, :mime_type, :bytes` ?

lib/openhab/dsl/items/image_item.rb
```diff
@@ -0,0 +1,163 @@
...
+
+        #
+        # Update image from image bytes
+        #
+        # @param [String] mime_type of image
+        # @param [Object] bytes image data
+        #
+        #
+        def update_from_bytes(bytes, mime_type: nil)
```
jimtng, line 85: I've been wondering about the name "bytes" here. I wonder if it would be better called `update_from_memory`, or `update_from_binary` ?

lib/openhab/dsl/items/image_item.rb
```diff
@@ -0,0 +1,163 @@
...
+        #
+        # Update image from file
+        #
+        # @param [String] file location
+        # @param [String] mime_type of image
+        #
+        #
+        def update_from_file(file, mime_type: nil)
+          mime_type ||= detect_mime_type(file: file)
+          update_from_bytes(IO.binread(file), mime_type: mime_type)
```
jimtng, line 60: We should avoid reading the file twice (once inside inside detect_mime_type -> mime_from_file_contents)

```suggestion
          file_data = IO.binread(file)
          mime_type ||= (MimeMagic.by_path(file) || MimeMagic.by_magic(file_data))&.type
          update_from_bytes(file_data, mime_type: mime_type)
```

By this, we won't need detect_mime_type and mime_from_file_contents methods

lib/openhab/dsl/items/image_item.rb
boc-tothefuture: I think that would result in an exception if state is nil.

lib/openhab/dsl/items/image_item.rb
boc-tothefuture: memory is pretty vague.. not sure what binary means in this context either.. it could be update_from_byte_array?

pacive: I think update_from_bytes sound best.

jimtng: OK, keep it as update_from_bytes

lib/openhab/dsl/items/image_item.rb
```diff
@@ -0,0 +1,164 @@
...
+        # @param [String] file location
+        #
+        # @return [String] mime type if it can be detected, nil otherwise
+        #
+        def mime_from_file_contents(file:)
+          logger.trace("Detecting mime type from file (#{file} contents")
+          File.open(file) do |f|
+            MimeMagic.by_magic(f)&.type
+          end
+        end
```
jimtng, line 148: Nothing is calling / using these methods now, so they should be removed?

lib/openhab/dsl/items/image_item.rb
jimtng: So would #update, unless you forgot to use the safe-navigation operator &.

lib/openhab/dsl/items/image_item.rb
boc-tothefuture: removed

boc-tothefuture:
@jimtng  Thanks for the comments and code review.  This good now?

jimtng:
Just need to squash the commits, then all good.

PR #225, opened 2021-03-08
fix: handle native java exceptions in clean_backtrace

### LINKED ISSUE #220: error logging undefined method `backtrace_locations'
I was in the middle of troubleshooting an issue and created this file:

```ruby
require 'openhab'

    on_time = BackyardPlants_Watering_Tap.previous_state(:influxdb, skip_equal: true)
    logger.error(on_time.to_s)
```

The error logged was:
```
06:33:43.376 [ERROR] [ript.internal.ScriptEngineManagerImpl] - Error during evaluation of script 'file:/openhab/conf/automation/jsr223/ruby/personal/test.rb': org.jruby.embed.EvalFailedException: java.lang.IllegalArgumentException: Cannot format given Object as a Date
```

However when I put those two lines inside a rule:
```ruby
rule 'x' do
  on_start
  run do
    on_time = BackyardPlants_Watering_Tap.previous_state(:influxdb, skip_equal: true)
    logger.error(on_time.to_s)
  end
end
```

The error being logged was:
```
06:36:30.999 [ERROR] [ript.internal.ScriptEngineManagerImpl] - Error during evaluation of script 'file:/openhab/conf/automation/jsr223/ruby/personal/test.rb': org.jruby.embed.EvalFailedException: (NoMethodError) undefined method `backtrace_locations' for #<Java::JavaLang::IllegalArgumentException:0x42c7c72e>
In rule: x
<script>:9:in `<main>'
```

It seems that java exceptions don't have `backtrace_locations`? I also noticed that the exception message is lost here.

pacive:
Doesn't get as clean as ruby exceptions, but cleans out the jruby stack frames and just leaves the java ones.

Fixes #220 

### DIFF
--- a/features/rule_language.feature
+++ b/features/rule_language.feature
@@ -108,8 +108,10 @@ Feature: rule_language
     When I deploy the rules file
     Then It should log 'OpenHAB ready for rule processing' within 5 seconds
 
+  @log_level_changed
   Scenario: Errors in a run block is logged with stack trace and exits rule
-    Given code in a rules file
+    Given log level INFO
+    And code in a rules file
       """
       def test
         test2
@@ -133,4 +135,25 @@ Feature: rule_language
     And It should log "<script>:8:in `test'" within 5 seconds
     And It should log "<script>:17:in `block in <main>'" within 5 seconds
     And It should log "<script>:15:in `<main>'" within 5 seconds
-    And It should not log 'This one works!' within 10 seconds
+    But It should not log 'This one works!' within 10 seconds
+
+  @log_level_changed
+  Scenario: Native java exceptions are handled
+    Given log level INFO
+    And code in a rules file
+      """
+      def test
+        Java::JavaLang::Integer.parseInt('k')
+      end
+
+      rule 'test' do
+        on_start
+        run { test }
+      end
+      """
+    When I deploy the rules file
+    Then It should log 'For input string: "k" (Java::JavaLang::NumberFormatException)' within 5 seconds
+    And It should log 'In rule: test' within 5 seconds
+    And It should log "RUBY.test(<script>:8)" within 5 seconds
+    And It should log "RUBY.<main>(<script>:13)" within 5 seconds
+    And It should log "RUBY.<main>(<script>:11)" within 5 seconds
--- a/lib/openhab/log/logger.rb
+++ b/lib/openhab/log/logger.rb
@@ -22,8 +22,20 @@ module OpenHAB
       #
       # Regex for matching internal calls in a stack trace
       #
-      INTERNAL_CALL_REGEX = %r{(openhab-scripting-.*/lib)|(org/jruby/)}.freeze
+      INTERNAL_CALL_REGEX = %r{(openhab-scripting-.*/lib)|org[./]jruby}.freeze
       private_constant :INTERNAL_CALL_REGEX
+
+      #
+      # Regex for matching internal calls in a java stack trace
+      #
+      EXCLUDED_JAVA_PACKAGES = /jdk\.internal\.reflect|java\.lang\.reflect|org\.openhab|java\.lang\.Thread\.run/.freeze
+      private_constant :EXCLUDED_JAVA_PACKAGES
+
+      #
+      # Regex for matching internal calls in a java stack trace
+      #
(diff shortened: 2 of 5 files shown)


boc-tothefuture:
Is this something we can test?

pacive:
It should be possible by writing a rule that raises e.g. a `Java::JavaLang::Exception`. Ideally it should be tested by changing the logging level to INFO during the test. I'll see if I can add that to the cucumber step definitions. Is there some way to do cleanup after a scenario is run, to reset i to TRACE afterwards?

pacive:
Added a test, and a function to set log level in the cucumber tests. Also managed to clean up the java stack traces nicely (if log level is INFO or above).

lib/openhab/log/logger.rb
```diff
@@ -22,9 +22,21 @@ class Logger
...
       # Regex for matching internal calls in a stack trace
       #
-      INTERNAL_CALL_REGEX = %r{(openhab-scripting-.*/lib)|(org/jruby/)}.freeze
+      INTERNAL_CALL_REGEX = %r{(openhab-scripting-.*/lib)|org.jruby}.freeze
       private_constant :INTERNAL_CALL_REGEX
 
+      #
+      # Regex for matching internal calls in a java stack trace
+      #
+      EXCLUDED_JAVA_PACKAGES = /jdk.internal.reflect|java.lang.reflect|org.openhab|java.lang.Thread.run/.freeze
```
jimtng, line 31: Should the dots be escaped?
```suggestion
      EXCLUDED_JAVA_PACKAGES = /jdk\.internal\.reflect|java\.lang\.reflect|org\.openhab|java\.lang\.Thread\.run/.freeze
```

lib/openhab/log/logger.rb
```diff
@@ -60,8 +72,13 @@ def initialize(name)
       def clean_backtrace(error)
         return error if debug_enabled?
 
-        backtrace = error.backtrace_locations.reject { |line| INTERNAL_CALL_REGEX.match? line.to_s }
-        error.set_backtrace(backtrace.map(&:to_s))
+        if error.respond_to? :backtrace_locations
+          backtrace = error.backtrace_locations.reject { |line| INTERNAL_CALL_REGEX.match? line.to_s }
+          error.set_backtrace(backtrace.map(&:to_s))
```
jimtng, line 77:
```suggestion
          backtrace = error.backtrace_locations.map(&:to_s).reject { |line| INTERNAL_CALL_REGEX.match? line }
          error.set_backtrace(backtrace)
```

pacive: The `backtrace_locations` method doesn't return strings, but an Array of `Thread::Backtrace::Location`. However, the `set_backtrace` method requires an array of strings. See https://ruby-doc.org/core-2.5.8/Exception.html. Don't really understand the reason for it, but that's why it looks like it does.

jimtng: Yes, I'm just trying to avoid calling to_s twice here

pacive: Ah, missed that you inserted `map(&:to_s)` before the reject, that would work!

lib/openhab/log/logger.rb
```diff
@@ -22,9 +22,21 @@ class Logger
       #
       # Regex for matching internal calls in a stack trace
       #
-      INTERNAL_CALL_REGEX = %r{(openhab-scripting-.*/lib)|(org/jruby/)}.freeze
+      INTERNAL_CALL_REGEX = %r{(openhab-scripting-.*/lib)|org.jruby}.freeze
```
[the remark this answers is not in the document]
pacive: Here we actually want the `.` to match both `org.jruby` and `org/jruby` depending on if it's a ruby or java stack trace. Could replace it with `[./]` to be less greedy, but would increase the complexity somewhat.

features/rule_language.feature
```diff
@@ -133,4 +135,25 @@ Feature: rule_language
...
+    Given log level INFO
+    And code in a rules file
+      """
+      def test
+        Java::JavaLang::Integer.parseInt('k')
+      end
+
+      rule 'test' do
+        on_start
+        run { test }
```
boc-tothefuture, line 151: Just curious, did this not work in a run block?  Why call out to the method?

pacive: Just to make sure the method calls are included in the stack trace.

boc-tothefuture: makes sense.

jimtng:
LGTM. thanks for this PR

Issue #224, opened 2021-03-08
between resolves to the wrong method


jimtng:
See https://community.openhab.org/t/jruby-openhab-rules-system/110598/132

When `between` is called from inside an execution block (run/triggered/otherwise), it resolves to OpenHAB::DSL::Rules::RuleConfig::between (created by prop :between).

When `between` is called outside an execution block, it would resolve to TimeOfDay.between.

Compare:

```ruby
# This will say outside: OpenHAB::DSL::TimeOfDay::TimeOfDayRangeElement
logger.info "outside: #{between('10:00'..'18:00').begin.class}"

rule 'x' do
  on_start
  #this will say inside: String
  run { logger.info "inside: #{between('10:00'..'18:00').begin.class}" }
end
```

Out docs suggesting the use of between, which can be interpreted as being usable inside a rule execution block
https://boc-tothefuture.github.io/openhab-jruby/usage/misc/time_of_day/#between

I think the problem is our cucumber test were done outside a rule / execution block hence why it passed.

pacive:
There was some namespacing confusion when the TimeOfDay module was renamed from TOD to TimeOfDay. Could this be the cause for the regression?

boc-tothefuture:
They are stored as strings and converted on-demand because we don't want to pre-compute them in the rule because they get converted into two instants.  Inside of a rule and outside a rule config that would be OK. 

I am not actually sure why within the run context it is resolving to the rule_context.  

We store the actual caller of the rule block in the rule_config into a variable named @caller

We could try changing our run code from:
```
task.block.call(event)
```

to
```
@config.caller.instance_exec(event, task.block) do |event, block|
 block.call(event)
end
```
This should run the block in the context of the caller of the rule method.  
Obviously needs to be tested.  I am not sure if it would break any other assumptions.

boc-tothefuture:
Has anyone started a branch to look at this?

jimtng:
> Has anyone started a branch to look at this?

This is a bit beyond my current knowledge of Ruby

boc-tothefuture:
Ok, I will start one.

jimtng:
> ```
> @config.caller.instance_exec(event, task.block) do |event, block|
>  block.call(event)
> end
> ```
> 
> 
> This should run the block in the context of the caller of the rule method.
> 
> Obviously needs to be tested.  I am not sure if it would break any other assumptions.

[...]

PR #226, opened 2021-03-08
fix(scope): change execution block binding to be object that based a …

boc-tothefuture:
…block to rule

Addresses #224

### DIFF
--- a/features/between.feature
+++ b/features/between.feature
@@ -61,3 +61,21 @@ Feature: between
       | '<%=(Time.now - (5*60)).strftime('%H:%M:%S')%>'..'<%=(Time.now + (5*60)).strftime('%H:%M:%S')%>'  | '<%=(Time.now).strftime('%H:%M:%S')%>' | should     |
       | '<%=(Time.now + (5*60)).strftime('%H:%M:%S')%>'..'<%=(Time.now + (10*60)).strftime('%H:%M:%S')%>' | '<%=(Time.now).strftime('%H:%M:%S')%>' | should not |
 
+
+  Scenario Outline: Between ranges work inside of rule execution blocks
+    Given a rule template:
+      """
+      rule 'Testing Between Range' do
+        on_start
+        run do
+          range = between <between>
+          logger.info("Within time range") if range.cover? <compare>
+        end
+      end
+      """
+    When I deploy the rule
+    Then It <should> log "Within time range" within 5 seconds
+    Examples: Checks Time, TimeOfDay and Strings
+      | between                                                                                           | compare  | should     |
+      | '<%=(Time.now - (5*60)).strftime('%H:%M:%S')%>'..'<%=(Time.now + (5*60)).strftime('%H:%M:%S')%>'  | Time.now | should     |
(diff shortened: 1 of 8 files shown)


jimtng (approved):
looks good

jimtng:
Ooops forgot - please update the docs related to #{name}. These can be found in docs/examples/index.md and docs/usage/triggers/every.md

features/every.feature
```diff
@@ -7,9 +7,9 @@ Feature:  every
   Scenario: Using cron shortcuts ':minute'
     Given a rule
       """
-      rule 'Log the rule name every minute' do
+      log_rule = rule 'Log the rule name every minute' do
```
pacive, line 10: Could it be possible to pass the rule as a parameter to the rule block instead? Or would that not be available in the run-block's scope anyway?

```ruby
rule 'Log the rule name every minute' do |r|
  every :minute
  run { logger.info "Rule '#{r.name}' executed" }
end
```

pacive (commented):
Does this change the behavior of other variables defined outside of rule blocks in any way? Perhaps even for the better?

features/every.feature
boc-tothefuture: That would in fact be better and much more ruby like..  right now after I look at this it wouldn't work with 'on start'.  Let me see if this works.

features/between.feature
```diff
@@ -61,3 +61,21 @@ Feature: between
       | '<%=(Time.now - (5*60)).strftime('%H:%M:%S')%>'..'<%=(Time.now + (5*60)).strftime('%H:%M:%S')%>'  | '<%=(Time.now).strftime('%H:%M:%S')%>' | should     |
       | '<%=(Time.now + (5*60)).strftime('%H:%M:%S')%>'..'<%=(Time.now + (10*60)).strftime('%H:%M:%S')%>' | '<%=(Time.now).strftime('%H:%M:%S')%>' | should not |
 
+
+  Scenario Outline: Between ranges work inside of rule execution blocks
+    Given a rule template:
+      """
+      rule 'Testing Between Range' do
+        on_start
```
jimtng, line 69: Do we need to also test this on other types of trigger, at least a cron trigger?

boc-tothefuture: The every test is testing it with a cron trigger.

features/every.feature
boc-tothefuture: Updated..  
I am really glad we do code reviews.  this is much better.

features/between.feature
jimtng: I am referring to testing the between range. I'm not sure whether on_start runs in a different level of scope compared to the other triggers.

boc-tothefuture:
> Ooops forgot - please update the docs related to #{name}. These can be found in docs/examples/index.md and docs/usage/triggers/every.md

Looks like you already took care of this in another PR?

jimtng:
> > Ooops forgot - please update the docs related to #{name}. These can be found in docs/examples/index.md and docs/usage/triggers/every.md
> 
> Looks like you already took care of this in another PR?

I was just trying to be helpful and save you a few keystrokes. That PR is against the `run_context` branch (this PR). But obviously it can be discarded if we go with @pacive's suggestion which seems like a better way to go about it.

pacive (approved):
Looks good!

Issue #34, opened 2021-01-08
Add Color Item Type


boc-tothefuture:

ccutrer:
okay yeah, accepting a hex string as a command incredibly useful. excerpt:

in the rules DSL, this:
```
rule "Everlights preset pattern"
when
  Item Everlights_Preset changed
then
  Everlights_PatternChanging.postUpdate(ON)

  if (Everlights_Preset.state == "Valentine's Day") {
    Everlights_Count.postUpdate(6)
    Everlights1_RGB.postUpdate(HSBType.fromRGB(255, 0, 0))
    Everlights2_RGB.postUpdate(HSBType.fromRGB(255, 255, 255))
    Everlights3_RGB.postUpdate(HSBType.fromRGB(255, 0, 0))
    Everlights4_RGB.postUpdate(HSBType.fromRGB(252, 0, 255))
    Everlights5_RGB.postUpdate(HSBType.fromRGB(252, 0, 255))
    Everlights6_RGB.postUpdate(HSBType.fromRGB(252, 0, 255))
    Everlights_Fade.postUpdate(0)
    Everlights_Blink.postUpdate(0)
    Everlights_Strobe.postUpdate(0)
    Everlights_Chase.postUpdate(40)
  } else if (Everlights_Preset.state == "St. Patrick's Day") {
    Everlights_Count.postUpdate(3)
    Everlights1_RGB.postUpdate(HSBType.fromRGB(0, 255, 65))
    Everlights2_RGB.postUpdate(HSBType.fromRGB(0, 255, 65))
    Everlights3_RGB.postUpdate(HSBType.fromRGB(0, 255, 0))
    Everlights_Fade.postUpdate(0)
    Everlights_Blink.postUpdate(0)
    Everlights_Strobe.postUpdate(0)
    Everlights_Chase.postUpdate(40)
  } else if (Everlights_Preset.state == "Easter") {
    Everlights_Count.postUpdate(5)
    Everlights1_RGB.postUpdate(HSBType.fromRGB(253, 19, 255))
    Everlights2_RGB.postUpdate(HSBType.fromRGB(249, 255, 32))
    Everlights3_RGB.postUpdate(HSBType.fromRGB(255, 1, 65))
    Everlights4_RGB.postUpdate(HSBType.fromRGB(90, 255, 18))
    Everlights5_RGB.postUpdate(HSBType.fromRGB(41, 73, 255))
    Everlights_Fade.postUpdate(0)
[...]

PR #230, opened 2021-03-10
feat(location): support for location items

### LINKED ISSUE #37: Add Location Item type

boc-tothefuture:
Closes #37 

Take a look at the `def_oh_item_delegator` method. It is new.


docs/usage/items/location.md
```diff
@@ -0,0 +1,37 @@
...
+title: Location
+nav_order: 1
+has_children: false
+parent: Items
+grand_parent: Usage
+---
+
+
+# Location Item
+Player Items allow control of elements like audio players, televisions, etc. All methods of the underlying OpenHAB Player Item exist along with Ruby like method / extensions
```
jimtng, line 12: This needs to be updated

lib/openhab/dsl/items/location_item.rb
```diff
@@ -0,0 +1,45 @@
...
+        # Alias '-' to distance_from
+        alias :- distance_from
+
+        item_type Java::OrgOpenhabCoreLibraryItems::LocationItem
+
+        #
+        # Creates a new ImageItem
+        #
+        # @param [Java::OrgOpenhabCoreLibraryItems::ImageItem] image_item
+        #   The OpenHAB ImageItem to delegate to
```
jimtng, line 32:
```suggestion
        # Creates a new LocationItem
        #
        # @param [Java::OrgOpenhabCoreLibraryItems::LocationItem] location_item
        #   The OpenHAB LocationItem to delegate to
```

lib/openhab/dsl/items/location_item.rb
```diff
@@ -0,0 +1,45 @@
...
+        extend OpenHAB::DSL::Items::ItemCommand
+        extend OpenHAB::DSL::Items::ItemDelegate
+        extend Forwardable
+
+        def_item_delegator :@location_item
+
+        def_oh_item_delegator :@location_item, :distance_from
+
+        # Alias '-' to distance_from
+        alias :- distance_from
```
pacive, line 24: The mathematician in me doesn't really like this, since generally:
```
a - b != b - a unless a == b
```
And that rule breaks here.

lib/openhab/dsl/items/item_delegate.rb
```diff
@@ -28,6 +28,25 @@ def self.extended(base)
...
 
+        #
+        # Delegate the supplied methods to the supplied item
+        #
+        # @param [Object] item
+        # @param [<Type>] *methods <description>
+        #
+        # @return [<Type>] <description>
+        #
+        def def_oh_item_delegator(item, *methods)
```
pacive, line 39: Would this be useful for other Item types as well? It's actually quite inconsistent (at the OH-side) that this method is available on the Item, in most other cases you have to retrieve it's state before doing anything, so not sure this would be of any benefit anywhere else.

docs/usage/items/location.md
boc-tothefuture: oops.

lib/openhab/dsl/items/item_delegate.rb
boc-tothefuture: Not sure.. Are you suggesting I take it out and just override distance_from?

lib/openhab/dsl/items/location_item.rb
boc-tothefuture: I can see that.  It did seem to flow to think that you are substracing for the distance from two points.  

Do we want to take it out?  I need to think if in ruby there are other places like when you subtract arrays that don't abide by that rule.

pacive: I think it can be fine to leave it, it's unlikely anyone would use it with the intent of getting a negative number.

lib/openhab/dsl/items/item_delegate.rb
pacive: I generally like the way you have made these functions generic, but in this case I think it would be enough to override the method. This way it would also be possible to have comparisons with strings, like `LocationItem.distance_from('30, 50')`

lib/openhab/dsl/items/item_delegate.rb
```diff
@@ -41,6 +41,8 @@ def def_item_delegator(item, *methods)
             logger.trace("Creating delegate for (#{method}) to #{item}) for #{self.class}")
             def_delegator item, method
           end
+          # Provide uniform access to the underlying oh_item
+          def_delegator item, :java_object, :oh_item
```
pacive, line 45: Does this work for all item types with this addition, or do the other classes need to be adapted somehow?

pacive (approved):
Nice work!

lib/openhab/dsl/items/item_delegate.rb
boc-tothefuture: yes, should work for all delegated item types.

jimtng (commented):
Currently Location_Item - Point.new(xxx) works, but Point.new(xxx) - Location_Item won't work.

PointType also has its distance_from method, which we could also override to accept LocationItem and string.

boc-tothefuture:
Once #286 is merged, I will update this PR to match that style.  Additionally, I will change it to match the discussion of HSBTypes in #34 so that it will accept a hash rather than a CSV. 

Hash will look like:
```
{lat: 34.55, long: 234.23} 
{latitude: 34.55, longitude: 234.23 }
```

Issue #233, opened 2021-03-17
Hash from metadata bug


jimtng:
I've spent several days and many long hours trying to figure this out, but I'm still stuck. I hope one of you could help me out on this.

I cannot demonstrate this bug using cucumber because it wipes and recreates the items on each run, so I'll have to describe how to reproduce this bug manually:

Step 1:
Create a dummy item
```
Switch SwitchItem1
```

Step 2:
Create this test.rb file in <CONF>/automation/jsr332/ruby/personal (the usual directory)
```ruby
require 'openhab'

SwitchItem1.meta['test'] = { '1' => { 'moo' => 'cow' } }
logger.error("Meta: for test/1 #{SwitchItem1.meta.dig('test', '1')}")
logger.error("Meta: for test/1/moo '#{SwitchItem1.meta.dig('test', '1', 'moo')}'")
```

You should see the error log in openhab console like this:
```
03:00:07.971 [ERROR] [jsr223.jruby.Object                  ] - Meta: for test/1 {"moo"=>"cow"}
03:00:07.972 [ERROR] [jsr223.jruby.Object                  ] - Meta: for test/1/moo 'cow'
```

So far so good.

Note that at this point, SwitchItem1 has stored the metadata, so in theory, the data is still there even if openhab gets restarted.

Step 3:
Now comment out the line that assigns the metadata in test.rb above, and save it. The file should look like this:
```ruby
require 'openhab'

#SwitchItem1.meta['test'] = { '1' => { 'moo' => 'cow' } }
logger.error("Meta: for test/1 #{SwitchItem1.meta.dig('test', '1')}")
logger.error("Meta: for test/1/moo '#{SwitchItem1.meta.dig('test', '1', 'moo')}'")
```

This is where I'm stuck
```
03:03:11.884 [ERROR] [jsr223.jruby.Object                  ] - Meta: for test/1 {"moo"=>"cow"}
03:03:11.885 [ERROR] [jsr223.jruby.Object                  ] - Meta: for test/1/moo ''
```

Note the second log line (for `SwitchItem1.meta.dig('test', '1', 'moo')`) is blank '' even though `SwitchItem1.meta.dig('test', '1')` shows that it returns a hash `{"moo"=>"cow"}`

I have tried investigating this, although not down to the level of jruby's internals, but still can't figure it out.

It seems that the String used for the key is weird after it's read back from the metadata (java?) but I cannot explain why.

The only way I can get it to work is by doing this:
```ruby
[...]

PR #235, opened 2021-03-19
fix(metadata): metadata config key hash mismatched

jimtng:
See #233

### DIFF
--- a/lib/openhab/dsl/monkey_patch/items/metadata.rb
+++ b/lib/openhab/dsl/monkey_patch/items/metadata.rb
@@ -30,7 +30,7 @@ module OpenHAB
 
             def initialize(metadata: nil, key: nil, value: nil, config: nil)
               @metadata = metadata || Metadata.new(key || MetadataKey.new('', ''), value&.to_s, config)
-              super(@metadata&.configuration)
+              super(to_ruby(@metadata&.configuration))
             end
 
             #
@@ -84,6 +84,44 @@ module OpenHAB
             #
             def to_a
               [@metadata&.value, @metadata&.configuration || {}]
+            end
+
+            private
+
+            #
+            # Recursively convert the supplied Hash object into a Ruby Hash and recreate the keys and values
+            #
+            # @param [Hash] Hash to convert
+            #
+            # @return [Hash] The converted hash
+            #
+            def to_ruby_hash(hash)
+              return unless hash.respond_to? :each_with_object
+
+              hash.each_with_object({}) { |(key, value), ruby_hash| ruby_hash[to_ruby(key)] = to_ruby(value) }
+            end
+
+            #
+            # Recursively convert the supplied array to a Ruby array and recreate all String values
+            #
+            # @param [Object] array to convert
+            #
+            # @return [Array] The converted array
+            #
+            def to_ruby_array(array)
+              return unless array.respond_to? :each_with_object
+
+              array.each_with_object([]) { |value, ruby_array| ruby_array << to_ruby(value) }
+            end
+
+            # Convert the given object to Ruby equivalent
+            def to_ruby(value)
+              case value
+              when Hash, Java::JavaUtil::Map then to_ruby_hash(value)
+              when Array, Java::JavaUtil::List then to_ruby_array(value)
+              when String then String.new(value)
+              else value
+              end
             end
           end
 
--- a/features/metadata_hash.feature
+++ b/features/metadata_hash.feature
@@ -1,0 +1,87 @@
+Feature:  metadata_hash
+  Items support accessing metadata from different files
+
+  Background:
+    Given Clean OpenHAB with latest Ruby Libraries
+    And items:
+      | type   | name       |
+      | Switch | TestSwitch |
+    And a deployed rule:
+      """
+      TestSwitch.meta['animalsounds'] = { 'land' => { 'cow' => 'moo'}}
+      logger.info("Metadata animalsounds created for TestSwitch: #{TestSwitch.meta['animalsounds']}")
+      """
+
+  Scenario Outline: Metadata key works when assigned in the same rule
+    Given code in a rules file:
+      """
+      TestSwitch.meta['animalsounds'] = {"land"=>{"cow"=>"moo"}}
+      code = '<metadata>'
+      logger.info("#{code}: #{<metadata>}")
+      """
+    When I deploy the rules file
+    Then It should log '<metadata>: <value>' within 5 seconds
+    Examples:
+      | metadata                                           | value                    |
+      | TestSwitch.meta["animalsounds"]                    | {"land"=>{"cow"=>"moo"}} |
+      | TestSwitch.meta.dig("animalsounds", "land")        | {"cow"=>"moo"}           |
+      | TestSwitch.meta.dig("animalsounds", "land", "cow") | moo                      |
+
+
+  Scenario Outline: Metadata key works when not assigned in the same rule
+    Given code in a rules file:
+      """
+      code = '<metadata>'
+      logger.info("#{code}: #{<metadata>}")
+      """
+    When I deploy the rules file
+    Then It should log '<metadata>: <value>' within 5 seconds
+    Examples:
+      | metadata                                           | value                    |
+      | TestSwitch.meta["animalsounds"]                    | {"land"=>{"cow"=>"moo"}} |
+      | TestSwitch.meta.dig("animalsounds", "land")        | {"cow"=>"moo"}           |
+      | TestSwitch.meta.dig("animalsounds", "land", "cow") | moo                      |
+
+  Scenario Outline: Comparing against a hash literal when metadata assigned in the same rule
+    Given code in a rules file:
+      """
+      TestSwitch.meta['animalsounds'] = {"land"=>{"cow"=>"moo"}}
+      logger.info("Metadata checks out") if <metadata> == <value>
+      """
+    When I deploy the rules file
+    Then It should log 'Metadata checks out' within 5 seconds
+    Examples:
+      | metadata                                           | value                    |
+      | TestSwitch.meta["animalsounds"]                    | {"land"=>{"cow"=>"moo"}} |
+      | TestSwitch.meta.dig("animalsounds", "land")        | {"cow"=>"moo"}           |
+      | TestSwitch.meta.dig("animalsounds", "land", "cow") | "moo"                    |
+
+  Scenario Outline: Comparing against a hash literal when metadata NOT assigned in the same rule
+    Given code in a rules file:
+      """
+      logger.info("Metadata checks out") if <metadata> == <value>
+      """
+    When I deploy the rules file
+    Then It should log 'Metadata checks out' within 5 seconds
+    Examples:
+      | metadata                                           | value                    |
+      | TestSwitch.meta["animalsounds"]                    | {"land"=>{"cow"=>"moo"}} |
+      | TestSwitch.meta.dig("animalsounds", "land")        | {"cow"=>"moo"}           |
+      | TestSwitch.meta.dig("animalsounds", "land", "cow") | "moo"                    |
+
+  Scenario Outline: Calling to_json on metadata config
+    Given a deployed rule:
+      """
+      TestSwitch.meta['animalsounds'] = { 'land' => <value> }
+      """
+    And code in a rules file:
+      """
+      require 'json'
+      logger.info("Metadata to_json: #{TestSwitch.meta.dig("animalsounds", "land").to_json}")
+      """
+    When I deploy the rules file
+    Then It should log 'Metadata to_json: <json>' within 5 seconds
+    Examples:
+      | value                             | json                            |
+      | {'cow'=>'moo'}                    | {"cow":"moo"}                   |
+      | [{'cow'=>'moo'},{'sheep'=>'baa'}] | [{"cow":"moo"},{"sheep":"baa"}] |


boc-tothefuture:
Take a look at the WIP here, it now passes the test cases, but needs to be cleaned up
 1. Not sure it hits all corner cases
 2. There is probably a better way to recursively convert to a java hash map 


The Metadata config accepts object, and we were passing in a ruby hash.
This creates, i think two problems.
1. Not sure what any other system (jython, the core, etc) would do when trying to access the ruby hash object
2. Had the problem where the objects were of different instances in different evaluation contexts, leading to the errors we saw.

I think anything we push into the core should probably be a Java type and not a Jruby type.  Jruby is doing this under the covers when we call any methods that accept 'String', etc.  But since this accepted object it just persisted the JRuby Hash.

As I said, we need to clean this up and document the methods.. but WDYT @jimtng and @pacive ?

jimtng (commented):
Great work! It seems to have solved the problem for me

jimtng:
Shouldn't JRuby itself convert Ruby Hash to Java HashMap automatically when passing a Ruby Hash to Java?

jimtng:
There are other places where a Ruby Hash for configuration goes into Metadata.new(.....), e.g. inside `MetadataItem` class

lib/openhab/dsl/monkey_patch/items/metadata.rb
```diff
@@ -72,7 +72,7 @@ def value=(value)
             def config=(config)
               raise ArgumentError, 'Configuration must be a hash' unless config.is_a? Hash
 
-              metadata = Metadata.new(@metadata&.uID, @metadata&.value, config)
+              metadata = Metadata.new(@metadata&.uID, @metadata&.value, to_java_hash(config))
```
pacive, line 75: to_java_map

pacive (approved):
Nice work!

lib/openhab/dsl/monkey_patch/items/metadata.rb
```diff
@@ -238,6 +243,28 @@ def update_from_value(value)
...
 
+            #
+            # Convert the supplied hash to a java.util.HashMap and all keys and values to java.lang.string objects
+            #
+            # @param [Hash] hash to convert
+            # @param [Java::JavaUtil::HashMap] java_map to put converted objects into
+            #
+            # @return [Java::JavaUtil::HashMap] java_map container strings and other Java HashMaps
+            #
+            def to_java_map(hash, java_map = Java::JavaUtil::HashMap.new)
```
pacive, line 254: Should this be made a class method so it can be used from the MetadataItem class as well? I guess that's why the tests are failing.

boc-tothefuture: done

boc-tothefuture:
@jimtng I didn't feel comfortable squashing your first commit, would you mind squashing mine with yours?

jimtng (commented):
> @jimtng I didn't feel comfortable squashing your first commit, would you mind squashing mine with yours?

I have never done this type of operation before. It seems that if I were to do what you suggested: I would have to first reset my branch to your commit, then squash it, and removing the record that you committed it. Does that sound right (in terms of what needed to be done)?

I have no problem if you could/would squash it from your end.

jimtng:
BTW I'm still getting an NPE from to_json. Let me see if I can put that into the cucumber test.

boc-tothefuture:
> > @jimtng I didn't feel comfortable squashing your first commit, would you mind squashing mine with yours?
> 
> I have never done this type of operation before. It seems that if I were to do what you suggested: I would have to first reset my branch to your commit, then squash it, and removing the record that you committed it. Does that sound right (in terms of what needed to be done)?
> 
> I have no problem if you could/would squash it from your end.

Ok, I will give it a shot.

boc-tothefuture:
> BTW I'm still getting an NPE from to_json. Let me see if I can put that into the cucumber test.

Thanks.. please add.

jimtng:
I might have been abusing the metadata feature here, however:

```ruby
require 'openhab'
require 'json'
v = { 'key' => 'value' }
SwitchItem1.meta['test'] = v
logger.error("v.to_json '#{v.to_json}'")
logger.error("Meta to_json '#{SwitchItem1.meta['test'].to_json}'")
```

output:
```
21:48:44.309 [ERROR] [jsr223.jruby.Object                  ] - v.to_json '{"key":"value"}'
21:48:44.312 [ERROR] [jsr223.jruby.Object                  ] - Meta to_json '"{\"key\"=>\"value\"}"'
```

I'm still trying to create a simple test case for triggering the NPE

boc-tothefuture:
@jimtng - once we convert it to a java hashmap I am not sure if to_json will actually work?

jimtng:
Should we monkey patch Java HashMap and add #to_h to convert it to ruby hash when needed? For most cases, it isn't necessary because it seems to have the same enumerable stuff.

I had a look at jython helper library's `get_metadata` and `set_metadata`. They don't do any special conversion from python `dict` to Java HashMap and back, yet it seems to work just fine (I haven't tested it). Perhaps it's done by Jython under the hood whenever a dict is passed to Java. Or perhaps dict itself is implemented as a HashMap hence they are interchangeable.

pacive:
> Should we monkey patch Java HashMap and add #to_h to convert it to ruby hash when needed? For most cases, it isn't necessary because it seems to have the same enumerable stuff.
> 
> I had a look at jython helper library's `get_metadata` and `set_metadata`. They don't do any special conversion from python `dict` to Java HashMap and back, ~yet it seems to work just fine (I haven't tested it). Perhaps it's done by Jython under the hood whenever a dict is passed to Java. Or perhaps dict itself is implemented as a HashMap hence they are interchangeable.~ I take that back. I just tried writing some dict to a metadata config value in python, and pulling it out from jruby, the class is 'Java::OrgPythonCore::PyDictionary'.

From looking at the jython source code it seems the dicts are just backed by a java HashMap though, and doesn't do any hashing of the keys like jruby does (and in the RubyString#strHashCode method they use a hash seed retrieved from the current runtime), so they wouldn't have this problem.

boc-tothefuture:
> Should we monkey patch Java HashMap and add #to_h to convert it to ruby hash when needed? For most cases, it isn't necessary because it seems to have the same enumerable stuff.
> 
> I had a look at jython helper library's `get_metadata` and `set_metadata`. They don't do any special conversion from python `dict` to Java HashMap and back, ~yet it seems to work just fine (I haven't tested it). Perhaps it's done by Jython under the hood whenever a dict is passed to Java. Or perhaps dict itself is implemented as a HashMap hence they are interchangeable.~ I take that back. I just tried writing some dict to a metadata config value in python, and pulling it out from jruby, the class is 'Java::OrgPythonCore::PyDictionary'.

Ok, so it would seem that you might have unexpected interoperability with the python uses.

boc-tothefuture:
> Should we monkey patch Java HashMap and add #to_h to convert it to ruby hash when needed? For most cases, it isn't necessary because it seems to have the same enumerable stuff.
> 
Is this the only place we are exposed to Java HashMaps?

jimtng:
> Is this the only place the users are exposed to Java HashMaps?

At the moment yes. So we could just implement it inside `OpenHAB::DSL::MonkeyPatch::Items::Metadata::MetadataItem` 

I am still trying to understand the crux of the matter. At the moment it still points to ruby's string hash calculation. 

The current solution in this PR doesn't fully solve all the possible use cases that I'm doing with metadata.

jimtng:
sorry, closed by accident

boc-tothefuture:
Did you want to see if this fixes it?
https://javadoc.io/static/org.jruby/jruby-core/9.2.9.0/org/jruby/RubyInstanceConfig.html#CONSISTENT_HASHING_ENABLED

pacive:
> Did you want to see if this fixes it?
> https://javadoc.io/static/org.jruby/jruby-core/9.2.9.0/org/jruby/RubyInstanceConfig.html#CONSISTENT_HASHING_ENABLED

I think that would be a good solution.

boc-tothefuture:
@jimtng and @pacive do either of you want to give it a try? Looks like it can be set by a system property.

jimtng:
> @jimtng and @pacive do either of you want to give it a try? Looks like it can be set by a system property.

I'd love to try but don't know how to do that. Is that done in the jruby addon code then recompile?

boc-tothefuture:
> > @jimtng and @pacive do either of you want to give it a try? Looks like it can be set by a system property.
> 
> I'd love to try but don't know how to do that. Is that done in the jruby addon code then recompile?

```
public static final boolean CONSISTENT_HASHING_ENABLED
Whether to calculate consistent hashes across JVM instances, or to ensure un-predicatable hash values using SecureRandom. Set with the jruby.consistent.hashing.enabled system property.
```
It says it is java system property.  So you can just set it as you would any other java property within the startup script for OpenHAB.
I think it is the EXTRA_JAVA_OPTS in the script.

jimtng:
> It says it is java system property. So you can just set it as you would any other java property within the startup script for OpenHAB.
> I think it is the EXTRA_JAVA_OPTS in the script.

Is it like this:
```
      EXTRA_JAVA_OPTS: "-Djruby.consistent.hashing.enabled=true"
```
Is there a way to verify that it is in effect?

boc-tothefuture:
> > It says it is java system property. So you can just set it as you would any other java property within the startup script for OpenHAB.
> > I think it is the EXTRA_JAVA_OPTS in the script.
> 
> Is it like this:
> 
> ```
>       EXTRA_JAVA_OPTS: "-Djruby.consistent.hashing.enabled=true"
> ```
> 
> Is there a way to verify that it is in effect?

Some more details here:  https://community.openhab.org/t/setting-the-extra-java-opts-on-oh2/24061/2

I think it is more like this:
`EXTRA_JAVA_OPTS="-Djruby.consistent.hashing.enabled=true"`

jimtng:
> I think it is more like this:
> `EXTRA_JAVA_OPTS="-Djruby.consistent.hashing.enabled=true"`

My syntax was for docker, so it was correct. I realised to test it, simply create a file and call logger.info("Hash #{'x'.hash}") and then make sure it gets reloaded. Upon successive reloads, the hash value should remain the same. In my case it doesn't stay the same, so the setting didn't take effect.

jimtng:
I've submitted a new solution - instead of converting things when saving the metadata, we convert things when reading the metadata, because that's when the data is going to be used / compared / converted to json, etc. 

This solves two issues: 
* the string hash difference - solved by recreating the string in `restring`. The re-stringing may be avoidable if we could somehow enable consistent hashing. However, this would need to be enabled in the binding, so as not to rely on users to set the setting. 
* the returned data is a plain ruby hash/array that can be used by to_json

I left the commits un-squashed so we can see the history, potentially reverting this commit. If accepted as the solution, I can squash the commits.

boc-tothefuture:
> I've submitted a new solution - instead of converting things when saving the metadata, we convert things when reading the metadata, because that's when the data is going to be used / compared / converted to json, etc.
> 
> This solves two issues:
> 
> * the string hash difference - solved by recreating the string in `restring`. The re-stringing may be avoidable if we could somehow enable consistent hashing. However, this would need to be enabled in the binding, so as not to rely on users to set the setting.
> * the returned data is a plain ruby hash/array that can be used by to_json
> 
> I left the commits un-squashed so we can see the history, potentially reverting this commit. If accepted as the solution, I can squash the commits.

This solution seems better.

Do you know why you can't enable consistent hashing? All I would do in the binding is set the same property you are setting in the command line?  Are you sure it is being set?  You should be able to dump System.properties from a rule and see if it is set?

jimtng:
> Do you know why you can't enable consistent hashing? 

Most likely I'm not setting it correctly. 


```ruby
logger.error("jruby.consistent.hashing.enabled = #{Java::JavaLang::System.properties['jruby.consistent.hashing.enabled']}")
logger.error("OrgJruby::RubyInstanceConfig::CONSISTENT_HASHING_ENABLED = #{Java::OrgJruby::RubyInstanceConfig::CONSISTENT_HASHING_ENABLED}")
```

Output:
```
00:17:50.518 [ERROR] [jsr223.jruby.Object                  ] - jruby.consistent.hashing.enabled = true
00:17:50.520 [ERROR] [jsr223.jruby.Object                  ] - OrgJruby::RubyInstanceConfig::CONSISTENT_HASHING_ENABLED = false
```

boc-tothefuture:
> > Do you know why you can't enable consistent hashing?
> 
> Most likely I'm not setting it correctly.
> 
> ```ruby
> logger.error("jruby.consistent.hashing.enabled = #{Java::JavaLang::System.properties['jruby.consistent.hashing.enabled']}")
> logger.error("OrgJruby::RubyInstanceConfig::CONSISTENT_HASHING_ENABLED = #{Java::OrgJruby::RubyInstanceConfig::CONSISTENT_HASHING_ENABLED}")
> ```
> 
> Output:
> 
> ```
> 00:17:50.518 [ERROR] [jsr223.jruby.Object                  ] - jruby.consistent.hashing.enabled = true
> 00:17:50.520 [ERROR] [jsr223.jruby.Object                  ] - OrgJruby::RubyInstanceConfig::CONSISTENT_HASHING_ENABLED = false
> ```

Took a look at the code, it is just
`jruby.consistent.hashing` not `jruby.consistent.hashing.enabled`.

I verified if you set this the OrgJruby::RubyInstanceConfig::CONSISTENT_HASHING_ENABLED gets set correctly.

jimtng:
> Took a look at the code, it is just
> `jruby.consistent.hashing` not `jruby.consistent.hashing.enabled`.

Aha! Yes confirmed - and once it's enabled, hash values became consistent, as it said. Also confirmed that the re-stringing of the key is not necessary when consistent hashing is enabled.

So the question is do we want to have it enabled? If we do, I propose that it's set within the binding, or at the very least, configurable through jruby.cfg - with the default of it being enabled. And if we do, I will remove the `restring` from this PR.

jimtng:
> And if we do, I will remove the `restring` from this PR.

I spoke too soon. When the keys/values aren't going through restring, to_json caused an NPE.

boc-tothefuture:
> > And if we do, I will remove the `restring` from this PR.
> 
> I spoke too soon. When the keys/values aren't going through restring, to_json caused an NPE. So restring is still needed, at least for now, which means it doesn't matter whether the consistent hashing is enabled, this PR will still work.

does string.dup work too?

jimtng:
> does string.dup work too?

no to_json still caused an NPE. 

Here's the stack trace that I can see:

```
22:51:29.197 [ERROR] [y.OpenHAB::DSL::Rules::AutomationRule] -  (Java::JavaLang::NullPointerException)
In rule: PhotoFrame: Load Next
json.ext.Generator$6.generate(Generator.java:386)
json.ext.Generator$6.generate(Generator.java:372)
json.ext.Generator$Handler.generateNew(Generator.java:175)
json.ext.Generator.generateJson(Generator.java:33)
json.ext.GeneratorMethods$RbString.to_json(GeneratorMethods.java:105)
json.ext.GeneratorMethods$RbString$INVOKER$s$0$0$to_json.call(GeneratorMethods$RbString$INVOKER$s$0$0$to_json.gen)
json.ext.Generator$8.generateNew(Generator.java:432)
json.ext.Generator$8.generate(Generator.java:443)
json.ext.Generator$5$1.visit(Generator.java:358)
json.ext.Generator$5.generate(Generator.java:339)
json.ext.Generator$5.generate(Generator.java:310)
json.ext.Generator$4.generate(Generator.java:296)
json.ext.Generator$4.generate(Generator.java:253)
json.ext.Generator$Handler.generateNew(Generator.java:175)
json.ext.Generator.generateJson(Generator.java:33)
json.ext.GeneratorMethods$RbArray.to_json(GeneratorMethods.java:79)
json.ext.GeneratorMethods$RbArray$INVOKER$s$0$0$to_json.call(GeneratorMethods$RbArray$INVOKER$s$0$0$to_json.gen)
java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
```

jimtng:
OK I figured out how I got an NPE but couldn't reproduce it using a cucumber test. I was running OpenHAB 3.1M2 (and subsequently tried snapshot builds too with similar result).

When I downgraded to OpenHAB 3.0.1, the NPE is gone even when I didn't call restring. So it seems to be caused by some changes in Openhab core, but one that affects to_json somehow.

This is the code that would cause an NPE on OH 3.1 but not on OH 3.0.x
```ruby
# SwitchItem1.meta['test'] = { '1' => { 'moo' => 'cow' }  } # this is the content of the metadata
logger.error("Meta to json'#{SwitchItem1.meta.dig('test', '1').to_json}'")
```

Note that the NPE goes away if I do call restring.

boc-tothefuture:
Ok, so is this one ready to merge then?

jimtng:
> Ok, so is this one ready to merge then?
> 

Yes

boc-tothefuture:
Can you update to pass the tests and I will merge?

jimtng:
What caused the rubocop failure?

jimtng:
I've tidied up the code a bit since the last commit

jimtng:
Yikes! It seems that our tests are now hitting the same issue as #236

boc-tothefuture (approved):

PR #238, opened 2021-04-02
fix(dependency): Update mimemagic dependency

boc-tothefuture:


jimtng:
We should follow Rails and change mimemagic to [marcel](https://github.com/rails/marcel). Therefore I propose updating this PR and use marcel

boc-tothefuture:
Happy to move to that.  I was just trying to get the build to work with the new version (which it should) while we evaluate options.

jimtng:
I've been trying to solve a different mystery the past few days, and just came across `marcel` a few minutes before you posted this PR. It seems to be quite a straight forward replacement, so we might as well change over right away.

boc-tothefuture:
@jimtng and @pacive please review.

pacive (approved):
Great! Are there any steps required to update when the new version is released? (I assume it does everything automatically?)

boc-tothefuture:
It should be a non-breaking change?  It is using a different database for mime types so there might be some differences, but I doubt it would impact our users.

jimtng (approved):
Thank you @boc-tothefuture 
LGTM

PR #242, opened 2021-05-18
Add oh_item to decorated items and fix store_states

### LINKED ISSUE #241: store_states failed on decorated items
OpenHAB::DSL::States::store_states calls BusEvent.storeStates which expects OrgOpenhabCoreItemsItem object. This means passing our decorated items won't work.

It appears that we haven't got around to standardising on jruby item::oh_item to make things like store_states able to access the underlying openhab object.

What's the best way of solving this?

jimtng:
This PR contains two separate commits, which when combined, will resolve #241

The first commit adds oh_item method to decorated items through def_item_delegator, so that all decorated item classes will have oh_item method that returns the underlying openhab object.

This in turn is utilised by the second commit to fix the problem with store_states.

### DIFF
--- a/features/items.feature
+++ b/features/items.feature
@@ -203,3 +203,38 @@ Feature:  items
       """
     When I deploy the rules file
     Then It should log "Number3 is in groups Numbers, PrimeNumbers" within 5 seconds
+
+  Scenario: Verify that all item decorators provide oh_item to refer to the OpenHAB object
+    Given code in a rules file
+      """
+      decorators = OpenHAB::DSL::Items.constants
+        .select { |m| m.to_s.end_with?('Item') }
+        .reject { |m| m.to_s == 'GroupItem' }
+      decorators.each do |m|
+        item_name = "#{m}1"
+        oh_item = Kernel.const_get("Java::OrgOpenhabCoreLibraryItems::#{m}").new(item_name)
+        $ir.add(oh_item)
+        logger.info("OH Object: #{oh_item.class} name: #{oh_item.name}")
+        item = items[item_name]
+        logger.info("DSL Object: #{item.class}")
+        unless item&.respond_to?(:oh_item)
+          logger.error("missing oh_item for item type: #{m}. Make sure def_item_delegator is called")
+        end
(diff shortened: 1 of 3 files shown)


boc-tothefuture:
Can you rebase off master I fixed the testing?

boc-tothefuture:
Is this ready to merge?  it is marked as draft?  It might be good to add a test just to make sure we add oh_item to the decorated items (in case we ever have a regression and forget why we put it here :-) )

jimtng:
> Is this ready to merge? it is marked as draft.

Would you prefer this solution to #243 ? I can add tests to either PR.

jimtng:
I've been thinking about how to best create the test. I can enumerate the modules inside OpenHAB::Core::DSL, and filter for only classes ending with 'Item'. I get a list of all our decorator classes, provided that the naming remains consistent, i.e. ending with 'Item'. This will make the test routine work on all future decorators, without having to manually maintain a list of item types, which will be tedious and error prone.

However, I am stuck on the next step: how to create an item of that type within Openhab itself. 

The cucumber support routines have the Rest class along with the Item struct which is used by the "Given items" clause. My next step is to try to use this within my test code (the part within the triple double quotes). 

If you can think of a better way, please let me know.

jimtng:
I've written the test

boc-tothefuture:
> I've been thinking about how to best create the test. I can enumerate the modules inside OpenHAB::Core::DSL, and filter for only classes ending with 'Item'. I get a list of all our decorator classes, provided that the naming remains consistent, i.e. ending with 'Item'. This will make the test routine work on all future decorators, without having to manually maintain a list of item types, which will be tedious and error prone.
> 
> However, I am stuck on the next step: how to create an item of that type within Openhab itself.
> 
> The cucumber support routines have the Rest class along with the Item struct which is used by the "Given items" clause. My next step is to try to use this within my test code (the part within the triple double quotes).
> 
> If you can think of a better way, please let me know.

i was just thinking we could test for a method existing in the delegate that we didnt' define..  But your test appears much more thorough.

PR #243, opened 2021-05-18
refactor: rename the openhab item to oh_item in the wrappers

jimtng:
An alternative to #242. If you know of a better way, feel free to discard this PR

### DIFF
--- a/lib/openhab/dsl/items/datetime_item.rb
+++ b/lib/openhab/dsl/items/datetime_item.rb
@@ -21,7 +21,8 @@ module OpenHAB
         extend OpenHAB::DSL::Items::ItemCommand
         include Comparable
 
-        def_item_delegator :@datetime_item
+        def_item_delegator :@oh_item
+        attr_reader :oh_item
 
         item_type Java::OrgOpenhabCoreLibraryItems::DateTimeItem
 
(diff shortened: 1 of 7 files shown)


boc-tothefuture:
Hey @jimtng  - I am afraid I am not following.  This is instead of #242?

boc-tothefuture:
Can you rebase off master? I fixed the testing.

jimtng:
> Hey @jimtng - I am afraid I am not following. This is instead of #242?

Correct, this is an alternative solution to having oh_item instead of #242 
I think I prefer #242 but each has its pros and cons. WDYT?

#242 also includes the fix for store_states but I can make a separate PR for that if necessary.

boc-tothefuture:
It certainly seems much easier than 243, so I would go with this one unless you have a reason to prefer the other?

jimtng:
> It certainly seems much easier than 243, so I would go with this one unless you have a reason to prefer the other?

I'm a bit confused. This one *is* 243, where I manually renamed all the underlying item to oh_item. 

The other one, #242, left the underlying item variable name as is, and just creates a method called oh_item that returns the underlying item given via def_item_delegator

boc-tothefuture:
@jimtng sorry - my bad.. I meant #242 seems easier and better.  If the do the same things.. lets close this one and add tests to the other so we can merge.

PR #246, opened 2021-05-25
refactor: make compatible with new UoM library

### LINKED ISSUE #245: Heads up: Changed UoM libraries in OH
OH has recently changed the UoM libraries (https://github.com/openhab/openhab-core/commit/91307993ad2fbe1b357f1dab8d694d679e67ca8c) which are also used in our NumberItem/Quantity classes. I would assume this breaks the helper library in recent snapshots and will soon in milestones when M5 is released. However, updating to use the new UoM libraries would instead break it if running 3.0. 

How would we best handle this? We should assume some users would still be on 3.0 even after 3.1 is released, so we would need to support both versions. Is it possible to rework so we won't have to import the libraries directly?

pacive:
Ensure compatibility with the new UoM library introduced in OH 3.1 while maintaining backwards-compatibility with previous OH-versions.

Replaces direct references to the UoM library with OH proxies, which should improve stability.

Other changes in OH causes somewhat different output when calling toString() on Number Items, added explicit formatting in several tests to compensate for this.

Some tests used calculations with temperatures in ways that doesn't work in the real world due to the nature of temperature scales, these were removed.

Finally configures the rest api to allow basic auth in the testing setup, since this is disabled by default in OH 3.1

Closes #245 

### DIFF
--- a/rakelib/openhab.rake
+++ b/rakelib/openhab.rake
@@ -9,7 +9,7 @@ require 'digest/md5'
 # rubocop: disable Metrics/BlockLength
 # Disabled due to part of buid / potentially refactor into classes
 namespace :openhab do
-  @openhab_version = '3.0.2'
+  @openhab_version = ENV['OPENHAB_VERSION'] || '3.0.2'
   @jruby_bundle = 'https://github.com/boc-tothefuture/openhab2-addons/releases/download/3.1.0/org.openhab.automation.jrubyscripting-3.1.0-SNAPSHOT.jar'
   karaf_client_path = File.join(OPENHAB_DIR, 'runtime/bin/client')
   karaf_client_args = [karaf_client_path, '-p', 'habopen'].freeze
@@ -116,14 +116,24 @@ namespace :openhab do
   task download: [OPENHAB_DIR] do |task|
     state(task.name) do
       openhab_zip = "openhab-#{@openhab_version}.zip"
+      download_url = case @openhab_version
+                     when /.*\.(M\d)|(RC\d)/
+                       'https://openhab.jfrog.io/openhab/libs-milestone-local/org/openhab/distro/openhab/'\
+                       "#{@openhab_version}/#{openhab_zip}"
+                     when /.*-SNAPSHOT/
+                       'https://ci.openhab.org/job/openHAB3-Distribution/lastSuccessfulBuild/artifact/'\
+                       "distributions/openhab/target/#{openhab_zip}"
+                     else
+                       'https://openhab.jfrog.io/openhab/libs-release/org/openhab/distro/openhab/'\
+                       "#{@openhab_version}/#{openhab_zip}"
+                     end
       Dir.chdir(OPENHAB_DIR) do
-        puts "Downloading #{openhab_zip}"
-        IO.copy_stream(
-          open('https://openhab.jfrog.io/openhab/libs-release/org/openhab/distro/openhab/'\
-               "#{@openhab_version}/openhab-#{@openhab_version}.zip"), openhab_zip
-        )
+        # rubocop: disable Security/Open
+        puts "Downloading #{openhab_zip} from #{download_url}"
+        IO.copy_stream(open(download_url), openhab_zip)
         fail_on_error("unzip #{openhab_zip}")
         rm openhab_zip
+        # rubocop: enable Security/Open
       end
     end
   end
@@ -177,6 +187,7 @@ namespace :openhab do
       karaf('log:set TRACE org.openhab.core.automation')
       karaf('log:set TRACE org.openhab.binding.jrubyscripting')
       karaf('openhab:users add foo foo administrator')
+      karaf('config:property-set -p org.openhab.restauth allowBasicAuth true')
       sh 'rsync', '-aih', 'config/userdata/', File.join(OPENHAB_DIR, 'userdata')
     end
   end
--- a/features/gem_install.feature
+++ b/features/gem_install.feature
@@ -4,7 +4,7 @@ Feature:  gem_install
   Background:
     Given Clean OpenHAB with latest Ruby Libraries
 
-  @reset_library
+  @reset_library @wip
   Scenario: Install OpenHAB helper library
     Given OpenHAB is stopped
     And GEM_HOME is empty
--- a/features/quantity.feature
+++ b/features/quantity.feature
@@ -9,49 +9,47 @@ Feature:  quantity
       """
         logger.debug("<quantity> <operator> <operand> = <result>")
         result = <quantity> <operator> <operand>
-        logger.info("Result is #{result}")
+        logger.info("Result is #{result.format('%.1f %unit%')}")
       """
     When I deploy the rules file
     Then It should log 'Result is <result>' within 5 seconds
     Examples:
-      | quantity               | operator | operand                | result  |
-      | Quantity.new('50 °F')  | +        | Quantity.new('50 °F')  | 100 °F  |
-      | Quantity.new('50 °F')  | -        | Quantity.new('25 °F')  | 25 °F   |
-      | Quantity.new('50 °F')  | *        | Quantity.new('2 °F')   | 100 °F  |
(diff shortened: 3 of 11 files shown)


pacive:
@boc-tothefuture could you give some hints on how to configure the tests to run against the latest snapshot of oh in addition to 3.0.2?

pacive:
Found a bug (probably in the UoM library), which can hopefully be fixed before 3.1.0 is released. See https://github.com/openhab/openhab-core/issues/2386

Other than that, it seems to handle calculations with quantities having AbstractUnit::ONE as unit really strange, it would be better if these can be converted to BigDecimal, which works correctly. Also noticed this in the tests, is it really supposed to work this way?
```
      | quantity               | operator | operand | result |
      | Quantity.new('50 °F')  | +        | '50 °F' | 100 °F |
      | Quantity.new('50 °F')  | -        | '25 °F' | 25 °F  |
      | Quantity.new('50 °F')  | *        | '2 °F'  | 100 °F |
      | Quantity.new('100 °F') | /        | '2 °F'  | 50     |
```
Seem as division works as it should (°F/°F = 1), but why is multiplication expected to produce the same unit?

pacive:
See also my later [comment](https://github.com/openhab/openhab-core/issues/2386#issuecomment-848714936) at the issue I raised. 

Doing calculations with temperatures like we are doing in the tests makes no sense really, could we replace them with say length-units instead? Then it would make more sense to say that e.g. 2 m * 2 m = 4 m², but 2 m * 2 = 4 m

jimtng:
I agree that for multiplication and division operators, only one operand can have a unit, in order to produce a result of that unit. Like you said, 2m * 2m should really be 4m^2, so I'd say let's go ahead and modify our tests.

pacive:
Could replace all direct imports with OH proxy classes, which should ensure forward- and backward compatibility. Had to remove som tests because the new library does calculations with temperatures in a different (scientifically more accurate way), but could replace them with some other type of unit to provide better test coverage if that is desired.

boc-tothefuture:
> @boc-tothefuture could you give some hints on how to configure the tests to run against the latest snapshot of oh in addition to 3.0.2?

I can take that on.

rakelib/openhab.rake
```diff
@@ -115,15 +115,25 @@ namespace :openhab do
...
+                       'https://openhab.jfrog.io/openhab/libs-milestone-local/org/openhab/distro/openhab/'\
+                       "#{@openhab_version}/openhab-#{@openhab_version}.zip"
+                     when /.*-SNAPSHOT/
+                       'https://ci.openhab.org/job/openHAB3-Distribution/lastSuccessfulBuild/artifact/'\
+                       "distributions/openhab/target/openhab-#{@openhab_version}.zip"
+                     else
+                       'https://openhab.jfrog.io/openhab/libs-release/org/openhab/distro/openhab/'\
+                       "#{@openhab_version}/openhab-#{@openhab_version}.zip"
+                     end
       openhab_zip = "openhab-#{@openhab_version}.zip"
```
jimtng, line 129:
```suggestion
      openhab_zip = "openhab-#{@openhab_version}.zip"
      download_url = case @openhab_version
                     when /.*\.M\d/
                       'https://openhab.jfrog.io/openhab/libs-milestone-local/org/openhab/distro/openhab/'\
                       "#{@openhab_version}/#{@openhab_zip}"
                     when /.*-SNAPSHOT/
                       'https://ci.openhab.org/job/openHAB3-Distribution/lastSuccessfulBuild/artifact/'\
                       "distributions/openhab/target/#{@openhab_zip}"
                     else
                       'https://openhab.jfrog.io/openhab/libs-release/org/openhab/distro/openhab/'\
                       "#{@openhab_version}/#{@openhab_zip}"
                     end
```

jimtng (commented):
Thanks for your great work! I only have one minor suggestion.

rakelib/openhab.rake
pacive: Good catch, this would DRY it up a little.

pacive:
> Without looking at the actual proxy classes, using them makes sense from a compatibility and consistency standpoint.

[org.openhab.core.library.unit.Units::ONE](https://github.com/openhab/openhab-core/blob/53e9de585848de23c0832e6e5caf0793bcb64bdb/bundles/org.openhab.core/src/main/java/org/openhab/core/library/unit/Units.java#L101) is just a reference to AbstractUnit::ONE, so this should work just the same. The [org.openhab.core.types.util.UnitUtils#parseUnit()](https://github.com/openhab/openhab-core/blob/53e9de585848de23c0832e6e5caf0793bcb64bdb/bundles/org.openhab.core/src/main/java/org/openhab/core/types/util/UnitUtils.java#L150-L175) uses a different approach to parse the unit (plus some additional null/empty checks and special handling of `%unit%` and `%%`), but I don't see any reason for the result to be different.

rakelib/openhab.rake
```diff
@@ -9,7 +9,7 @@ require 'digest/md5'
 # rubocop: disable Metrics/BlockLength
 # Disabled due to part of buid / potentially refactor into classes
 namespace :openhab do
-  @openhab_version = '3.0.2'
+  @openhab_version = ENV['OPENHAB_VERSION'] || '3.0.2'
```
boc-tothefuture, line 12: perfect, that is how I was going to tackle it. 

Does that work for the RC versions?

pacive: Good point, will have to add that to the regex, they should be in the same path as the milestone builds.

pacive: The next issue I'm working on is to be able to have multiple versions installed side by side, to solve the caching in the github workers (and allowing for easier local testing). Having separate directories is no problem, the bigger issue is to ensure only one instance is running at a time.

boc-tothefuture:
Can you rebase off master and add the version you want to test to the OPENHAB_VERSIONS array in rakelib/github.rake?

rakelib/openhab.rake
boc-tothefuture: > The next issue I'm working on is to be able to have multiple versions installed side by side, to solve the caching in the github workers (and allowing for easier local testing). Having separate directories is no problem, the bigger issue is to ensure only one instance is running at a time.

I already took care of that in it is in main.

boc-tothefuture: Well, not the local part.. but it should work in github.

pacive: Scratch that, didn't know you could have separate caches for the different versions, nice work!

pacive:
The gem_install.feature test fails because it installs the current version of the helper library, which references the old UoM library.

pacive:
> The gem_install.feature test fails because it installs the current version of the helper library, which references the old UoM library.
> 
> Edit: Apart from that, testing for different versions works great!

Tagged that test as @wip temporarily to make sure all other tests pass, it can be removed after the next release that includes this PR.

boc-tothefuture:
> The gem_install.feature test fails because it installs the current version of the helper library, which references the old UoM library.
> 
> Edit: Apart from that, testing for different versions works great!

Why did that cause a failure?  Do we actually try and test the UoM library in that same feature?

pacive:
> Why did that cause a failure? Do we actually try and test the UoM library in that same feature?

No, but I guess that import is made somewhere just by requiring the library in a rule, which the test does after installing the gem.

boc-tothefuture:
Ok, this ready for final review/merge?

pacive:
Yes, I believe so, all tests pass for both stable and milestone.

pacive:
Do you want me to squash all commits in this pr, or should I split it into one for the code changes and one for the test updates?

boc-tothefuture:
> Do you want me to squash all commits in this pr, or should I split it into one for the code changes and one for the test updates?

Squashing is fine.  Once we merge will you open a PR to untag that test as WIP?

pacive:
> Once we merge will you open a PR to untag that test as WIP?

Absolutely!

PR #253, opened 2021-06-02
fix(items): turn wrapped items into singletons

### LINKED ISSUE #252: Decorated items cannot be used as hash keys
It seems that every time we refer to a decorated item, a new object is created / instantiated, and therefore, it has a unique object hash key. As a result, they cannot be used as a hash key to look up a hash whose keys are item objects.

This problem doesn't exist for undecorated items (those for which no decorators have been written yet) because the item just refers to the openhab item object.

To demonstrate this issue:

Items file:
```
// String has a ruby wrapper/decorator - so it suffers from this issue
String String1
// Dimmer doesn't yet have a ruby wrapper/decorator, so it just points to the openhab object
Dimmer Dimmer1
```

```ruby
logger.info("*** String1: #{String1.hash} #{String1.hash} #{String1.hash} ")
logger.info("*** Dimmer1: #{Dimmer1.hash} #{Dimmer1.hash} #{Dimmer1.hash}")
```

Output:
```
08:16:07.222 [INFO ] [jsr223.jruby.Object                  ] - *** String1: 1954029922 969136136 1229397979
08:16:07.223 [INFO ] [jsr223.jruby.Object                  ] - *** Dimmer1: -964186594 -964186594 -964186594
```
Notice the hash values for String1 are all different, vs Dimmer1 staying the same.

Not only that this creates memory and processing waste, the object cannot be used as a hash key. One would expect this to work but it doesn't:

```ruby
ITEM_LOOKUP = { String1 => "Hi, I'm a string", Dimmer1 => "Hi I'm a dimmer" }
rule 'test' do
  received_command String1, Dimmer1
  triggered do |item|
    logger.info("*** Message: #{ITEM_LOOKUP[item]}")
  end
end

String1 << 'Update me'
Dimmer1 << 10
```

Output:

```
08:22:32.584 [INFO ] [jsr223.jruby.Object                  ] - *** Message:
08:22:32.587 [INFO ] [jsr223.jruby.Object                  ] - *** Message: Hi I'm a dimmer
```

jimtng:
Create a decorated item cache to avoid creating multiple wrapped instances of the same item. Resolve #252

### DIFF
--- a/features/items.feature
+++ b/features/items.feature
@@ -238,3 +238,15 @@ Feature:  items
       """
     When I deploy the rules file
     Then It should not log "missing oh_item for DSL GroupItem" within 5 seconds
+
+  Scenario: Ensure only one instance is created for a decorated item
+    # Ref: https://github.com/boc-tothefuture/openhab-jruby/issues/252
+    Given items:
+      | type   | name    |
+      | Number | Number1 |
+    And code in a rules file
+      """
+      logger.info("Consistent hash") if Number1.hash == Number1.hash
+      """
+    When I deploy the rules file
+    Then It should log "Consistent hash" within 5 seconds
--- a/lib/openhab/core/entity_lookup.rb
+++ b/lib/openhab/core/entity_lookup.rb
@@ -21,6 +21,7 @@ module OpenHAB
     module EntityLookup
       include OpenHAB::Log
 
+      @decorated_items = {}
       #
       # Automatically looks up OpenHAB items and things in appropriate registries
       #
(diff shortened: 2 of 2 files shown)


jimtng:
Hmm, this caused a problem when dynamically changing group type. Should we exclude GroupItem from the cache/singleton?

jimtng:
> Hmm, this caused a problem when dynamically changing group type. Should we exclude GroupItem from the cache/singleton?

OK I've solved this problem. Apparently when the group type changed, perhaps because a new group with the same name is created that overwrites the old one, a new openhab group item object is created. I've added code to check whether the cached item's underlying oh_item is the same. If not, re-create the decorated item. It passed the group check now.

pacive:
Would it be better to create a specific delegation for the `#hash`-method to [GenericItem.hashCode()](https://github.com/openhab/openhab-core/blob/f8c2def1307701a59e7ef477d1abdfd6c25e1d33/bundles/org.openhab.core/src/main/java/org/openhab/core/items/GenericItem.java#L317-L322)? This should produce consistent hashes, but not sure if it could cause other issues. Cannot see anything obvious though.

jimtng:
> Would it be better to create a specific delegation for the `#hash`-method to [GenericItem.hashCode()](https://github.com/openhab/openhab-core/blob/f8c2def1307701a59e7ef477d1abdfd6c25e1d33/bundles/org.openhab.core/src/main/java/org/openhab/core/items/GenericItem.java#L317-L322)? This should produce consistent hashes,

The problem is that every time an item is being referenced in the code, a new decorator / wrapper instance is created. It seems very wasteful to me.

pacive:
That's a good point, this solution could potentially reduce the memory usage a bit.

> Hmm, this caused a problem when dynamically changing group type. Should we exclude GroupItem from the cache/singleton?

I would guess this is true for all Items, the `name` and `type` fields are declared `final`, so changing either will require creation of a new instance, which breaks the reference to `oh_item`. The `hashCode()` and `equals()` methods, however, only use the `name` field, which means the hash lookup still works. 

This makes me think, could this cause memory leaks if a lot of Items are changed/removed, but the reference is still in the Hash? Or are all objects in the library disposed between rule runs? @boc-tothefuture, do you know how the script engine handles this?

boc-tothefuture:
> > Would it be better to create a specific delegation for the `#hash`-method to [GenericItem.hashCode()](https://github.com/openhab/openhab-core/blob/f8c2def1307701a59e7ef477d1abdfd6c25e1d33/bundles/org.openhab.core/src/main/java/org/openhab/core/items/GenericItem.java#L317-L322)? This should produce consistent hashes,
> 
> The problem is that every time an item is being referenced in the code, a new decorator / wrapper instance is created. It seems very wasteful to me.

I am not opposed to performance improvements, but I do think we need to be careful when looking at them to ensure we are delivering an optimization.

Creating a new object is a O(1) operation, looking up and inserting into a hash is an O(1) operation.  From a complexity standpoint they are the same.  However, storing the objects in the hash increases memory consumption, so we have made a memory trade-off for an algorithm of the same complexity.

I am not an expert on the internal of the JVM, my understanding at a high level is that with the implementation of the generational garbage collector, the JVM is optimized to create and destroy objects that do not live a long time.

Creating the wrapper object could potentially be faster than hashing the object, indexing the hash to the bucket in the hashmap, iterating over the elements in that bucket (in case of hash collisions) and then comparing the equality of the object.  In this case, not only do we do an equals on the object in the ruby hash, we then get the oh_item and we call equals on it with the supplied item.  That is going to look at every single field of both openhab objects and ensure they are the same.  If some of those fields are strings and those strings don't point to the same object in memory, it is going to walk char by char through both strings to check if they are the same.  In the best case scenario (a cache hit) it will do this for every single field in the item.  In some cases that will be faster than creating the wrapper, in some cases it won't be.  But without looking at the bytecode it compiles down to, there is a lot of guessing.  The point I am trying to make is that the solution implemented isn't free.

This seems to solve other problems (hash lookups), so I am not opposed.  Without a thorough analysis we really don't know  if this is offering any real world performance improvement.  In some cases probably, in other cases probably not.  Without any sort of detailed analysis it is hard to say it is better or worse than passing through hashcode and equals to the underlying object.

I am ok with this solution - in general I do not like to implement performance improvements unless there is a measured performance impact we are trying to address.

jimtng:
I believe that without the hash storage/singleton, the memory consumption will be far more. We'll have a lot more instances in memory.  However, I agree that this may not offer much, if any, performance improvement, if object creation is not an expensive operation. A more thorough analysis is required, and I'm not an expert in this area.

My main objective with this PR is just to solve the object hash id issue. Given the fact that there may be a memory leak issue using the hash approach, I'll try delegating the hash method as suggested by @pacive and see if that could work.

jimtng:
I've submitted #254 as an alternative solution as suggested by @pacive. If that's accepted, this PR should be closed.

PR #254, opened 2021-06-03
fix(items): decorated items could not be used as hash keys

### LINKED ISSUE #252: Decorated items cannot be used as hash keys

jimtng:
Resolve #252 by delegating decorated items' #hash to GenericItem.hashCode() as suggested by @pacive in https://github.com/boc-tothefuture/openhab-jruby/pull/253#issuecomment-853558929

### DIFF
--- a/features/items.feature
+++ b/features/items.feature
@@ -238,3 +238,25 @@ Feature:  items
       """
     When I deploy the rules file
     Then It should not log "missing oh_item for DSL GroupItem" within 5 seconds
+
+  Scenario Outline: Items can be used as hash keys
+    # Ref: https://github.com/boc-tothefuture/openhab-jruby/issues/252
+    Given items:
+      | type   | name   |
+      | <type> | <name> |
+    And code in a rules file
+      """
+      ITEMS_HASH = { <name> => "I'm <name>" }
+      logger.info("#{ITEMS_HASH[<name>]}")
+      logger.info("<name>.hash == <name>.hash: #{<name>.hash == <name>.hash}")
+      logger.info("<name>.eql?(<name>): #{<name>.eql?(<name>)}")
+      """
+    When I deploy the rules file
+    Then It should log "I'm <name>" within 5 seconds
+    And It should log "<name>.hash == <name>.hash: true" within 5 seconds
+    And It should log "<name>.eql?(<name>): true" within 5 seconds
+    Examples:
+      | type   | name    |
+      | Number | Number1 |
+      | String | String1 |
+      | Player | Player1 |
--- a/lib/openhab/dsl/items/item_delegate.rb
+++ b/lib/openhab/dsl/items/item_delegate.rb
@@ -42,6 +42,8 @@ module OpenHAB
             def_delegator item, method
           end
           define_method(:oh_item) { instance_variable_get(item) }
+          define_method(:hash) { oh_item.hash_code }
+          define_method(:eql?) { |other| hash == other.hash }
         end
 
         #


jimtng:
Would there be any side effects to overriding #hash in the decorators?

boc-tothefuture:
I think we also have to override `eql?`

jimtng:
> I think we also have to override `eql?`

As it is now, eql? returns true, equal? returns false:

```
      logger.info("eql?: #{Number1.eql?(Number1)} equal?: #{Number1.equal?(Number1)}")
```

Output:
```
08:15:14.695 [INFO ] [jsr223.jruby.Object                  ] - eql?: true equal?: false
```

I believe it should be left as is?

boc-tothefuture:
From the Ruby docs:
```
Generates an Integer hash value for this object. This function must have the property that a.eql?(b) implies a.hash == b.hash.
```

I would think if we want to delegate hash we also want to delegate eql to the java equals method.

jimtng:
* Without this PR, Number1.eql?(Number1) is true, but Number1.hash == Number1.hash is false
* This PR delegates #hash to GenericItem.hash_code, and as a result, Number1.hash == Number1.hash becomes true

Since Number1.eql?(Number1) was already true before this PR and is still true after this PR, should we still override it? Sure it's easy enough to implement but I'm wondering if it's necessary.

boc-tothefuture:
> * Without this PR, Number1.eql?(Number1) is true, but Number1.hash == Number1.hash is false
> * This PR delegates #hash to GenericItem.hash_code, and as a result, Number1.hash == Number1.hash becomes true
> 
> Since Number1.eql?(Number1) was already true before this PR and is still true after this PR, should we still override it? Sure it's easy enough to implement but I'm wondering if it's necessary.

Do we know why it is true in this case? I am not sure how eql works on objects in ruby?

jimtng:
> Do we know why it is true in this case? I am not sure how eql works on objects in ruby?

It might have something to do with "Number" type. I tried it on String type and it returned false. So I guess we do need to implement it.

pacive:
I found a blog post about the differences between ==, eql? and equal?: https://medium.com/@khalidh64/difference-between-eql-equal-in-ruby-2ffa7f073532

The short version is that eql? should return true if the objects have the same hash key, but equal? should only return true if they are the same object. If eql? is not defined, it seems it's just an alias for ==.

jimtng:
I've added eql? implementation that compares the hash.

boc-tothefuture:
> I found a blog post about the differences between ==, eql? and equal?: https://medium.com/@khalidh64/difference-between-eql-equal-in-ruby-2ffa7f073532
> 
> The short version is that eql? should return true if the objects have the same hash key, but equal? should only return true if they are the same object. If eql? is not defined, it seems it's just an alias for ==.

Does that mean the default eql of object already just compares the hash?

boc-tothefuture:
> > I found a blog post about the differences between ==, eql? and equal?: https://medium.com/@khalidh64/difference-between-eql-equal-in-ruby-2ffa7f073532
> > The short version is that eql? should return true if the objects have the same hash key, but equal? should only return true if they are the same object. If eql? is not defined, it seems it's just an alias for ==.
> 
> Does that mean the default eql of object already just compares the hash?

Looking at the source, it doesn't appear that is the case.

jimtng:
> Looking at the source, it doesn't appear that is the case.

My findings confirm this, i.e. without overriding eql? it would return false.

pacive:
> Does that mean the default eql of object already just compares the hash?

I think it defaults to be an alias of ==, which would explain why it worked even without overriding hash

PR #256, opened 2021-06-04
feat(quantity): implement positive?, negative?, and zero? for quantity and dimensioned numberitem

### LINKED ISSUE #255: Quantity .positive? and negative? failed
Using #positive? and #negative? on quantity, and NumberItem with quantity failed with an error. I would expect that we should be able to use .positive? and negative? on a quantity / numberitem.

Jruby library 3.6.3 on openhab 3.1M4

```
  Scenario Outline: Quantity can use positive? and negative?
    Given code in a rules file
      """
        logger.info("Result is #{<quantity>}")
      """
    When I deploy the rules file
    Then It should log 'Result is <result>' within 5 seconds
    Examples:
      | quantity                          | result |
      | Quantity.new('50 °F').positive?   | true   |
      | Quantity.new('-100 °F').positive? | false  |
      | Quantity.new('50 °F').negative?   | false  |
      | Quantity.new('-100 W').negative?  | true   |
```

Output:
```
08:26:04.232 [INFO ] [jsr223.jruby.Module                  ] - OpenHAB JRuby Scripting Library Version 3.6.3 Loaded
08:26:04.233 [INFO ] [jsr223.jruby.Object                  ] - Processing Started for 5107bace-d831-40ed-906d-83efe6ec4c04
08:26:04.235 [TRACE] [3.jruby.OpenHAB::DSL::Types::Quantity] - Comparing -100 W to 0
08:26:04.235 [TRACE] [3.jruby.OpenHAB::DSL::Types::Quantity] - Coercing -100 W as a request from Integer
08:26:04.236 [TRACE] [3.jruby.OpenHAB::DSL::Types::Quantity] - Unitizing (-100 W) and (0)
08:26:04.250 [ERROR] [ript.internal.ScriptEngineManagerImpl] - Error during evaluation of script 'file:/openhab-jruby/tmp/openhab/conf/automation/jsr223/ruby/personal/cucumber_test20210604-21478-q14uhr.rb': org.jruby.embed.EvalFailedException: java.lang.IllegalArgumentException: Can not compare incompatible units.
```

jimtng:
Resolve #255

### DIFF
--- a/lib/openhab/dsl/types/quantity.rb
+++ b/lib/openhab/dsl/types/quantity.rb
@@ -20,6 +20,7 @@ module OpenHAB
         include OpenHAB::Log
 
         def_delegator :@quantity, :to_s
+        def_delegators '@quantity.double_value', :positive?, :negative?, :zero?
 
         java_import org.openhab.core.library.types.QuantityType
         java_import org.openhab.core.library.types.DecimalType
@@ -185,8 +186,8 @@ module OpenHAB
         #
         # @return [Array]  Array of QuantityType objects
         #
-        def to_qt(*quanities)
-          [quanities].flatten.compact.map { |item| item.is_a?(Quantity) ? item.quantity : item }
+        def to_qt(*quantities)
+          [quantities].flatten.compact.map { |item| item.is_a?(Quantity) ? item.quantity : item }
         end
 
         #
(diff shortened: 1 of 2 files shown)


lib/openhab/dsl/types/quantity.rb
```diff
@@ -171,6 +171,33 @@ def inspect
           end
         end
 
+        #
+        # Test if the value is positive
+        #
+        # @return [boolean] true if the quantity is > 0 in the same unit
+        #
+        def positive?
+          quantity > QuantityType.new(0, quantity.unit)
```
pacive, line 180: Wouldn't it be more efficient to use `quantity.double_value > 0`? Or perhaps even `quantity.double_value.positive?` would work?

jimtng: Great idea! Thanks, I didn't think of that.

PR #257, opened 2021-06-12
fix(logging): logger doesn't show the rule name

### LINKED ISSUE #84: logger doesn't show the rule name

jimtng:
Resolve #84

### DIFF
--- a/lib/openhab/dsl/rules/automation_rule.rb
+++ b/lib/openhab/dsl/rules/automation_rule.rb
@@ -29,7 +29,7 @@ module OpenHAB
           set_name(config.name)
           set_description(config.description)
           set_triggers(config.triggers)
-          @run_context = config.caller
+          @run_context = config
           @run_queue = config.run
           @guard = config.guard
           between = config.between&.yield_self { between(config.between) }


jimtng:
This PR doesn't include a fix for the test. The test produced a false positive. The test passed even though logger didn't show the rule name in the log. It passed because a previous trace log had the rule name. I wrote the details in https://github.com/boc-tothefuture/openhab-jruby/issues/84#issuecomment-858137736

jimtng:
It seems that #226 caused the regression on #84, because now this PR brings back the error that #226 fixed.

boc-tothefuture:
I think we will have to manually set the rule name when we create the logger.

jimtng:
> I think we will have to manually set the rule name when we create the logger.

not ideal. I wonder if we could just make "between" prop and call the caller's context `between` when it's running inside the rule execution (run, triggered)?

boc-tothefuture:
Hey @jimtng - Just saw this has been lurking out here.. 

By manually above I just meant in the code here, not by the end user.

jimtng:
> By manually above I just meant in the code here, not by the end user.

Thanks for the clarification. This would be great, but I'm a bit lost. When the run context is set to the caller, the `name` property is not visible to the logger method.

boc-tothefuture:
I think (haven't tested), that we would have to make the following changes:

Right now we do 
`include OpenHAB::Log`

We would have to remove that and then in initialize do a
`@logger = Logger.new(config.name)`

then change the calls in the class to be @logger instead of logger.

jimtng:
> I think (haven't tested), that we would have to make the following changes:
> 
> Right now we do
> `include OpenHAB::Log`
> 
> We would have to remove that and then in initialize do a
> `@logger = Logger.new(config.name)`
> 
> then change the calls in the class to be @logger instead of logger.

I tried this on automation_rule.rb but maybe I did something wrong. It didn't work, the behaviour remain unchanged. config.caller context is `main` so the @logger defined inside AutomationRule isn't being used by the code executed in the run block.

boc-tothefuture:
Maybe I am misunderstanding the problem.  

Can you provide an example of where you are logging that we are seeing the issue?

jimtng:
> Maybe I am misunderstanding the problem.
> 
> Can you provide an example of where you are logging that we are seeing the issue?

I updated https://github.com/boc-tothefuture/openhab-jruby/issues/84 with further details.

pacive:
> I think (haven't tested), that we would have to make the following changes:
> 
> Right now we do
> `include OpenHAB::Log`
> 
> We would have to remove that and then in initialize do a
> `@logger = Logger.new(config.name)`
> 
> then change the calls in the class to be @logger instead of logger.

How would this affect users' logging in their rules? Would one have to use `@logger.info("Message")` instead of just `logger.info`?

jimtng:
Since I'm not sure how to solve the scope issue, should I close this PR? The problem remains, however.

PR #259, opened 2021-09-14
fix(build): updated to new version of jruby scripting

boc-tothefuture:


rakelib/openhab.rake
```diff
@@ -10,7 +10,7 @@ require 'digest/md5'
 # Disabled due to part of buid / potentially refactor into classes
 namespace :openhab do
   @openhab_version = ENV['OPENHAB_VERSION'] || '3.0.2'
```
ccutrer, line 12: should you also update the default openhab version to 3.1.0?

boc-tothefuture: yes, but i will push that as a different PR.

PR #261, opened 2021-09-14
fix(build): commit the Gemfile.lock

ccutrer:
this will prevent random build failure when rubocop releases a new
version, but we haven't updated to it and addressed any new cops


ccutrer:
so... this was actually run with MRI bundler (I hadn't installed bundler under jruby yet, so system bundler got used). your docs don't seem to clarify _which_ ruby is best to develop with -- MRI 2.5, or JRuby 9.2 (corresponding to MRI 2.5). do you have a preference?

ccutrer:
btw, the answer _can_ be "either one, it doesn't matter", but I should update the PR then. it will just add an additional platform to the lockfile.

boc-tothefuture:
The testing is done against JRuby - so I would think that is the platform we would want to lock to.

ccutrer:
👍  now locked against _only_ jruby 9.2

PR #265, opened 2021-09-15
feat(duration): support duration methods on Float too

### LINKED ISSUE #263: .seconds only support on integer, not float
I want a timer to run in less than 1 second:

`after 0.1.seconds {}`, but I get an error that seconds isn't defined on the float

ccutrer:
closes #263

have to actually do math to not lose precision finer than 1ms,
and then truncate to integer once we're at ms


boc-tothefuture:
LGTM!

PR #267, opened 2021-09-15
feat(load_path): apply $RUBYLIB

ccutrer:
since it doesn't seem to be happening by JRuby embedded automatically,
do it manually. also document that this is the preferred way to share
code among multiple rules

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -148,7 +148,7 @@ jobs:
        uses: actions/cache@v2
        with: 
         path: tmp/
-        key: OpenHAB-setup-${{ hashFiles('.bundlehash', 'Rakefile') }}-${{ matrix.openhab_version }}
+        key: OpenHAB-setup-${{ hashFiles('.bundlehash', 'Rakefile', 'rakelib/openhab.rake') }}-${{ matrix.openhab_version }}
      - name: Setup OpenHAB
        if: steps.cache.outputs.cache-hit != 'true'
        env:
@@ -212,7 +212,7 @@ jobs:
         uses: actions/cache@v2
         with: 
           path: tmp/
-          key: OpenHAB-setup-${{ hashFiles('.bundlehash', 'Rakefile') }}-${{ matrix.openhab_version }}
+          key: OpenHAB-setup-${{ hashFiles('.bundlehash', 'Rakefile', 'rakelib/openhab.rake') }}-${{ matrix.openhab_version }}
       - name: Cucumber
         run: bundle exec rake features[${{ matrix.file }}]
       - name: Generate OpenHAB Dump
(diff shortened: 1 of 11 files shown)


boc-tothefuture:
Can you add a test to prevent future regressions?

ccutrer:
I'd love to, but I might need a bit of guidance. I've never used cucumber before, on top of the magic where it's all running inside of openhab...

give me a bit to see if I can figure it out on my own, and if I need some help I'll ask specific questions

boc-tothefuture:
Happy to help and provide guidance, let me know if you hit anything.

jimtng:
Great work, and I'm grateful that someone's tackling this. However, I'm wondering whether this should be done within the addon itself (in java) and not by the ruby library. I also vote for not creating the directory and just expecting the directory to already exist. This would reduce the amount of code to execute at load time.

ccutrer:
I agree that it might be slightly cleaner on the java side, but I'm not familiar with java AT ALL. really the only restriction of doing it this way is that your code must `require "openhab"` before it can require your own libraries, which doesn't seem awful. and the fact that this is done and works means it's better than a java side that doesn't work :)

and yes, this doesn't auto create any directories. nor does it even modify the load path by default. if you set up the rubylib in your jruby.conf, it'll add the env var, which will then add it to the load path. ruby won't care if that directory doesn't exist.

.github/workflows/workflow.yml
```diff
@@ -148,7 +148,7 @@ jobs:
        uses: actions/cache@v2
        with: 
         path: tmp/
-        key: OpenHAB-setup-${{ hashFiles('.bundlehash', 'Rakefile') }}-${{ matrix.openhab_version }}
+        key: OpenHAB-setup-${{ hashFiles('.bundlehash', 'Rakefile', 'rakelib/openhab.rake') }}-${{ matrix.openhab_version }}
```
ccutrer, line 151: can I get a little help here? it seems it's still using the key `OpenHAB-setup-06fe71c519ac01ca8b7a3bb24cb4ca1cf8cf9a6f71de0bdc21ec247155d0ab56-3.0.2` even with this change (and thus not properly setting up the openhab instance with the changes in openhab.rake). did I specify the additional file to include in the cache key hash wrong?

ccutrer: doh, never mind, I see. I need to change the hash both when the setup is cached, and when it's restored

ccutrer:
yay, it passed!

boc-tothefuture:
I am good with merging this tactically now with the strategic solution have it handled in the java side.  I think it provides immediate value while we figure out how to handle it in jruby.

jimtng:
@boc-tothefuture that sounds like a good approach

PR #269, opened 2021-09-15
fix(tests): speed up tests by avoiding unnecessary work

ccutrer:
* keep a persistent open connection to the openhab console, instead of
   re-launching every time we need it
 * don't do a REST query to ensure everything we just deleted is gone
   when we didn't delete anything

### DIFF
--- a/features/step_definitions/openhab_rules.rb
+++ b/features/step_definitions/openhab_rules.rb
@@ -50,6 +50,17 @@ def deploy_rule(**kwargs)
   deploy_ruby_file(directory: rules_dir, code: @rule, **kwargs)
 end
 
+def wait_for_rule(log_line)
+  if ENV['RELOAD_SCRIPTS_BUNDLE']
+    # force the bundle for ScriptFileWatcher to re-scan immediately; otherwise
+    # it can take up to 20s to notice the new rule file when Java's
+    # WatchService doesn't support actively watching, and only polls (i.e. on
+    # MacOS)
+    openhab_client("bundle:restart `bundle:id 'openHAB Core :: Bundles :: Automation Script RuleSupport'`")
+  end
+  wait_until(seconds: 60, msg: 'Rule not added') { check_log(log_line) }
+end
+
 def deploy_ruby_file(code:, directory:, filename: '', check_position: :end, check: true)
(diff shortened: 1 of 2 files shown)


ccutrer:
I'm not sure where I want to go with this. the first commit is good, but isn't a huge change. for the second commit, it's a MASSIVE win locally, but not much of a change on GitHub. local execution times (on a MacBook Pro) for `bin/cucumber features/dimmer_item.feature` is ~ 3:45 without these commits, and ~33s with both. I'm almost certain it's because Java's WatcherService on MacOS doesn't use fsevents, and instead polls every 20s, whereas it probably uses inotify on linux so notices file changes almost immediately. This is actually why it's currently breaking on GitHub - the rules get run _twice_, and for some tests they're not idempotent (like dimming something by two steps; gets applied twice). So.... do I make the forced-reload dependent on host OS (and ignore that technically the double run can still happen on Mac, it's just highly unlikely)?

boc-tothefuture:
For some reason I didn't remember it taking that long on the mac to find the new file.  I think for local testing it is OK to make it OS specific since we are dealing with OS specific behavior.

Is the first optimization (keeping the shell open) worth the complexity or is there a gem that does that already? 
Not against it, but am curious on the savings.. I am all for improving testing speed.

ccutrer:
In the latest revision, I went with an env var instead of platform detection. I wonder if it has to do with which java sdk I have installed (and openhab is running under): `OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.10+9, mixed mode)`

As for the cached openhab console connection... it saves about 1s per `openhab_client` call after the first due to java startup times. Typically that's 1-3 calls per individual test - one to clear the things on cleanup, one to do an update/send/trigger/etc., one per thing created (and then one per rule file, if you're using it force openhab to refresh immediately). On my laptop, _with_ the force-bundle-reload commit but not the cache-console commit, runtime from dimmer_item.feature goes from 33s to 3m45s. so yeah, pretty significant improvement depending on the case. I guess I was underselling it before - tests aren't fast for me unless you have _both_ commits.

boc-tothefuture:
I am going to rerun the tests here because when I look at the times for the one passing the test times have increased.

PR #270, opened 2021-09-17
fix(build): have release process keep Gemfile.lock up to date

ccutrer:
see https://github.com/Gusto/semantic-release-rubygem for docs on the option

### DIFF
--- a/.releaserc.json
+++ b/.releaserc.json
@@ -26,7 +26,8 @@
     [
       "semantic-release-rubygem",
       {
-        "gemFileDir": "pkg/"
+        "gemFileDir": "pkg/",
+        "updateGemfileLock": true
       }
     ],
     [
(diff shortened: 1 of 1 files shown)


boc-tothefuture:
Looks good @ccutrer does this change our workflow in any way?

ccutrer:
Yeah... it changes it that I don't have to keep resetting the lockfile back to its old version before I commit! More seriously, anytime I run `bundle exec` locally, it automatically updates the lockfile to reference the current version of the gem, and yeah, it's annoying. As for if it changes the development/review/release workflow at all - no. It should be business as usual, just the commit by GitHub actions should have a change in Gemfile.lock as well as version.rb.

But I can't be 100% certain this commit doesn't break the release process, as I have no way to test it. I tried to run `npx semantic-release` locally in dry run mode, but even that way the dependencies on credentials and the GitHub actions environment quickly grew out of control. So I'll admit there's some amount of 🤞  that it will just work. I feel pretty good though because the semantic-release actions already supported it, and it was just a minor configuration change.

PR #272, opened 2021-09-19
refactor(rubocop): add optional rubocop gems and fix issues

ccutrer:
all are autocorrects, except the eliding of the md5 method in github.rake


bin/cucumber
```diff
@@ -15,7 +15,7 @@ ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../../Gemfile',
 bundle_binstub = File.expand_path('bundle', __dir__)
 
 if File.file?(bundle_binstub)
-  if File.read(bundle_binstub, 300) =~ /This file was generated by Bundler/
```
boc-tothefuture, line 18: Should we be fixing things in autogenerated files we don't own? Will we get in a loop of whenever we update binstubs we have to autocorrect them?

Should we instead exclude bin/ from rubocop?

rakelib/github.rake
```diff
@@ -7,18 +7,10 @@ OPENHAB_VERSIONS = ['3.0.2', '3.1.0'].freeze
...
-      file.each(nil, 4096) { |chunk| md5 << chunk }
-    end
-    md5.hexdigest
-  end
-
   desc 'Release JRuby Binding'
   task :release, [:file] do |_, args|
     bundle = args[:file]
-    hash = md5(bundle)
+    hash = Digest::MD5.hexdigest(File.read(bundle))
```
boc-tothefuture, line 13: much better

boc-tothefuture:
Like more static analysis in general, only open question in my mind is if we should be running any rubocop on autogenerated binstubs in bin/

ccutrer:
it's a good question. I don't have strong feelings either way. I'd be more than happy to push up a change that excludes it (and regenerates them back to their original form, since I'm pretty sure some have already been applied). I know in other projects we haven't excluded bin, and I'm trying to think why. It's possible that we had some non-autogenerated scripts in there too, but I'm not 100% sure. WDYT?

boc-tothefuture:
My preference is to exclude them because it is not our code and I think we will end up having to autocorrect every time we update those gems.  

Rest of it looks great to me! 

One minor things (and we have not been great about this).  For your commit message, it gets automatically added to the generated changlog.  You may want to consider making it more descriptive, just as if you were writing the actual changelog. 

`incorporated additional static analysis tools into the build process (rubocop-rake, rubocop-performance) to enforce best practices and autocorrected existed defects`

BTW - Thanks for all the contributions, this is all great work!

ccutrer:
Binstubs excluded (and reverted) in latest patchset.

boc-tothefuture:
I fixed the release system so I can merge this now.  
Thanks for the work on this!  Always a fan of automatic best practice enforcement!

Issue #237, opened 2021-03-29
event.command when ON didn't work


jimtng:
This cucumber test failed:

```
  Scenario: Rule supports received_command
    Given items:
      | type   | name         | state |
      | Switch | Alarm_Switch | OFF   |

    And a deployed rule:
      """
      rule 'Execute rule when item received command' do
        received_command Alarm_Switch
        run do |event|
          case event.command
          when ON then logger.info("Item received command: ON")
          when OFF then logger.info("Item received command: OFF")
          end
        end
      end
      """
    When item "Alarm_Switch" state is changed to "ON"
    Then It should log 'Item received command: ON' within 5 seconds
```

pacive:
Probably related to #231

ccutrer:
I want to address this within the next week or two.

PR #273, opened 2021-09-19
improvements to working with event objects

ccutrer:
a series of commits that improve working with event (both update/change and command) objects. each commit is atomic, but they seemed related enough to send as a single PR


docs/usage/execution/run.md
```diff
@@ -14,18 +14,39 @@ The run property is the automation code that is executed when a rule is triggere
 ## State/Update Event Properties
 The following properties exist when a run block is triggered from an [updated](#updated) or [changed](#changed) trigger. 
 
-| Property | Description                      |
-| -------- | -------------------------------- |
-| item     | Triggering item                  |
-| state    | Changed state of triggering item |
-| last     | Last state of triggering item    |
```
boc-tothefuture, line 21: This is a breaking change.

features/received_command.feature
```diff
@@ -20,6 +20,23 @@ Feature:  received_command
     When item "Alarm_Mode" state is changed to "7"
     Then It should log 'Item received command: 7' within 5 seconds
 
+  Scenario: Rule supports helper predicates for named commands
+    Given a deployed rule:
+      """
+      rule 'Execute rule when item received command' do
+        received_command Alarm_Mode
+        run do |event|
+          %i[refresh? on? off? increase? decrease? up? down? stop? move? play? pause? rewind? fastforward? next? previous?].each do |pred|
```
boc-tothefuture, line 29: I can't figure out quickly by looking at it what we are testing here? 
While not as fast, often you do this sort of multi-test with a table in cucumber so you know exactly which scenarios aren't behaving.

lib/openhab/dsl/items/item_command.rb
```diff
@@ -54,7 +55,7 @@ def item_state(command_enum, methods = {})
           command_enum.values.each do |command|
             status_method = "#{command.to_s.downcase}?"
             status_method = methods.transform_keys(&:to_sym).fetch(status_method.to_sym, status_method)
-            logger.trace("Creating status method (#{status_method}) for #{self.class}")
+            logger.trace("Creating status method (#{status_method}) for #{self}")
```
boc-tothefuture, line 58: Just curious (no comment on the change), was the .class just not needed?

lib/openhab/dsl/monkey_patch/types/stop_move_type.rb
```diff
@@ -0,0 +1,23 @@
...
+      # Patches OpenHAB types
+      #
+      module Types
+        java_import Java::OrgOpenhabCoreLibraryTypes::StopMoveType
+
+        #
+        # Monkey patch for DSL use
+        #
+        class StopMoveType
+          alias inspect to_s
```
boc-tothefuture, line 18: Fair amount of duplicated code, out of curiosity did you try Monkey Patching the a common super class to alias inspect to_s? 

I can't remember how the dynamic dispatch works in JRuby and if that would work instead?

boc-tothefuture:
Major PR here!  Thanks for the all the work.

This implements a breaking change and by the rules of semver and conventional commits will need to be updated as such.  

I am not sure where I stand on the last vs. was, but we may want to alias last, last_null? and last_undef? to the was variants to avoid the breaking change. 

If the collective group here is fine with a breaking change so am I.

ccutrer:
Aliasing `last` to `was` would definitely avoid the more obvious breaking change. And we don't even _have_ to make that rename either. It _does_ read a little better when you do `event.was&.on?` vs. `event.last&.on?`. I think if I had been you when I originally threw it in I probably would have gone with `previous`. `last` just seems to conflict with `Array#last`, which is maybe why you _did_ pick it, to pick a method name familiar with rubyists. My feeling on `last` is that `event` is _not_ an array, or even anything resembling an array, and for array-like semantics representing all the values the items have had, the current value would actually be `last`, and the prior value would be `[-2]`, so you _don't_ want to use `last` for it.

BUT... I had considered the fact that `state` would now return `nil` instead of `NULL` or `UNDEF` as a breaking change as well. That one's not as clearly a break in interface though. So yeah.. still waiting to hear the other's thoughts.

boc-tothefuture:
Good point on last having meaning in the Ruby world.  I can't recall if the DSL uses previous or last.  I probably felt that previous was "too long".  

I agree that "was" reads better when chained.

features/received_command.feature
ccutrer: basically testing that the ItemCommandEvent responds to all these methods. I'm not that worried about testing each one individually will return `true` for their corresponding command being sent, but I do want to make sure that they all exist. so the scenario tests that one of them returns true, but the rule executes the code for all of them, and would fail completely if it doesn't exist.

lib/openhab/dsl/items/item_command.rb
ccutrer: yeah. `self` is already `DimmerItem`, etc. here. so with `self.class`, the logging was coming out `Creating status method (on?) for Class`, which seemed not useful.

lib/openhab/dsl/monkey_patch/types/stop_move_type.rb
ccutrer: I avoided it for now, because a) I'm not super familiar with JRuby yet, and how class hierarchies from Java come over to the Ruby object. and b) I was slightly worried about implementations for things like QuantityType/Quantity that are current sometimes-wrapped from their OH object. I'm happy to look into it more.

jimtng:
I noticed some duplications in item_state.rb and item_state_changed.rb. Is there a way to DRY them?

ccutrer:
Updated PR. `last` is now a back-combat alias for `was` on `ItemStateChangedEvent`. I still marked the commit as breaking change, since `state` and `last` can now return `nil` instead of `NULL` or `UNDEF`. `ItemStateChangedEvent` and `ItemStateEvent` now both mix in `ItemStateUnDefTypeHelpers` to avoid duplication.

pacive:
Great work! Other than the mentioned breaking changes, I see no problems with this. The duplicated code can probably be DRYed up later without breaking anything.

pacive (approved):

jimtng:
This resolves #237

ccutrer:
latest update adds @jimtng 's test for #237

jimtng (approved):

boc-tothefuture:
@ccutrer  - This looks awesome!  Great work.

Looking over the commit messages, can we squish this to a single commit with an "end user" description?  Those commit messages automatically go into the changelog and when I looked over them I wasn't sure they would fully communicate the impact of the change such that we could paste it into the community and have it understood?

What do you think?

boc-tothefuture:
I agree it would be nice to keep them separate - it looks like they were updated?  Are you good with what we now have?

ccutrer:
yes, I'm good to merge with what we have now

PR #278, opened 2021-10-03
fix(things, items): avoid wrapping every OpenHAB object

ccutrer:
a common use case is to use `items[dynamic_item_name]` or
`things[dynamic_thing_name]` to get a non-constant item or thing.
the old code would instantiate a new array or set of all items/things
in OpenHAB, and then immediately discard it when the [] method is used
to look up the item individually. this change delays instantiating
until an actual enumerable method is used

### DIFF
--- a/lib/openhab/dsl/items/items.rb
+++ b/lib/openhab/dsl/items/items.rb
@@ -2,6 +2,7 @@
 
 require 'java'
 require 'openhab/core/entity_lookup'
+require 'singleton'
 
 module OpenHAB
   module DSL
@@ -12,11 +13,14 @@ module OpenHAB
       #
       # Delegates to underlying set of all OpenHAB Items, provides convenience methods
       #
-      class Items < SimpleDelegator
+      class Items
+        include Enumerable
+        include Singleton
+
         # Fetches the named item from the the ItemRegistry
         # @param [String] name
         # @return Item from registry, nil if item missing or requested item is a Group Type
-        def[](name)
+        def [](name)
           OpenHAB::Core::EntityLookup.lookup_item(name)
         rescue Java::OrgOpenhabCoreItems::ItemNotFoundException
           nil
@@ -26,19 +30,35 @@ module OpenHAB
         # @param name [String] Item name to check
         # @return [Boolean] true if the item exists, false otherwise
         def include?(name)
-          # rubocop: disable Style/GlobalVars
-          !$ir.getItems(name).empty?
-          # rubocop: enable Style/GlobalVars
+          !$ir.getItems(name).empty? # rubocop: disable Style/GlobalVars
         end
         alias key? include?
+
+        # Calls the given block once for each Item, passing that Item as a
+        # parameter. Returns self.
+        #
+        # If no block is given, an Enumerator is returned.
+        def each(&block)
+          # ideally we would do this lazily, but until ruby 2.7
+          # there's no #eager method to convert back to a non-lazy
+          # enumerator
+          to_a.each(&block)
+        end
+
+        # explicit conversion to array
+        # more efficient than letting Enumerable do it
+        def to_a
+          items = $ir.items.grep_v(Java::OrgOpenhabCoreItems::GroupItem) # rubocop:disable Style/GlobalVars
+          OpenHAB::Core::EntityLookup.decorate_items(items)
+        end
+        # implicitly convertible to array
+        alias to_ary to_a
       end
 
       # Fetches all non-group items from the item registry
       # @return [OpenHAB::DSL::Items::Items]
       def items
-        # rubocop: disable Style/GlobalVars
-        Items.new(OpenHAB::Core::EntityLookup.decorate_items($ir.items.grep_v(Java::OrgOpenhabCoreItems::GroupItem)))
-        # rubocop: enable Style/GlobalVars
+        Items.instance
       end
     end
   end
--- a/lib/openhab/dsl/rules/triggers/trigger.rb
+++ b/lib/openhab/dsl/rules/triggers/trigger.rb
@@ -61,9 +61,13 @@ module OpenHAB
         # @return [Array] A new flat array with any GroupMembers object left intact
         #
         def separate_groups(item_array)
-          return item_array if item_array.grep(Array).length.zero?
+          # we want to support anything that can be flattened... i.e. responds to to_ary
+          # we want to be more lenient than only things that are currently Array,
+          # but Enumerable is too lenient because Array#flatten won't traverse interior
+          # Enumerables
+          return item_array unless item_array.find { |item| item.respond_to?(:to_ary) }
 
-          groups, items = item_array.partition { |item| item.is_a? OpenHAB::DSL::Items::GroupItem::GroupMembers }
+          groups, items = item_array.partition { |item| item.is_a?(OpenHAB::DSL::Items::GroupItem::GroupMembers) }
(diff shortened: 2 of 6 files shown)


lib/openhab/dsl/items/items.rb
```diff
@@ -26,19 +30,23 @@ class Items < SimpleDelegator
...
         def include?(name)
-          # rubocop: disable Style/GlobalVars
-          !$ir.getItems(name).empty?
-          # rubocop: enable Style/GlobalVars
+          !$ir.getItems(name).empty? # rubocop: disable Style/GlobalVars
         end
         alias key? include?
+
+        def each(&block)
+          # ideally we would do this lazily, but until ruby 2.7
```
boc-tothefuture, line 38: Even in 2.7 will need to look very closely.  I think ruby uses fibers for lazy, which converted into threads for JRuby.

features/received_command.feature
```diff
@@ -66,7 +66,7 @@ Feature:  received_command
     Given a deployed rule:
       """
       rule 'Execute rule when member of group receives any command' do
-        received_command AlarmModes.items
+        received_command AlarmModes.members
```
boc-tothefuture, line 69: Is this a breaking change or fixing a test case?

ccutrer, line 69: noet that this change is because the test is wrong. it only _used_ to pass out of sheer dumb luck. the global `items` method is leaking onto every object (which I've mentioned might be a possibility before, but now that I've seen it happen I'm more motivated to fix), so this was setting up a trigger on every openhab object, _not_ members of a group. and the internal method separating non-groups from groups was short-circuiting on array-like objects, instead of only GroupMembers objects. but since Items is now an Enumerable, instead of delegating to an Array, it was missing that and blowing up because it wasn't a GroupMembers object either (see below where that method gets simplified)

lib/openhab/dsl/items/items.rb
ccutrer: doesn't ruby implement _any_ block-to-enumerator conversion (i.e. `get_enum` or `to_enum`) via fibers?

and then we get into my lack of JVM familiarity - is it necessarily bad to spin up threads? OpenHAB already has a gazillion of them!

lib/openhab/dsl/rules/triggers/trigger.rb
```diff
@@ -61,9 +61,9 @@ def append_trigger(type, config)
         # @return [Array] A new flat array with any GroupMembers object left intact
         #
         def separate_groups(item_array)
-          return item_array if item_array.grep(Array).length.zero?
+          groups, items = item_array.partition { |item| item.is_a?(OpenHAB::DSL::Items::GroupItem::GroupMembers) }
```
ccutrer, line 64: and now I see with the other failing test you're trying to accept a regular Array here too. will fix to accept anything that's Enumerable (which will then still allow GroupMembers to get properly separated, even though it's now an Enumerable, but does not act like an Array)

features/received_command.feature
ccutrer: note that https://github.com/boc-tothefuture/openhab-jruby/pull/279 will make this case behave as you think it would

jimtng: It would be neater if this change is moved into #279

ccutrer: Done.

lib/openhab/dsl/items/items.rb
ccutrer: I did a bit of reading. tl;dr: JRuby still has problems with fibers, particularly with lazy enumerators. (see https://github.com/jruby/jruby/issues/6773). there have been multiple references to using Loom to re-implement fibers as true JVM native coroutines, but afaict that's still aspirational. given jruby _just_ released ruby 2.6 support, it could be a while before 2.7, and then we can re-evaluate

pacive:
Just to provide some historical context wrt the GroupItem vs GroupMembers part. This was done because of an ambiguity in using Groups in triggers, which can be interpreted in three different ways:

1. As a regular Item trigger (the GroupItem itself changed state/recieved a command)
2. As a "member of ..." trigger (one of the Group's members changed/recieved a command)
3. To create a separate trigger for each of the Group's members (which would exclude any item added after the rule is loaded)

The GroupMembers class (and the #members method) was created to prevent the group from being interpreted as an array, which would then create the trigger as (3), since providing an Array of Items is supported.

Given your changes here (which looks good!), do you see a better approach to accurately differentiate between the different trigger variants?

ccutrer:
> Just to provide some historical context wrt the GroupItem vs GroupMembers part. This was done because of an ambiguity in using Groups in triggers, which can be interpreted in three different ways:
> 
> 1. As a regular Item trigger (the GroupItem itself changed state/recieved a command)
> 2. As a "member of ..." trigger (one of the Group's members changed/recieved a command)
> 3. To create a separate trigger for each of the Group's members (which would exclude any item added after the rule is loaded)
> 
> The GroupMembers class (and the #members method) was created to prevent the group from being interpreted as an array, which would then create the trigger as (3), since providing an Array of Items is supported.
> 
> Given your changes here (which looks good!), do you see a better approach to accurately differentiate between the different trigger variants?

No, I like the current syntax we have. The leaking of `Items#items` onto `GroupItem` and returning *everything* was really the only problem functionally... you think you had your rule set up right, when really you were creating a trigger on every item in the system. Honestly, my initial impression was that `GroupItem#members` did _not_ return a collection-like object, and was simply a dummy object to enable "member of" triggers, but I now know it is usable (and that `all_members` is more akin to `members.flatten` -- in fact maybe we want to override `Enumerable#flatten` in `GroupMembers#flatten` to call `group.all_members`?) as an array-like object. which means if you _really_ want (3), then it's as easy as `changed GroupItem.members.to_a`. But maybe we should beef up the documentation in this area slightly.

pacive:
> Honestly, my initial impression was that `GroupItem#members` did _not_ return a collection-like object, and was simply a dummy object to enable "member of" triggers

This was the only thing it was meant to do, but I really like that you fixed it to also work like the java `GroupItem.members()` 👍 

> in fact maybe we want to override `Enumerable#flatten` in `GroupMembers#flatten` to call `group.all_members`?

Sounds like a good idea, although I don't expect any noticeable performance difference, especially since it's mostly used for creating triggers (i.e only at rule load time)

pacive (approved):

boc-tothefuture:
Can you update the commit message to be:
perf(things, items): improves performance on array accessors '[]' by delaying object conversation until enumeration

ccutrer:
> Can you update the commit message to be:
> perf(things, items): improves performance on array accessors '[]' by delaying object conversation until enumeration

Done.

ccutrer:
the build is failing because the commit subject line is too long now (which was my normal instinct here - "too long, put details below the relatively short subject"). but looking at CHANGELOG.md, it seems like the bot only puts the subject in there. some guidance would be appreciated.

boc-tothefuture:
Hmm.. It looks like it doesn't want the first line to be over 100 chars:

`perf(things, items): improves performance on array accessors '[]'`

Might be good for this one.  I am curious if commitlint is going to throw away the rest of the comment or do something with it.

ccutrer:
> Hmm.. It looks like it doesn't want the first line to be over 100 chars:
> 
> `perf(things, items): improves performance on array accessors '[]'`
> 
> Might be good for this one. I am curious if commitlint is going to throw away the rest of the comment or do something with it.

Done.

PR #280, opened 2021-10-05
fix(on_off_type): onofftype comparison error in case statements

jimtng:
This PR fixed the problem when OnOffType failed in case comparisons such as this:

```ruby

case ON
when ON then logger.info('case statement works with OnOffType')
when OFF then logger.info('error 1')
else logger.info('error 2')
end
```

The log will show: `error 2`


jimtng:
#273 also contains a similar (and better) fix.

lib/openhab/dsl/monkey_patch/types/on_off_type.rb
```diff
@@ -33,16 +33,15 @@ def !
           #  nil if object cannot be compared
           #
           def ===(other)
-            # A case statement here causes and infinite loop
-            # rubocop:disable Style/CaseLikeIf
-            if self == ON
-              other.on? if other.respond_to? :on?
+            if other.instance_of?(self.class)
```
ccutrer, line 36: The other command types use is_a? instead of instance_of?.  They should probably be uniform

ccutrer:
Ah, yes it does. But it'd still be nice to get your test in to be explicit about the functionality.

lib/openhab/dsl/monkey_patch/types/on_off_type.rb
jimtng: I initially used is_a? but rubocop rewrote it as instance_of?

jimtng:
> Ah, yes it does. But it'd still be nice to get your test in to be explicit about the functionality.

Please feel free to include my test into your PR

ccutrer:
Done.

PR #284, opened 2021-10-05
feat(DecimalType): zero?, positive?, negative? predicates directly on DecimalType

ccutrer:
so that you can inspect them directly from an event.state or event.was, without an Item involved.


jimtng:
I'm wondering if somehow `QuantityType` can be DRY'ed with `DecimalType` somehow

boc-tothefuture:
QuantityType was hard and evolved a lot.. could probably use a redesign.

PR #286, opened 2021-10-07
feat(types): allow comparison and arithmetic directly against states

ccutrer:
this pushes all comparisons and arithmetic down to state types, not item types,
and delegates appropriately. it also removes all inheritance from Numeric, in
favor of properly implementing ruby type system methods (most notably coerce,
and calling coerce). lots of manual overrides were removed in the process, and
several things were removed from being delegated types because the delegation
no longer added any value and just confused the type system.

notable module introductions are NumericItem and NumericType, which are
mixed in to NumberItem, DimmerItem, RollershutterItem; and DecimalType
(and by extension PercentType), and QuantityType respectively. this
DRYs up a lot of code shared between numeric-like OpenHAB classes that
didn't previously have a common ancestor.

Quantity and DateTime are now completely subsumed into their native
OpenHAB types, and are simply left as aliases for backwards compatibility.
QuantityType in particular dropped a LOT of code once arithmetic was
properly called on the Java type, and the delagation object was
eliminated.


ccutrer:
Adding pipe to Numeric is exactly what I was thinking of last night! I like it a lot.

boc-tothefuture:
There is a lot to think about in your comment.  I can't provide a thorough response tonight.

One thing I wanted to comment on was the HSBType - because last night I started looking at the open issue because I was about to start working on it when I saw that you had already started.  

I wrote a long comment in here about color objects/HSBtype but decided it would be better tracked in the original color issue #34 so I moved it over there. 

Take a look at that issue and let me know what you think.  Since the HSBType seems important to the overall decision making, we could see if what I proposed is agreeable and if it impacts what we do here?

boc-tothefuture:
@ccutrer - did the conversation we have in #34 impact your stance in https://github.com/boc-tothefuture/openhab-jruby/pull/286#pullrequestreview-780205260 at all or is it the same?

ccutrer:
> @ccutrer - did the conversation we have in #34 impact your stance in https://github.com/boc-tothefuture/openhab-jruby/pull/286#pullrequestreview-780205260 at all or is it the same?  
> 

Yes it affected my opinion, but not the result. Basically I'm more okay with implicit conversion from string, even for more exotic formats that aren't already supported by OpenHAB. Which just means that naming the full type will be even rarer than I thought, giving _more_ weight to _not_ dropping the `Type` suffix in the name of convenience. I still feel like we should import them into the global namespace for "compatibility" with the native rules DSL for the rare times when you do still need them for type discrimination of received commands. So... the current state of this PR.

ccutrer:
I threw it in the code, but haven't tested it yet. I'll polish that up.

boc-tothefuture:
@ccutrer this is a breaking change right?  

If so can you update the commit message with an !

PR #292, opened 2021-10-12
feat(attachments): adds attachments to triggers

boc-tothefuture:
This is a WIP with RFC.

As I am converting more and more rules to jruby I am hitting the same pattern, in which I have multiple rules that are 99% similar but I want to do something slightly different based on the triggers, so I end up with duplicate rules.

For example:
```
def check_garage_doors(reason)
  open_doors = groups['GarageDoors'].select(&:open?)
  logging.debug("#{open_doors.length} garage doors open")
  open_doors.map{ |door| "#{reason} and #{door.label} is open" }
            .each { logger.warn(text); notify(text) }
end

rule 'Check garage doors at night' do
  every :day, at: '8pm'
  run { check_garage_doors "It's 8PM"}
end

rule 'Check garage doors when alarm armed for night' do
  changed Alarm_Mode, to: [10,14]
  run { check_garage_doors "Night mode set"}
end
```

The problems with this are:
1. The rule is almost exactly the same and so I copy and paste them - when you copy and paste it is a great way to create errors if not initially, later on when modifying.
2. I have to extract the check_garage_doors out to a method because I have multiple rules

I created a WIP for adding attachments to triggers, which would enable this instead:

```
def check_garage_doors(reason)
  open_doors = groups['GarageDoors'].select(&:open?)
  logging.debug("#{open_doors.length} garage doors open")
  open_doors.map{ |door| "#{reason} and #{door.label} is open" }
            .each { logger.warn(text); notify(text) }
end

rule 'Check if garage doors are open' do
  every :day, at: '8pm', attach: "It's 8PM"
  changed Alarm_Mode, to: [10,14], attach: 'Night mode set'
  run { |event| check_garage_doors(event.attachment)}
end
```

Now that it is a single rule, there is no need for the check_garage_doors method, but you could keep it if you want:

```
rule 'Check if garage doors are open' do
  every :day, at: '8pm', attach: "It's 8PM"
  changed Alarm_Mode, to: [10,14], attach: 'Night mode set'
  run do |event|
     reason = event.attachment
     open_doors = groups['GarageDoors'].select(&:open?)
     logging.debug("#{open_doors.length} garage doors open")
     open_doors.map{ |door| "#{reason} and #{door.label} is open" }
                        .each { logger.warn(text); notify(text) }
  end 
end
```

The other way to handle this is with looping, which I was doing in rules, but I think it is more difficult to understand what is going on:

```
Appliance = Struct.new(:name, :item, :url)
[Appliance.new("washer", Washer_Progress, Washer_Progress_URL), 
 Appliance.new("dryer", Dryer_Progress, Dryer_Progress_URL), ].each do |appliance|
  rule "Update #{appliance.name} chart URL" do
    changed appliance.item
    run do 
        progress = appliance.item&.state&.float_value || 0
        appliance.url << "https://quickchart.io/chart?c=" +ERB::Util.url_encode("{type:'radialGauge',data:{datasets:[{data:[#progress}],backgroundColor:'#283592'}]},options:{title:{display:true,text:'Progress',fontSize:20}}}"
    end
  end
end
```

would become:

```
rule "Update appliance chart URLs" do
  changed Washer_Progress, attach: Washer_Progress_URL
  changed Dryer_Progress, attach: Dryer_Progress_URL
  run |event| do 
      progress = event.item&.state&.float_value || 0
      url = event.attachment
      url << "https://quickchart.io/chart?c=" +ERB::Util.url_encode("{type:'radialGauge',data:{datasets:[{data:[#progress}],backgroundColor:'#283592'}]},options:{title:{display:true,text:'Progress',fontSize:20}}}"
    end
  end
end
```

You can attach arbitrary objects, arrays, hashes, etc.

Open items:
1. For the RFC I only did a WIP for changed, not the other triggers, although they will mostly be the same
2. I like 'attach:' for adding it to the trigger, I don't like 'attachment' to get the attached object
3. Not sure how to handle the 'triggered' execution block, could add to the item the event object. 

WDYT?

### DIFF
--- a/lib/openhab/dsl/rules/automation_rule.rb
+++ b/lib/openhab/dsl/rules/automation_rule.rb
@@ -5,6 +5,8 @@ require 'set'
 require 'openhab/core/thread_local'
 require 'openhab/log/logger'
 
+require_relative 'item_event'
+
 module OpenHAB
   module DSL
     #
@@ -29,6 +31,8 @@ module OpenHAB
         # @param [Config] config Rule configuration
         #
         # Constructor sets a number of variables, no further decomposition necessary
+        # rubocop:disable Metrics/MethodLength
+        # Metrics disabled because only setters are called or defaults set.
         def initialize(config:)
           super()
           set_name(config.name)
@@ -41,7 +45,9 @@ module OpenHAB
           @between = between || OpenHAB::DSL::TimeOfDay::ALL_DAY
           # Convert between to correct range or nil if not set
           @trigger_delays = config.trigger_delays
+          @attachments = config.attachments
         end
+        # rubocop:enable Metrics/MethodLength
 
         #
         # Execute the rule
@@ -51,7 +57,7 @@ module OpenHAB
         #
         #
         def execute(mod = nil, inputs = nil)
-          logger.trace { "Execute called with mod (#{mod&.to_string}) and inputs (#{inputs&.pretty_inspect}" }
+          logger.trace { "Execute called with mod (#{mod&.to_string}) and inputs (#{inputs&.pretty_inspect})" }
           logger.trace { "Event details #{inputs['event'].pretty_inspect}" } if inputs&.key?('event')
           if trigger_delay inputs
             trigger_delay = trigger_delay(inputs)
@@ -89,6 +95,15 @@ module OpenHAB
         end
 
         #
+        # Get the trigger_id for the trigger that caused the rule creation
+        #
+        # @return [Hash] Input hash potentially containing trigger id
+        #
+        def trigger_id(inputs)
+          inputs&.keys&.grep(/\.event$/)&.first&.chomp('.event')
+        end
+
+        #
         # Returns trigger delay from inputs if it exists
         #
         # @param [Map] inputs map from OpenHAB containing UID
@@ -100,7 +115,24 @@ module OpenHAB
           # ["72698819-83cb-498a-8e61-5aab8b812623.event", "oldState", "module", \
           #  "72698819-83cb-498a-8e61-5aab8b812623.oldState", "event", "newState",\
           #  "72698819-83cb-498a-8e61-5aab8b812623.newState"]
-          @trigger_delays[inputs&.keys&.grep(/\.event$/)&.first&.chomp('.event')]
+          @trigger_delays[trigger_id(inputs)]
+        end
+
+        # If an attachment exists for the trigger for this event add it to the event object
+        # @param [Object] Event
+        # @param [Hash] Inputs into event
+        # @return [Object] Event with attachment added
+        #
+        def add_attachment(event, inputs)
+          attachment = @attachments[trigger_id(inputs)]
+          return event unless attachment
+
+          # Some events, like those from cron triggers don't have an event
+          # so an event is created
+          event ||= Struct.new(:event).new
+
+          event.attachment = attachment
+          event
         end
 
         #
@@ -268,8 +300,11 @@ module OpenHAB
         # @param [Map] inputs OpenHAB map object describing rule trigge
         #
         #
+        # rubocop:disable Metrics/MethodLength
+        # No logical way to break this method up
         def process_queue(run_queue, mod, inputs)
           event = inputs&.dig('event')
+          event = add_attachment(event, inputs)
 
           while (task = run_queue.shift)
             if task.is_a? RuleConfig::Delay
@@ -281,6 +316,7 @@ module OpenHAB
         rescue StandardError => e
           print_backtrace(e)
         end
+        # rubocop:enable Metrics/MethodLength
 
         #
         # Dispatch execution block tasks to different methods
--- a/docs/usage/execution/run.md
+++ b/docs/usage/execution/run.md
@@ -25,39 +25,42 @@ The following properties exist when a run block is triggered from an [updated](#
 | was?       | Prior state of triggering item was not NULL or UNDEF   |
 | was_null?  | Prior state was NULL                                   |
 | was_undef? | Prior state was UNDEF                                  |
+| attachment | Optional user provided attachment to trigger           |
 
 For compatibility, `last` is also aliased to `was`.
 
 ## Command Event Properties
 The following properties exist when a run block is triggered from a [received_command](#received_command) trigger.
 
-| Property     | Description                   |
-| ------------ | ----------------------------- |
-| command      | Command sent to item          |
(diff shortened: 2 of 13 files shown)


boc-tothefuture:
There are certainly multiple ways to solve this.  In my old rules DSL I certainly did what you have above - doing string based association within the rules.  I think awkward is probably different for each person.  Although both styles have pros/cons.  One nice benefit of the attach method is that you get validation when the rule is compiled if it will work, which avoids typos. In the string based it won't fail until the rule is invoked.

It also requires your items are named a specific and compatible way.  

However, it is one less element to contend with in the language. Although adding it, doesn't force usage. 

I could also solve this with a hash outside of the rule:
```
appliances = { Washer_Progress => Washer_Progress_URL,
                        Dryer_Progress => Dryer_Progress_URL ) }
```
Then my lookup would be in the rule itself on the triggering item.

It is a bit more to type of course, but does have the same benefit of validation during rule compilation. 

I won't force this in.
@jimtng and @pacive WDYT?

ccutrer:
the other thing I plan on using more in ruby rules is metadata. I already use it in a few key areas, but it's so awkward to use from the DSL I only use it if it's the only solution. it could help with your appliances rule, but not so much the garages one.

another thing I was thinking of that I'll throw out there, but is possibly even more awkward than your proposal, is being able to store the trigger, and then in the run method be able to tell which one triggered it. like:

```
rule "my rule" do
  t1 = cron ...
  t2 = SomeItem changed
  run do |event|
    do_something if event.trigger == t1
  end
end
```

boc-tothefuture:
I just ran into another similar use case:

I ended up doing this:

```
rule "UPB Bedroom Link activated" do 
  channel "#{upb_channel}_activated_event", triggered: ['62','63','64','66','67']
  run do |trigger|
    case trigger.event.to_i
    when 63,64
         Brian_Bedroom_Light.on
    when 62,66,67
         Brian_Bedroom_Light.off
    end
  end
end
```

If I had the trigger attach mechanism, I would have done this:
```
rule "UPB Bedroom Link activated" do 
  channel "#{upb_channel}_activated_event", triggered: ['63','64'], attach: ON 
  channel "#{upb_channel}_activated_event", triggered: ['62','66','67'], attach: OFF
  run { |trigger| Brian_Bedroom_Light << event.attachment }
end
```

jimtng:
In terms of naming, how about 'data' or 'tag' or 'arg' or 'param'

jimtng:
@boc-tothefuture your thoughts on this use case?

lib/openhab/dsl/rules/automation_rule.rb
```diff
@@ -253,12 +262,18 @@ def check_guards(event:)
...
               process_delay_task(inputs, mod, run_queue, task)
             else
+              event = inputs&.dig('event')
+              attachment = attachment(inputs)
+              if attachment
+                class << event
+                  attr_reader :attachment
+                end
+                event.instance_variable_set('@attachment', attachment)
+              end
```
jimtng, line 276: Is there a reason why this block of code needs to be inside the `while` loop? It seems to me that it could be done once before the loop.

jimtng (commented):
Thanks for reopening this PR! 

Are we still not sure about the naming? I have no objections to attach/attachment.

Some comments on this feature: whilst some of the simpler use cases can be achieved without this feature, i.e. by having a hash / association between an item => data, there are instances where this association differs based on which rule(s) it applies to.

For my case where I have multiple rules involving item1, it isn't associated to the same data across all the rules. In my case, Item1 can be associated with `X` in rule1 but it is associated with `Y` in rule2. A simple global hash won't work, and this is where the attachment feature is really useful because I can specify the association for each rule.

boc-tothefuture:
I am ok with the item attachments.. I know @ccutrer wasn't a fan to start.  Does @jimtng use case change your mind?

lib/openhab/dsl/rules/automation_rule.rb
```diff
@@ -100,7 +108,24 @@ def trigger_delay(inputs)
...
+        end
+
+        # If an attachment exists for the trigger for this event add it to the event object
+        # @param [Object] Event 
+        # @param [Hash] Inputs into event
+        # @return [Object] Event with attachment added
+        #
+        def add_attachment(event, inputs)
+          attachment = @attachments[trigger_id(inputs)]
+          return unless attachment
```
jimtng, line 121:
```suggestion
          return event unless attachment
```

lib/openhab/dsl/rules/triggers/cron.rb
```diff
@@ -72,22 +72,13 @@ def cron_expression_map
...
-        end
-
-        #
-        # Create a OpenHAB Cron trigger
-        #
-        # @param [String] expression OpenHAB style cron expression
-        #
-        def cron(expression)
-          @triggers << Trigger.trigger(type: Trigger::CRON, config: { 'cronExpression' => expression })
+          append_trigger(Trigger::CRON, { 'cronExpression' => expression }, attach: attach)
```
jimtng, line 81: We still need the `cron` trigger:

```suggestion
          cron(expression, attach)
        end

        #
        # Create a OpenHAB Cron trigger
        #
        # @param [String] expression OpenHAB style cron expression
        #
        def cron(expression, attach: nil)
          append_trigger(Trigger::CRON, { 'cronExpression' => expression }, attach)
```

boc-tothefuture: Oops.. thanks.

boc-tothefuture:
I have hit a snag currently as I can't seem to find a way to match up a cron trigger to the event it fires.  There is no context provided by OpenHAB from a cron trigger so I can't match it to a trigger UID.  I have crawled through the code base and the only way I can see to remediate this is to create another (dynamic) rule that keeps track of the UID and inserts it before calling the real rule.  It wouldn't be too difficult, just more complex than I wish it was.  

Seem odd that OpenHAB doesn't provide the same context of the trigger no matter how the event was fired.

jimtng:
> I have hit a snag currently as I can't seem to find a way to match up a cron trigger to the event it fires. There is no context provided by OpenHAB from a cron trigger so I can't match it to a trigger UID.

Perhaps for now we say that this feature doesn't work for cron triggers, and open an issue to track this in the future.

boc-tothefuture:
@jimtng ready for final review.

jimtng:
LGTM. I have been testing this on my production environment, including using it for my motion trigger rule. It works beautifully.

PR #295, opened 2021-10-18
feat(color): support for color items and hsb type

### LINKED ISSUE #34: Add Color Item Type

ccutrer:
closes #34

this still needs docs written

### DIFF
--- a/docs/usage/items/color.md
+++ b/docs/usage/items/color.md
@@ -1,0 +1,45 @@
+---
+layout: default
+title: ColorItem
+nav_order: 1
+has_children: false
+parent: Items
+grand_parent: Usage
+---
+
+
+# ColorItem
+
+ColorItem represents a color. Note that it inherits from DimmerItem, so you can
+call `on`, `off`, `on?`, `off?`, etc. on it. It's state type is an `HSBType`, which
+is generally stored as Hue, Saturation, and Brightness, but has easy helpers for
+working with RGB values of various forms.
+
+| Method     | Description                                                                     |
+| ---------- | ------------------------------------------------------------------------------- |
+| hue        | Returns the color's hue component as a QuantityType of unit DEGREE_ANGLE        |
+| saturation | Returns the color's saturation component as a PercentType                       |
+| brightness | Returns the color's brightness component as a PercentType                       |
+| red        | Returns the color's red component as a PercentType                              |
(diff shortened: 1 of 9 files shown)


ccutrer:
oh yeah, you're right, I didn't get to them. it's been a crazy week! I've had a good vacation though ;)

I don't care if we get them into this PR, or add them separately. I guess I have a slight preference to get this in sooner, as I'm already depending on it in my rules, so the sooner it's in the sooner I can get back to a regular release :).

boc-tothefuture:
Are you planning to add the docs to this PR?  I am fine w/ merging this and then I can open a PR for the hash support.

ccutrer:
> @ccutrer - Couple questions on the usage notes:
> 
> ```
> HueBulb << "#ff0000" # send 'red' as a command
> HueBulb.red # => 100%
> HueBulb.hue # => 0 °
> HueBulb.brightness # => 100%
> HueBulb.to_rgb # => [100%, 0%, 0%]
> HueBulb.rgb.to_s(16) # => "0xff0000"
> HueBulb.on? # => true
> ```
> 
>  1. When things are returning a % what type is that?
> 2. This my ignorance, I am sure, but Hue comes back as a degree?  What type is that?
> 3. We have `to_rgb' that returns an array of percents / we were going to accept a hash - shouldn't that be consistent?
> 4. We have `to_rgb' but to get the hex color code back out we do `.rgb.to_s(16)'.  Should that be `to_rgb.to_s(16)` or should the `to_rgb` instead be `rgb`?
> 5. Should we have a `to_hex` also?

1. It's a PercentType.
2. It's a QuantityType with unit DEGREE_ANGLE. HSB is modeled as a cylinder, with hue (the color wheel) being an angle, saturation (how much color) a percentage, and brightness a percentage.
3/4. 🤷  I was just documenting the java methods (https://www.openhab.org/javadoc/latest/org/openhab/core/library/types/hsbtype). `toRGB` returns an array with PercentType components, and `getRGB` returns an int (it actually always has the top 8 bits set to 0xff for 100% opacity, which seemed less than useful which is why I renamed it to `argb` and masked those out for `rib`).
5. Oh yeah, I totally added that method, need to document it.

ccutrer:
docs tweaked. let me know if you think we should do some method renames though.

PR #297, opened 2021-10-18
feat(rules): add terse rule syntax for simple rules

ccutrer:


ccutrer:
On the subject of rule names, do you know the difference between rule name and rule description? In OpenHAB's UI, the rules will show up with the name as bold, and then some UUID in not-bold: https://www.dropbox.com/s/zx54j0nztdxkx5t/Screen%20Shot%202021-10-18%20at%202.50.17%20PM.png?dl=0

AFAICT we're not accidentally setting description when we should be setting name. So where is the UUID coming from? For DSL rules the lower part ends up being "<filename>-<index>" (like "av-10"). It'd be nice if we could set that lower part to "hiome.rb-<uuid>" (or even something along the lines of a terse description of the trigger, like e"hiome.rb-gOccupancy-member-changed), and the upper part as is (or for terse rules, generating a friendly name, like "Member of gOccupancy Changed").

PR #314, opened 2021-10-19
Cancels outstanding timers when rules file is unloaded

### LINKED ISSUE #115: When a rule is removed any associated timers should be automatically cancelled
'for' timers, delay timers, etc.

boc-tothefuture:
This closes #115

### DIFF
--- a/features/changed_duration.feature
+++ b/features/changed_duration.feature
@@ -178,3 +178,22 @@ Feature: changed_duration
     When I deploy the rule
     And item "String_One" state is changed to "TWO"
     Then It should log "Changed rule: String_One changed" within 5 seconds
+
+  Scenario: Timers in changed duration are cancelled if the script is removed
+    Given items:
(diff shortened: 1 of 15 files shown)


boc-tothefuture:
@ccutrer assuming this passes checks, please take a look.  I had to change several log statements in classes you wrote.  When I did a require of the 'script_handler.rb' the load order changed and method calls you had to logger did not work. I required and included the logger in those classes.  

Let me know if I need to revert that and instead do something different so those classes can log.

boc-tothefuture:
@ccutrer I had to comment out the terse rules or else channels.feature was throwing an infinite loop after I made the change to the require order..  Do you want to take a look?

boc-tothefuture:
> what happens if you use `after` at top-level?

Doesn't work... I am going to reorganize the code a bit based on that and some other things going on in the reentrant PR.

PR #320, opened 2021-10-20
[WIP] Timers support reentrancy

### LINKED ISSUE #18: After blocks in rules should support reentrancy
Issue #16 describes where this could be helpful.

boc-tothefuture:
Closes #18

Adds timed commands

### DIFF
--- a/docs/usage/items.md
+++ b/docs/usage/items.md
@@ -54,18 +54,18 @@ end
 Item types have methods added to them to make it flow naturally within the a ruby context.  All methods of the OpenHAB item are available plus the additional methods described below.
 
 
-| Method  | Description                                                                          | Example                                                      |
-|---------|--------------------------------------------------------------------------------------|--------------------------------------------------------------|
-| <<      | Sends command to item                                                                | `VirtualSwitch << ON`                                        |
-| command | alias for shovel operator (<<)                                                       | `VirtualSwitch.command(ON)`                                  |
-| ensure  | Only send following command/update if the item is not already in the requested state | `VirtualSwitch.ensure.on`                                    |
-| update  | Sends update to an item                                                              | `VirtualSwitch.update(ON)`                                   |
-| id      | Returns label or item name if no label                                               | `logger.info(#{item.id})`                                    |
-| undef?  | Returns true if the state of the item is UNDEF                                       | `logger.info("SwitchTest is UNDEF") if SwitchTest.undef?`    |
-| null?   | Returns true if the state of the item is NULL                                        | `logger.info("SwitchTest is NULL") if SwitchTest.null?`      |
-| state?  | Returns true if the state is not UNDEF or NULL                                       | `logger.info("SwitchTest has a state") if SwitchTest.state?` |
-| state   | Returns state of the item or nil if UNDEF or NULL                                    | `logger.info("SwitchTest state #{SwitchTest.state}")`        |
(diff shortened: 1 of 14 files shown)


jimtng:
Here's one of my rules that use a timer. What I want to achieve with these rules is: - When the aircon is turned on and if there's a door/window opened, issue a voice warning in 5 minutes, and repeat that warning every 15 minutes until all doors/windows are closed. There are some other conditions to deal with but that's not as relevant to discuss in relation to the timer feature.

```ruby
def door_window_alert?
  Aircon_Power.on? && Aircon_Warning.on? && Aircon_Mode != 'FAN' && !Alarm_Armed.on? && gWindowsAndDoors.open?
end

@@door_timer = nil
def door_timer
  @@door_timer ||= after 15.minutes do |timer|
    next unless door_window_alert?

    open_doors = gWindowsAndDoors.all_members.grep(OPEN).map(&:label)
    say "The #{open_doors.to_sentence} #{open_doors.linking_verb} open while the Aircon is running"
    timer.reschedule
  end
end

rule 'Aircon: Door Alert Timer' do
  changed Aircon_Power, to: ON
  changed Aircon_Mode, from: 'FAN'
  changed gWindowsAndDoors, from: CLOSED, to: OPEN
  on_start
  only_if { door_window_alert? }
  run do
    door_timer.reschedule 5.minutes
    logger.info("RUBY: Warning will be issued #{gWindowsAndDoors}")
  end
end

rule 'Aircon: Cancel Door Alert Timer' do
  changed gWindowsAndDoors, to: CLOSED
  changed Aircon_Power, to: OFF
  changed Alarm_Armed, to: ON
  run do
    @@door_timer&.cancel
    @@door_timer = nil
  end
end
```

I suppose I can remove the last rule, and not cancel the timer, because when the timer executes, it will just return should it no longer need to issue a voice warning at that time (e.g. all doors and windows are closed by then). The check / cancellation are two redundant actions. In my mind however, it is more efficient to cancel the timer as soon as we know it's not needed (i.e. what the last rule does).

---

To use this PR, my rules can be modified to:

```
rule 'Aircon: Door Alert Timer' do
  changed Aircon_Power, to: ON
  changed Aircon_Mode, from: 'FAN'
  changed gWindowsAndDoors, from: CLOSED, to: OPEN
  on_start
  only_if { door_window_alert? }
  run do
    after (5.minutes, reentrant: Aircon_Power) do |timer|
      next unless door_window_alert?

      open_doors = gWindowsAndDoors.all_members.grep(OPEN).map(&:label)
      say "The #{open_doors.to_sentence} #{open_doors.linking_verb} open while the Aircon is running"
      timer.reschedule 15.minutes
    end
  end
end
```
Although this will forego the ability to cancel the timer from another rule, unless I save the timer in a variable.

jimtng:
Perhaps this syntax can be used to cancel the timer:
```ruby
after(nil, reentrant: Object)
```

So in my example above, the last rule can be written as:
```ruby
rule 'Aircon: Cancel Door Alert Timer' do
  changed gWindowsAndDoors, to: CLOSED
  changed Aircon_Power, to: OFF
  changed Alarm_Armed, to: ON
  run do
    after(nil, reentrant: Aircon_Power) # Cancel the timer without having to use an external variable to hold the timer object
  end
end
```

jimtng:
or: `after(reentrant: Object)` can return the timer object, so this can be used to cancel it:

```ruby
after(reentrant: Object)&.cancel
```

PR #343, opened 2021-10-30
build(ci): optimize CI builds by splitting features across number of …

### LINKED ISSUE #310: Decrease test time by reusing runners
Presently we have a runner for each feature file.  This is nice because you have a runner for each future, therefore when a test fails it is obvious which runner caused it to fail. 

However, the issue is that setting up each runner takes a long time as openhab has to start, warmup, etc. 

Instead we could spread the individual tests out among the 20 runners.  That would start 20 runners and divide the individual tests among each runner. For example:

```
runner1:  features/thing.feature:11, features/timer.feature:118
runner2: features/thing.feature:24, features/time_of_day:45
```

This should reduce test time and spread the load somewhat evenly across the runners.  However, when a test fails you would need to look at the output to figure out exactly which feature failed. 

WDYT?

boc-tothefuture:
…runners

closes #310

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -176,9 +176,12 @@ jobs:
        run: |
           bundle config path vendor/bundle
           bundle install --jobs 4 --retry 3
+     - name: Remove env.rb file
+       run: |
+          rm features/support/env.rb
      - id: set-matrix
        run: |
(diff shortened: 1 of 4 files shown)


boc-tothefuture:
Looks like this cuts build time by 1/2 to 1/3.

PR #353, opened 2021-10-31
fix(triggers): trigger methods return the trigger objects

jimtng:
The source code documentation stated that trigger methods (update, changed, received_command) return the OpenHAB triggers. However they just return the items being passed to them. 

This PR makes a slight modification so that these methods return the openhab trigger objects.


boc-tothefuture:
Would you mind adding a test?  I am guessing not having a test here is how we deviated from the docs.

jimtng:
Not sure of the best way to check. I've just added a test to see if it returns `Java::OrgOpenhabCoreAutomationInternal::TriggerImpl` WDYT?

PR #371, opened 2021-11-03
feat(between): support for month-day ranges

### LINKED ISSUE #116: between should also work with dates
for example
between '06/06'..'12/31'

boc-tothefuture:
Closed #116

### DIFF
--- a/lib/openhab/dsl/time/month_day.rb
+++ b/lib/openhab/dsl/time/month_day.rb
@@ -1,0 +1,140 @@
+# frozen_string_literal: true
+
+module OpenHAB
+  module DSL
+    # Support for time related functions
+    module Between
+      # Manages Month Day Ranges
+      module MonthDayRange
+        include OpenHAB::Log
+
+        java_import java.time.Year
+
+        # Lambdas are used to calculate the year for the month day
+        # which must happen during evaluation time to support that rules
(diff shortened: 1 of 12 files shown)


lib/openhab/dsl/time/month_day.rb
```diff
@@ -0,0 +1,140 @@
...
+        include OpenHAB::Log
+        # Parse MonthDay string as defined with by Monthday class without leading double dash "--"
+        def self.parse(string)
+          ##          string = "--#{string}" unless string.to_s.start_with? '--'
+          java_send :parse, [java.lang.CharSequence], "--#{string}"
+        end
+
+        # Can the supplied object be parsed into a MonthDay
+        def self.day_of_month?(obj)
+          /^-*[01][0-9]-[0-3]\d$/.match? obj.to_s
```
jimtng, line 118: Why not use parse and catch the exception? This way if we ever wanted to change parse to support '6 Jan' it would still work?

boc-tothefuture: If we want to update with additional formats we would need to update this check.

Java exceptions are fairly expensive and I try to follow the effective Java guidelines.

http://www.thefinestartist.com/effective-java/57

Issue #366, opened 2021-11-02
Timers don't always get cancelled on script reloads


jimtng:
I am scratching my head about scriptunload / automatic cancelling of the timers. It doesn't seem to happen reliably, i.e. sometimes the outstanding/active timers get cancelled but sometimes they don't seem to get cancelled.

Here's how to reproduce it: create a script
```ruby
require 'openhab'
# x
t = Time.now
after(1.minute) { logger.error("Time's up #{t}") }
```

I just kept adding x to the comments and re-saving it (so that the file is different and therefore gets reloaded by openhab), before the 1 minute had passed.

In theory, it should cancel all outstanding timers, and you'll only see ONE "Time's up <timestamp>" in the log once you've stopped re-saving the file for a minute. But if you re-save it often enough (and within less than 1 minute), you'll see a bunch of log entries with different timestamps which indicates that those timers didn't get cancelled, and multiple timers got executed.

_Originally posted by @jimtng in https://github.com/boc-tothefuture/openhab-jruby/issues/316#issuecomment-956306110_

boc-tothefuture:
Can you add a logging statement after the timer is added?  I am curious if scriptUnloaded is being called each time? OpenHAB is watching file events.. it is certainly possible the deletions or reloads are getting lost if it is happening too often.  Are you seeing it in production or just when you make this test?

ccutrer:
I can't even get the test for "Scenario: Timers in rules are cancelled if the script is removed" to pass for me locally. Log looks like this:

```
2021-11-01 20:08:46.443 [TRACE] [jsr223.jruby.Object                 ] - Script loaded: file:/Users/cody/src/openhab-jruby/tmp/openhab/conf/automation/jsr223/ruby/personal/cucumber_test20211101-21439-1wdeaa4.rb
2021-11-01 20:08:46.444 [DEBUG] [rt.internal.loader.ScriptFileWatcher] - Script loaded: /Users/cody/src/openhab-jruby/tmp/openhab/conf/automation/jsr223/ruby/personal/cucumber_test20211101-21439-1wdeaa4.rb
2021-11-01 20:08:48.940 [INFO ] [utomation.rest.internal.RuleResource] - Received HTTP DELETE request at 'rules/32732332-9a0b-475b-b783-d1d3a9ee91a3' for the unknown rule '32732332-9a0b-475b-b783-d1d3a9ee91a3'.
[...]

PR #376, opened 2021-11-04
fix(core): execute script unloaded if method is missing

boc-tothefuture:


boc-tothefuture:
@jimtng could you try this and see if it fixes the issue you were seeing in #366 
?

jimtng:
I tried this, but I still got the same message:

```
08:02:57.714 [TRACE] [ript.internal.ScriptEngineManagerImpl] - scriptUnloaded() is not defined in the script
```

PR #390, opened 2021-11-07
fix(binding): fix binding support for gem upgrades

boc-tothefuture:


boc-tothefuture:
@ccutrer and @jimtng  - Can you check if the version of the binding referenced here works on your 3.2M2+ versions?

boc-tothefuture:
I tested it locally and it seems to work.  But would like verification from one of you two running it in prod.

jimtng:
I just got a macbook pro m1, so I had to update my ruby version, reinstall xcode, but then after that rake openhab:setup worked fine.

PR #394, opened 2021-11-08
fix(dev): add mac M1 platform to Gemfile.lock

jimtng:
Gemfile.lock keeps getting out of sync on my system. Would you mind merging this in please?


ccutrer:
adding your platform should be no problem, but the commit message still needs to follow the standard. Ie "fix(dev): add mac M1 platform to Gemfile.lock"

PR #403, opened 2021-11-18
ci(bundle): use the official bundle

### LINKED ISSUE #396: scriptLoaded and scriptUnloaded not called
I'm not sure if I should create this issue here, because this seems to be a bundle/addon related issue and not the openhab-jruby library issue. In any case, I'm opening the issue here to keep track of this.

Related: #366

jimtng:
This PR does 4 things:

- Make the rake openhab:update command without a file argument to pull the snapshot bundle remotely
- Change the bundle source url to the snapshot build from openhab-addons since the bundle has been merged into openhab
- Because the bundle sets the default local_context to singlethread, remove `local_context` setting from the test environment, as it shouldn't even be set to `threadsafe` anyway.
- Set the gem_home to conf/scripts/lib/..... instead of conf/automation/lib/...

### DIFF
--- a/features/gem_install.feature
+++ b/features/gem_install.feature
@@ -11,7 +11,7 @@ Feature:  gem_install
     And a services template filed named "jruby.cfg"
       """
       org.openhab.automation.jrubyscripting:gem_home=<%= gem_home %>
-      org.openhab.automation.jrubyscripting:gems=openhab-scripting=~>3.0
+      org.openhab.automation.jrubyscripting:gems=openhab-scripting=~>4.0
       org.openhab.automation.jrubyscripting:rubylib=<%= ruby_lib_dir %>
       """
     And code in a rules file:
@@ -19,7 +19,7 @@ Feature:  gem_install
       logger.info("OpenHAB helper library is at #{OpenHAB::VERSION}")
       """
     When I start OpenHAB
-    Then It should log 'Installing Gem: openhab-scripting' within 120 seconds
+    Then It should log 'Installing Gem: openhab-scripting' within 180 seconds
     And I deploy the rules file
-    Then It should log 'OpenHAB helper library is at' within 160 seconds
+    Then It should log 'OpenHAB helper library is at' within 200 seconds
 
--- a/features/support/openhab.rb
+++ b/features/support/openhab.rb
@@ -75,7 +75,7 @@ def ruby_lib_dir
 end
 
 def gem_home
-  File.join(openhab_dir, 'conf/automation/lib/ruby/gem_home')
+  File.join(openhab_dir, 'conf/scripts/lib/ruby/gem_home')
 end
 
 def services_dir
--- a/rakelib/openhab.rake
+++ b/rakelib/openhab.rake
@@ -11,7 +11,7 @@ require 'digest/md5'
 # rubocop: disable Rake/MethodDefinitionInTask Legacy code
 namespace :openhab do
   @openhab_version = ENV['OPENHAB_VERSION'] || '3.1.0'
-  @jruby_bundle = 'https://github.com/boc-tothefuture/openhab2-addons/releases/download/3.2.0/org.openhab.automation.jrubyscripting-3.2.0-SNAPSHOT.jar'
+  @jruby_bundle = 'https://ci.openhab.org/job/openHAB-Addons/lastStableBuild/artifact/bundles/org.openhab.automation.jrubyscripting/target/org.openhab.automation.jrubyscripting-3.2.0-SNAPSHOT.jar'
   karaf_client_path = File.join(OPENHAB_DIR, 'runtime/bin/client')
   karaf_client_args = [karaf_client_path, '-p', 'habopen'].freeze
   @karaf_client = karaf_client_args.join(' ')
@@ -89,7 +89,7 @@ namespace :openhab do
 
   def gem_home
     full_path = File.realpath OPENHAB_DIR
-    File.join(full_path, '/conf/automation/lib/ruby/gem_home')
+    File.join(full_path, '/conf/scripts/lib/ruby/gem_home')
   end
 
   def ruby_lib_dir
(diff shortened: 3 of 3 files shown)


boc-tothefuture:
That new bundle isn't in the CI yet.  Should this be set to singlethread until then?

jimtng:
The merged bundle is already in 3.2 snapshot. I was just looking into this to figure out how to change the bundle source to point there... WDYT?

jimtng:
Would this be the latest snapshot build? https://ci.openhab.org/job/openHAB-Addons/lastStableBuild/artifact/bundles/org.openhab.automation.jrubyscripting/target/org.openhab.automation.jrubyscripting-3.2.0-SNAPSHOT.jar

boc-tothefuture:
Yep.  If you update to point to that in the rake file it should take care of everything else.

jimtng:
@boc-tothefuture I've updated the PR's title and description.

jimtng:
It failed on OpenHAB 3.0 due to bundle dependencies.

boc-tothefuture:
Do we drop 3.0 support?  I know @ccutrer was not in favor of that before.

jimtng:
```
  Unresolved requirement: Import-Package: com.sun.nio.file; resolution:="optional"
  Unresolved requirement: Import-Package: com.sun.security.auth.module; resolution:="optional"
  Unresolved requirement: Import-Package: org.apache.tools.ant; resolution:="optional"
  Unresolved requirement: Import-Package: org.bouncycastle.jce.provider; version="[1.52.0,2.0.0)"; resolution:="optional"
  Unresolved requirement: Import-Package: org.joda.convert; resolution:="optional"
  Unresolved requirement: Import-Package: sun.management; resolution:="optional"
  Unresolved requirement: Import-Package: sun.nio.ch; resolution:="optional"
  Unresolved requirement: Import-Package: jakarta.annotation
```

TBH, I don't really know what those are for and why they need to be in the pom.xml

jimtng:
> Do we drop 3.0 support? I know @ccutrer was not in favor of that before.

If we were to follow the fact that the addon was only included in 3.2, technically we don't "need" to support anything lower than that. It has the benefit of reducing the testing duration.

boc-tothefuture:
> ```
>   Unresolved requirement: Import-Package: com.sun.nio.file; resolution:="optional"
>   Unresolved requirement: Import-Package: com.sun.security.auth.module; resolution:="optional"
>   Unresolved requirement: Import-Package: org.apache.tools.ant; resolution:="optional"
>   Unresolved requirement: Import-Package: org.bouncycastle.jce.provider; version="[1.52.0,2.0.0)"; resolution:="optional"
>   Unresolved requirement: Import-Package: org.joda.convert; resolution:="optional"
>   Unresolved requirement: Import-Package: sun.management; resolution:="optional"
>   Unresolved requirement: Import-Package: sun.nio.ch; resolution:="optional"
>   Unresolved requirement: Import-Package: jakarta.annotation
> ```
> 
> TBH, I don't really know what those are for and why they need to be in the pom.xml

I have no idea.  I copied the pom from the jython bundle and then during code review wborn changed it to this.    Bnd/osgi remains a mystery to me.

boc-tothefuture:
I am ok with dropping support for 3.0.  however, I don't want to disregard @ccutrer concerns.

ccutrer:
I'm +1 dropping 3.0 (and 3.1; @jimtng makes a good point) since we have a good reason now.  I'm just not in favor of dropping support for older things for no reason other than "testing it takes too long".

boc-tothefuture:
I would like to keep 3.1 around until 3.2 is released.

boc-tothefuture:
@jimtng feel free to drop 3.0 support.  

Should probably be a different PR?

jimtng:
I agree that ideally we should try to support as older version as possible, i.e. 3.0, and the only reason we'd have to drop 3.0 support is because of this bundle dependencies.

So right now we could either:
- drop 3.0 support, or
- revert the add-on bundle dependencies to what it was before the changes requested by @wborn. 

for the record, the dependencies that worked with 3.0 up to 3.2 were:

```
<bnd.importpackage>com.ibm.icu.*;resolution:=optional,org.abego.treelayout.*;resolution:=optional,org.apache.ivy.*;resolution:=optional,org.stringtemplate.v4.*;resolution:=optional</bnd.importpackage>
```

and now (not working on 3.0):
```
<bnd.importpackage>com.sun.nio.*;resolution:=optional,com.sun.security.*;resolution:=optional,org.apache.tools.ant.*;resolution:=optional,org.bouncycastle.*;resolution:=optional,org.joda.*;resolution:=optional,sun.management.*;resolution:=optional,sun.nio.*;resolution:=optional</bnd.importpackage>
```
How should we proceed?

boc-tothefuture:
Feel free to rebase and then the tests should  pass and we can merge.

jimtng:
rebased

jimtng:
how do I make it test against 3.2 as well without force pushing another commit?

jimtng:
@boc-tothefuture do you know why gem_install.feature is failing?

boc-tothefuture:
I am not 100% sure.. it looks like maybe it just hasn't started to install the Gems  yet?  You could push the wait period from 120 to 180?

Also it is installing version 3.x of the gem, we may want to update that to 4.x? Although not sure that matters.

boc-tothefuture:
@jimtng have you been able to try extending the test time?

jimtng:
> @jimtng have you been able to try extending the test time?

I haven't had a chance yet. I was working on logging, and got side tracked replacing my watering timer with tasmota and rewriting its rule. I'll look into this one today. I have a feeling it won't be as easy as increasing the timeout but I hope I'm wrong.

boc-tothefuture:
Does this pass locally for you?  I don't understand why a gem change would make a difference but the automation directory no longer exists which means there is no rule for it to test.  I am not quite sure what is going on here.  

Probably unrelated, but can you bump gem install to install >4.0 of the gem?

jimtng:
> Probably unrelated, but can you bump gem install to install >4.0 of the gem?

done.



> Does this pass locally for you?

Yes it does. Even with the previously lower timeout.

boc-tothefuture:
FYI - Gem install fails locally for me if I do a `bin/rake openhab:clobber` , `bin/rake openhab:warmup`, and `bin/cucumber features/gem_install.feature`

PR #404, opened 2021-11-18
fix(logging): multiple logging path fixes

### LINKED ISSUE #84: logger doesn't show the rule name

jimtng:
Fix #84 

- Logging at the top level will use the file name as the logger suffix as originally specified in the docs
- Logging inside a timer will use the same suffix as the top level logging, i.e. it will use the file name as the suffix
- The rule name will be used as the logger suffix when called inside a rule
- A custom suffix can be specified through `logger('custom suffix').<level>` 
- Clean up of redundant logger methods

### DIFF
--- a/lib/openhab/dsl/timers/timer.rb
+++ b/lib/openhab/dsl/timers/timer.rb
@@ -4,6 +4,7 @@ require 'java'
 require 'delegate'
 require 'forwardable'
 require 'openhab/log/logger'
+require 'openhab/core/thread_local'
 
 module OpenHAB
   module DSL
@@ -22,6 +23,7 @@ module OpenHAB
       # @since 2.0.0
       class Timer < SimpleDelegator
         include OpenHAB::Log
+        include OpenHAB::Core::ThreadLocal
         extend Forwardable
 
         def_delegator :@timer, :has_terminated, :terminated?
@@ -88,7 +90,10 @@ module OpenHAB
           proc {
             semaphore.synchronize do
               Timers.timer_manager.delete(self)
-              yield(self)
+              rule_name = "_timer_#{caller_locations(2)&.first&.path}_#{caller_locations(2)&.first&.lineno}"
+              thread_local(rule_name: rule_name) do
+                yield(self)
+              end
             end
           }
         end
--- a/lib/openhab/dsl/timers/timer.rb
+++ b/lib/openhab/dsl/timers/timer.rb
@@ -4,7 +4,6 @@ require 'java'
 require 'delegate'
 require 'forwardable'
 require 'openhab/log/logger'
-require 'openhab/core/thread_local'
 
 module OpenHAB
   module DSL
@@ -23,7 +22,6 @@ module OpenHAB
       # @since 2.0.0
       class Timer < SimpleDelegator
         include OpenHAB::Log
-        include OpenHAB::Core::ThreadLocal
         extend Forwardable
 
         def_delegator :@timer, :has_terminated, :terminated?
@@ -45,7 +43,6 @@ module OpenHAB
             @timer = ScriptExecution.createTimer(
               ZonedDateTime.now.plus(@duration), timer_block(semaphore, &block)
             )
-            @rule_timers = Thread.current[:rule_timers]
             super(@timer)
             Timers.timer_manager.add(self)
           end
@@ -90,10 +87,7 @@ module OpenHAB
           proc {
             semaphore.synchronize do
               Timers.timer_manager.delete(self)
-              rule_name = "_timer_#{caller_locations(2)&.first&.path}_#{caller_locations(2)&.first&.lineno}"
-              thread_local(rule_name: rule_name) do
-                yield(self)
-              end
+              yield(self)
             end
           }
         end
--- a/lib/openhab/log/logger.rb
+++ b/lib/openhab/log/logger.rb
@@ -139,7 +139,7 @@ module OpenHAB
     #
     def logger(name = nil)
       name ||= Thread.current[:logger_name]
-      return Log.logger_for(name) if name
+      return Log.logger_for(name.to_s.strip) if name
 
       Log.logger_for_class(self.class)
     end
@@ -158,8 +158,9 @@ module OpenHAB
            !klass.java_class.name.start_with?('org.jruby.Ruby')
           klass = klass.java_class
         end
-        name = klass.name
-        @loggers[name] ||= Log.logger_for(name)
+
+        name = klass.name.gsub('::', '.') unless klass.name == 'Object'
+        Log.logger_for(name)
       end
 
       #
@@ -170,30 +171,26 @@ module OpenHAB
       # @return [Logger] Logger for the supplied name
       #
       def logger_for(name)
-        log_prefix = Configuration.log_prefix
-        log_prefix += if name
-                        ".#{name}"
-                      else
-                        ".#{log_caller}"
-                      end
-        Logger.new(log_prefix)
+        log_prefix = [Configuration.log_prefix, rules_file, name].compact
+                                                                 .join('.')
+                                                                 .gsub(/\s+/, '_')
+        @loggers[log_prefix] ||= Logger.new(log_prefix)
       end
 
       private
 
       #
-      # Figure out the log prefix
+      # Figure out the rules file name
       #
       # @return [String] Prefix for log messages
       #
-      def log_caller
-        logger.error('LOGGING: log_caller called')
+      def rules_file
         caller_locations.map(&:path)
                         .grep_v(%r{openhab/core/})
                         .grep_v(/rubygems/)
                         .grep_v(%r{lib/ruby})
-                        .first
-                        .yield_self { |caller| File.basename(caller, '.*') }
+                        .last
+                        &.then { |caller| File.basename(caller, '.*') }
       end
     end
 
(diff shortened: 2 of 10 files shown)


jimtng:
One peculiar thing I noticed was that as I was tailing the logs in the openhab console, the log would say something like this:
```
08:56:55.008 [INFO ] [jsr223.jruby.rule_1                  ] - 1 received_command Switch1
```

However, I couldn't just copy paste 
```
[jsr223.jruby.rule_1                  ] - 1
```
into the cucumber rule. It won't match. I had to remove one space between `rule_1` and `]`, so that in the rule I'd have to use:
```
[jsr223.jruby.rule_1                 ] - 1
```

I couldn't figure out why. Perhaps Karaf console has a different padding width than what goes into the log file.

ccutrer:
> One peculiar thing I noticed was that as I was tailing the logs in the openhab console, the log would say something like this:
> 
> ```
> 08:56:55.008 [INFO ] [jsr223.jruby.rule_1                  ] - 1 received_command Switch1
> ```
> 
> However, I couldn't just copy paste
> 
> ```
> [jsr223.jruby.rule_1                  ] - 1
> ```
> 
> into the cucumber rule. It won't match. I had to remove one space between `rule_1` and `]`, so that in the cucumber rule I'd have to use:
> 
> ```
> [jsr223.jruby.rule_1                 ] - 1
> ```
> 
> I couldn't figure out why. Perhaps Karaf console has a different padding width than what goes into the log file.

That's usually a sign of a tab character, and console is converting to spaces, so when you copy it out it doesn't match what's actually in the file.

jimtng:
A timer may not necessarily belong to a specific rule. Yes it may be defined inside a rule, but it could be "rescheduled" by another rule. Furthermore, it could be defined in a top-level method that's called by multiple rules.

In this instance, I propose that the timer's filename and line number will be used e.g. `jsr223.jruby._timer_filename_123`

jimtng:
I'm having a difficult time getting a proper file location for the timer logger.

I will leave the timer logger as is for the time being and if need be, it can be done in a future PR. Meanwhile this PR should fix the logger name inside rules.

boc-tothefuture:
My local codebase handles the timers for the most part.  

Can we wait to merge this until I can push my branch up and then we can compare?

jimtng:
I fixed logging the rule file name, which covers the timer too. About to push it now. Juggling multiple commits in one PR, knowing that each commit's comment makes up the changelog, is a bit tricky though

lib/openhab/dsl/timers/timer.rb
```diff
@@ -43,7 +43,6 @@ def initialize(duration:, &block)
             @timer = ScriptExecution.createTimer(
               ZonedDateTime.now.plus(@duration), timer_block(semaphore, &block)
             )
-            @rule_timers = Thread.current[:rule_timers]
```
jimtng, line 46: I'm not sure why this was here.

jimtng:
> Can we wait to merge this until I can push my branch up and then we can compare?

There's a lot of changes in this PR though.

> Juggling multiple commits in one PR, knowing that each commit's comment makes up the changelog, is a bit tricky though

Should I just squash the commits under one feat(logging) and use multiple line commit descriptions to mention the fixes?

lib/openhab/log/logger.rb
```diff
@@ -148,38 +152,26 @@ class << self
...
-      def logger(klass)
+      def logger_for_class(klass)
         if klass.respond_to?(:java_class) &&
            klass.java_class &&
            !klass.java_class.name.start_with?('org.jruby.Ruby')
           klass = klass.java_class
         end
-        name = klass.name
+        name = klass.name if klass.name != 'Object'
         @loggers[name] ||= Log.logger_for(name)
```
boc-tothefuture, line 162: This memorization doesn't work because the class name can be shared between rules and therefore the logging won't be correct if they try and log for a specific rule.  The logger name needs to be calculated on each call.

lib/openhab/log/logger.rb
```diff
@@ -148,38 +152,26 @@ class << self
...
       #
       # @param [String] classname to create logger for
       #
-      # @return [Logger] Logger for the supplied classname
+      # @return [Logger] Logger for the supplied name
       #
-      def configure_logger_for(name)
+      def logger_for(name)
         log_prefix = Configuration.log_prefix
+        Logger.new('XXX').error(caller_locations.to_s) if name == 'Object'
```
boc-tothefuture, line 174: If name is Object, object should be removed from the logger name.

lib/openhab/log/logger.rb
```diff
@@ -198,8 +192,8 @@ def log_caller
                         .grep_v(%r{openhab/core/})
```
boc-tothefuture, line 192: log_caller is a confusing name for what we are really trying to do here.  I think rules_file is more descriptive.

boc-tothefuture: An issue with this code is that the rule_file name isn't always available - like for example if a timer thread is kicked off - it isn't tied to a rules file because that isn't its base.  However, every rules file is tied to a single JRuby context, therefore if we get a rule_file name once, you can memoize it at the class level so then every log entry can have the corresponding rules file in its prefix.

lib/openhab/log/logger.rb
```diff
@@ -148,38 +152,26 @@ class << self
...
-      # Configure a logger for the supplied classname
+      # Configure a logger for the supplied name
       #
       # @param [String] classname to create logger for
       #
-      # @return [Logger] Logger for the supplied classname
+      # @return [Logger] Logger for the supplied name
       #
-      def configure_logger_for(name)
+      def logger_for(name)
```
boc-tothefuture, line 172: loggers should be constructed in this format:
`jsr223.jruby.<rules_file>.<rule>.<any other classes>`

That lets users set log levels at a rules file level easily. 

The <rule> is optional in case in the logging happens outside of the context of a rule.

boc-tothefuture: the entire log name should have spaces replaced with underscores and ruby '::' replaced with '.'  That way people can specify logging levels across the entire logger name.

boc-tothefuture (changes requested):

lib/openhab/log/logger.rb
jimtng: Oops this was just for a quick debugging, not supposed to be included.

lib/openhab/log/logger.rb
jimtng: Does this look like what you had in mind:

```
23:28:07.857 [TRACE] [jsr223.jruby.log_test.Module         ] - Defining Java::OrgOpenhabCoreLibraryItems::ContactItem#open? for OPEN
23:28:07.858 [TRACE] [jsr223.jruby.log_test.Module         ] - Defining Java::OrgOpenhabCoreLibraryItems::ContactItem#closed? for CLOSED
23:28:07.995 [TRACE] [HAB.DSL.Between.TimeOfDayRangeElement] - SOD(0) other SOD(86399) Other Class (OpenHAB::DSL::Between::TimeOfDayRangeElement) Result (-1)
23:28:08.026 [TRACE] [jsr223.jruby.log_test.Module         ] - Loaded ACTION: org.openhab.core.model.script.actions.Audio
23:28:08.027 [TRACE] [jsr223.jruby.log_test.Module         ] - Loaded ACTION: org.openhab.core.model.script.actions.Ephemeris
23:28:08.028 [TRACE] [jsr223.jruby.log_test.Module         ] - Loaded ACTION: org.openhab.core.persistence.extensions.PersistenceExtensions
```

```
23:31:19.409 [INFO ] [jsr223.jruby.rule_file.rule_1        ] - A received_command Switch1
```

lib/openhab/log/logger.rb
jimtng: > An issue with this code is that the rule_file name isn't always available - like for example if a timer thread is kicked off - it isn't tied to a rules file because that isn't its base.

I haven't come across this scenario. Could you write an example code / cucumber test to demonstrate?

The only time I encountered where rule_file isn't available was inside scriptLoaded / scriptUnloaded, because they get invoked directly from java.

lib/openhab/log/logger.rb
jimtng: I think what you want is: `jsr223.jruby.rules_file.rule_name.Module`?

What other classes do you have in mind?

lib/openhab/log/logger.rb
jimtng: I moved the memorisation into `logger_for` just right before the actual Logging object is created. It should cover all cases, since the key is the actual `log_prefix` itself.

lib/openhab/log/logger.rb
jimtng: The latest push renamed it to `rules_file`

jimtng:
I've just come across another scenario where the current PR will fail. But before we get to it, I'm wondering what this mechanism is called, where `rule_var` is propagated into the scope of the block / proc even when that proc is running in a timer (another thread?). If we could use a similar mechanism, we won't even need to use a thread variable

```ruby
rule 'myname' do |rule_var|
  on_start
  run { logger.info("Hi, I'm running inside #{rule_var.name}") }
end
```

boc-tothefuture:
@jimtng see #439

PR #407, opened 2021-11-20
ci(regex): support regular expressions in log checks

### LINKED ISSUE #249: [tests] Allow regex to check log lines
Would it be possible to allow regex checks when reading log lines? I noticed e.g that the tests for error logging failed in recet OH versions because they (finally) changed from outputting `<script>` to the full path of the script. I could compensate for this by just testing for a small stub of the line, but would feel better if I could use
```ruby
And It should log /RUBY.test\(.*:8\)/ within 5 seconds
```
instead, to be more precise.

Would this be possible to add?

jimtng:
Resolve #249

### DIFF
--- a/features/step_definitions/openhab.rb
+++ b/features/step_definitions/openhab.rb
@@ -25,6 +25,13 @@ Then(/^It should log '([^']*)' within (\d+) seconds$/) do |string, seconds|
   end
 end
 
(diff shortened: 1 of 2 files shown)


jimtng:
@boc-tothefuture any objections to this PR? I could use this in the logging test

PR #410, opened 2021-11-20
add things[ThingUID] lookup and item.linked_thing and item.all_linked_things

ktgeek:
This patchset adds the ability to lookup a thing given a Channel UID by adding a by_channel methods to things. It also adds a helper method to GenericItem to get the Channel UID from the item's metadata, and then call things to find the Thing.

An example use case for this is when Thing configuration data contains information that would be useful in a rule. As an example, given a unlock keycode id from a zwave lock device:

```ruby
lock_thing = lock_item.linked_thing
keycode_owner_name = lock_thing.configuration.get("usercode_label_#{keycode_id}")
notify "#{lock_item.label} unlocked by user #{keycode_owner_name}"
```


boc-tothefuture:
Hi @ktgeek - Thanks for you first contribution!

I haven't looked at the code yet, did want to let you know that if it fails a build you don't need to close and resubmit.  

In your local branch you can make your changes then:
```
git commit -a --amend --no-edit  # Add to latest commit without changing commit message
git push --force-with-lease # push changes up
```
In your case you needed to change the commit message so just leave off the '--no-edit'

lib/openhab/dsl/things.rb
```diff
@@ -96,6 +96,18 @@ def [](uid)
         alias include? []
         alias key? []
 
+        # Gets a specific thing related to a channel in the format binding_id:type_id:thing_id:channel_id
+        # or by the ChannelUID object
+        # @return Thing specified by channel or nil if the thing does not exist in the thing registry
+        def by_channel(channel_uid)
+          channel_uid = org.openhab.core.thing.ChannelUID.new(channel_uid) if channel_uid.is_a?(String)
+          thing = $things.get(channel_uid.thing_uid) # rubocop: disable Style/GlobalVars
```
boc-tothefuture, line 104: This code seems almost the same as the lookup code above in `[]`.  Is the main difference in how it gets the thing_uid?

If so, is it worth refactoring to not have the duplicate code?

jimtng: You beat me to it... I was just about to suggest the same. Also channel_uid can be nil here, in which case, we should probably call `logger.error`

boc-tothefuture:
It seems useful to be able get the thing from an item, so I am +1 on this. 

@ktgeek - can you update the docs so people know this exists.  

Should the `by_channel` method be public? I think this needs another name.  Maybe `from_channel`?  @jimtng @ccutrer ?

jimtng:
In my environment, I have items linked to multiple channels.

boc-tothefuture:
My mental model must be off.  I can see how you would link a channel to multiple items - what is the scenario where multiple items are linked to different channels?

jimtng:
> My mental model must be off. I can see how you would link a channel to multiple items - what is the scenario where multiple items are linked to different channels?

Example: I have 2 hallway lights:

```
Thing mqtt:topic:mosquitto:hallway_light_1 "Hallway Light1" (mqtt:broker:mosquitto)
Thing mqtt:topic:mosquitto:hallway_light_2 "Hallway Light2" (mqtt:broker:mosquitto)
```

I define 1 item:
```
Switch Hallway_Light_Power { channel="mqtt:topic:mosquitto:hallway_light_1:power", channel="mqtt:topic:mosquitto:hallway_light_2:power", autoupdate="false" }
```

Sending a command to the item will send it to both things. I _think_ the status coming back from the things are ORed.

Yes it is debatable whether this is a good idea versus having two items, and sending a command to the group, nevertheless, it is perfectly valid and supported to have an item linked to multiple things.

boc-tothefuture:
Ok, that helps me understand it.

jimtng:
The problem is, `Item.meta` only returns the first channel

ktgeek:
> I haven't looked at the code yet, did want to let you know that if it fails a build you don't need to close and resubmit.

Thanks for that, I'll use that in the future.  I was fighting with git and GitHub and needed to leave, and I really wanted to get this up and get feedback.

lib/openhab/dsl/things.rb
ktgeek: I was trying not to conflate concerns, but that shouldn't be hard.  I guess if a string was passed in I can look for more than 3 colons as a sign that we're dealing with a channel.  I'll work that in tomorrow.

ktgeek:
> Should the `by_channel` method be public? I think this needs another name. Maybe `from_channel`? @jimtng @ccutrer ?

I don't see why it shouldn't be public. (Although, given the other thread it might be moot.)  I went back and forth with "from_channel" and "by_channel" a few times and ended up with by_channel as that felt closer to things I've done in rails, and "from" felt more more java to me.

ktgeek:
> The problem is, `Item.meta` only returns the first channel

Similar to what @boc-tothefuture said, thanks for the education.  I had made a bad assumption there.

Doing some quick research, it looks like for sending you're ok, but receiving status updates might not be an OR, but will actually be "last in wins."  That probably amounts to OR anyway in most cases, but there's probably subtle behavior differences that I can't think of off the top of my head.

That said, I can probably go via links and change those methods to return an array. For most users, they could just use the first result, and still saved some monkeying around.  I'm not sure if I'll get to this tomorrow, but it'll definitely be in the next week.

(And I'll make sure to update the docs as well.)

ktgeek:
> That said, I can probably go via links

Looking at the javaside of the ruby scripting, I think we might want to expose the ItemChannelLinkRegistry as a global, similar to the item registry and the thing registry is now.  itemChannelLinkRegistry.getBoundChannels("item_name") seems like exactly what we want.

@boc-tothefuture, any thoughts?

jimtng:
> I think we might want to expose the ItemChannelLinkRegistry as a global, similar to the item registry and the thing registry is now. itemChannelLinkRegistry.getBoundChannels("item_name") seems like exactly what we want.

+1 

> Doing some quick research, it looks like for sending you're ok, but receiving status updates might not be an OR, but will actually be "last in wins." That probably amounts to OR anyway in most cases, but there's probably subtle behavior differences that I can't think of off the top of my head.

"last in wins" seems to be the case indeed. 

Example, I have 3 items:
1. Light_Power (linked to both light1 and light2)
2. Light1_Power 
3. Light2_Power

Light_Power << ON results in Light1_Power and Light2_Power to turn on

Light1_Power << OFF results in Light_Power to become OFF, even though Light2_Power is still ON

jimtng:
> That said, I can probably go via links and change those methods to return an array. For most users, they could just use the first result, and still saved some monkeying around.

Is there some sort of Ruby construct that could make this a bit more elegant?

I understand that for 99.9% of the cases out there, you'll be dealing with just one channel. To sacrifice the simplicity to support multiple channel isn't a great solution.

I'd want `lock_item.thing` instead of having to use `lock_item.things.first` all the time.

Alternatively we could just support the case with a single thing, and not support multiple things for this `item.thing`

ktgeek:
> > That said, I can probably go via links and change those methods to return an array. For most users, they could just use the first result, and still saved some monkeying around.
> 
> Is there some sort of Ruby construct that could make this a bit more elegant?

I'll have to think about it some.  Nothing is jumping out at me except returning a single element or an array, and I don't like that for a return value for a single method.

> I understand that for 99.9% of the cases out there, you'll be dealing with just one channel. To sacrifice the simplicity to support multiple channel isn't a great solution.
> 
> I'd want `lock_item.thing` instead of having to use `lock_item.things.first` all the time.

+1

> Alternatively we could just support the case with a single thing, and not support multiple things for this `item.thing`

The more I think about it, the metadata approach I took will work most of the time, but if you look at my test code, it is not as full proof as using the link registry. No matter what I should shift gears towards using that. (First I need to figure out how to expose the registry on the java side, I'm a bit lost at the moment about how the existing globals are actually matched up after being named in the global set.)

The only idea I have right now is to do both a .thing and a .things method. The first would always do an implicit .first and send the first single item back, the later would send an array at all times.  I'm not wild about that either, but it will serve both uses cases in a way that won't confuse the end users.

jimtng:
> The only idea I have right now is to do both a .thing and a .things method. The first would always do an implicit .first and send the first single item back, the later would send an array at all times. I'm not wild about that either, but it will serve both uses cases in a way that won't confuse the end users.

+1

boc-tothefuture:
`thing` and `things` makes sense to me. 

The binding wouldn't add the Link registry, it would be a PR upstream, the binding would just make sure it was a global.

For now, can you get it via osgi?

lib/openhab/dsl/items/generic_item.rb
```diff
@@ -131,6 +131,23 @@ def groups
...
+        #
+        # @return [Thing] The thing associated with this item or nil
+        def thing
+          all_things.first
+        end
+
+        # Returns all of the item's associated things.
+        #
+        # @return [Array] An array of things or an empty array
+        def all_things
```
ktgeek, line 145: I had to name this all_things instead of the preferred things. I was getting a namespace collision with things[] that managed to put my openHAB instance into a bad state. I would love a better name suggetion

ktgeek:
I still need to add documentation, but I wanted to get feedback on this approach. I just removed the by_channel all together since it was just a few lines of logic in generic_item

lib/openhab/dsl/items/generic_item.rb
ccutrer: Hmm. I know we're probably leaking `Object.things` onto too many things, but when you define it on GenericItem it should be hiding the inherited method.

lib/openhab/dsl/things.rb
```diff
@@ -86,7 +86,7 @@ class Things
         # Gets a specific thing by name in the format binding_id:type_id:thing_id
         # @return Thing specified by name or nil if name does not exist in thing registry
         def [](uid)
-          thing_uid = org.openhab.core.thing.ThingUID.new(*uid.split(':'))
+          thing_uid = uid.is_a?(String) ? org.openhab.core.thing.ThingUID.new(*uid.split(':')) : uid
```
ccutrer, line 89: Maybe switch this. If it is a UID, do nothing. Otherwise, to_s and split it. Thinking of StringState, which isn't actually a string, but is implicitly convertible to one

lib/openhab/dsl/items/generic_item.rb
```diff
@@ -131,6 +131,23 @@ def groups
...
+          registry = OpenHAB::Core::OSGI.service('org.openhab.core.thing.link.ItemChannelLinkRegistry')
+          channels = registry.get_bound_channels(name).to_a
+          channels.map(&:get_thing_uid).uniq.map { |tuid| things[tuid] }
```
ccutrer, line 148: I think you can drop the get_ cause jruby

ktgeek:
> The binding wouldn't add the Link registry, it would be a PR upstream, the binding would just make sure it was a global.

I needed to read the jsr223 scripting documentation again and it clicked where everything was coming from.

I also don't yet see a need to make it global or default in the JRuby binding as it seems outside the average use. (Which is probably also why its not a default jsr223 variable.)
 
> For now, can you get it via osgi?

That's what I ended up doing.  Looking at the time when you left the comment, I was full steam into getting tests around the new code. :)

lib/openhab/dsl/items/generic_item.rb
ktgeek: That's probably what the problem was. Items were calling this things method when they really wanted Object.things and odd things were happening.

I had though maybe "all_linked_things" to really separate it out.

lib/openhab/dsl/items/generic_item.rb
ktgeek: :facepalm: I knew that from you saying that to me last week, and I just got tunnel vision

lib/openhab/dsl/items/generic_item.rb
ccutrer: Ooh I like that. `linked_thing` and `all_linked_things` for the method names. Definitely gives these methods a closer association with how links show up in the UI.

lib/openhab/dsl/things.rb
ktgeek: +1

features/items.feature
```diff
@@ -225,3 +225,53 @@ Feature:  items
...
+  Scenario: Item returns nil for an unlinked item
+    Given items:
+      | type | name |
+      | Number | Number1 |
+    And code in a rules file
+      """
+      logger.info("Thing: #{Number1.linked_thing}")
+      """
+    When I deploy the rules file
+    Then It should log "Thing: " within 5 seconds
```
ccutrer, line 238: You should probably use inspect and check for nil in the log line. Otherwise this test will always match regardless of what's logged

features/support/openhab_rest.rb
```diff
@@ -107,4 +107,12 @@ def self.item_groups(body, item)
     body[:groupNames] = groups unless groups.empty?
     body[:groupType] = item.group_type unless item.group_type.to_s.empty?
   end
+
+  def self.link_item(item_name:, channel_uid:)
+    body = { itemName: item_name, channelUID: channel_uid }
+    escaped_channel_uid = URI.encode_www_form_component(channel_uid)
+    put("/rest/links/#{item_name}/#{escaped_channel_uid}", headers: json, body: body.to_json)
+    # For some reason linking via the api doesn't create the metadata, this ensures it exists
+    put("/rest/items/#{item_name}/metadata/channel", headers: json, body: { value: channel_uid }.to_json)
```
jimtng, line 116: Do we still need to add the metadata? I tried adding an item through the GUI and linked it to a thing, and then tried accesing the A_Test.meta['channel']&.value and it returned nil.

features/thing.feature
```diff
@@ -24,6 +24,14 @@ Feature:  thing
...
 
+  Scenario: Things support [] lookup using a ThingUID
+    Given code in a rules file
+      """
+      logger.info("Thing: #{things[org.openhab.core.thing.ThingUID.new('astro:sun:home')].uid}")
+      """
+    When I deploy the rules file
+    Then It should log 'Thing: astro:sun:home' within 5 seconds
+
   Scenario Outline: Rule supports thing status changes for changed and updated
```
jimtng, line 35: This should be in a separate commit along with the associated code changes in `things.rb`. It would therefore create a separate entry in the changelog to note this change, as a `feat`

lib/openhab/dsl/things.rb
```diff
@@ -86,8 +86,8 @@ class Things
         # Gets a specific thing by name in the format binding_id:type_id:thing_id
         # @return Thing specified by name or nil if name does not exist in thing registry
         def [](uid)
-          thing_uid = org.openhab.core.thing.ThingUID.new(*uid.split(':'))
-          thing = $things.get(thing_uid) # rubocop: disable Style/GlobalVars
+          uid = org.openhab.core.thing.ThingUID.new(*uid.split(':')) unless uid.is_a?(org.openhab.core.thing.ThingUID)
+          thing = $things.get(uid) # rubocop: disable Style/GlobalVars
           return unless thing
```
jimtng, line 91: See note above

features/thing.feature
```diff
@@ -24,6 +24,14 @@ Feature:  thing
     When I deploy the rules file
     Then It should log 'Thing: astro:sun:home' within 5 seconds
 
+  Scenario: Things support [] lookup using a ThingUID
+    Given code in a rules file
+      """
+      logger.info("Thing: #{things[org.openhab.core.thing.ThingUID.new('astro:sun:home')].uid}")
+      """
+    When I deploy the rules file
+    Then It should log 'Thing: astro:sun:home' within 5 seconds
```
jimtng, line 33: This should be in a separate commit along with the associated code changes in `things.rb`. It would therefore create a separate entry in the changelog to note this change, as a `feat`

lib/openhab/dsl/things.rb
```diff
@@ -86,8 +86,8 @@ class Things
         # Gets a specific thing by name in the format binding_id:type_id:thing_id
         # @return Thing specified by name or nil if name does not exist in thing registry
         def [](uid)
-          thing_uid = org.openhab.core.thing.ThingUID.new(*uid.split(':'))
-          thing = $things.get(thing_uid) # rubocop: disable Style/GlobalVars
+          uid = org.openhab.core.thing.ThingUID.new(*uid.split(':')) unless uid.is_a?(org.openhab.core.thing.ThingUID)
+          thing = $things.get(uid) # rubocop: disable Style/GlobalVars
```
jimtng, line 90: See note above

features/support/openhab_rest.rb
ktgeek: Yeah, this can go. In the current implementation it doesn't matter anymore.

ktgeek:
With that last push, I think I should have everything addressed from the earlier feedback and added documentation.

lib/openhab/dsl/items/generic_item.rb
```diff
@@ -131,6 +131,23 @@ def groups
...
+        def linked_thing
+          all_linked_things.first
+        end
+
+        # Returns all of the item's linked things.
+        #
+        # @return [Array] An array of things or an empty array
+        def all_linked_things
+          registry = OpenHAB::Core::OSGI.service('org.openhab.core.thing.link.ItemChannelLinkRegistry')
+          channels = registry.get_bound_channels(name).to_a
```
ktgeek, line 147: In my OpenHAB environment, it errors out if I do "registry.bound_channels", there's probably some JRuby magic we're missing there. In any case, that's why this is still a get_ call

ccutrer: I think if the method takes any arguments, it doesn't treat it as a trivial accessor doesn't generate the non-`get_` method

docs/usage/items.md
```diff
@@ -54,18 +54,20 @@ end
...
+| command           | alias for shovel operator (<<) - has optional arguments 'for:, on_expire:' see timed command below | `VirtualSwitch.command(ON)`                                  |
+| ensure            | Only send following command/update if the item is not already in the requested state               | `VirtualSwitch.ensure.on`                                    |
+| update            | Sends update to an item                                                                            | `VirtualSwitch.update(ON)`                                   |
+| id                | Returns label or item name if no label                                                             | `logger.info(#{item.id})`                                    |
+| linked_thing      | Returns the item's thing. If multiple, returns first. If non, returns nil.                         | `logger.info(#{item.linked_thing.uid})`                      |
+| all_linked_things | Returns an array of all things item is linked with                                                 | `logger.info(#{item.all_linked_things.map(&:uid)})`          |
```
jimtng, line 65: As these are more esoteric methods that are less often used than other methods below, should we move them to the bottom of the list? @boc-tothefuture wdyt?

ktgeek: I had placed these before all the ? methods (and the related state) but I'm open to putting these wherever you folks think best.

jimtng: @ktgeek leave it as is for now unless others say otherwise. I'm merely making an uneducated suggestion. I don't know what the convention is.

ccutrer: I'm fine where they are. Though I do wonder why we go through the trouble of documenting individual methods on the markdown, and _also_ commenting the source so that they render properly for yardoc (https://rdoc.info/gems/openhab-scripting). I feel like the markdown docs are fine for narrative and example, but for details of methods we should be focusing on writing good comments so they render well. I put some amount of effort into that with the types and items refactor.

jimtng: Thanks for that info. As a newbie, I didn't know where "yard docs" went and didn't know that it can be viewed so nicely on a page somewhere. Could/should the yard docs be integrated into what I consider the primary documentation which is at https://boc-tothefuture.github.io/openhab-jruby/ ? Alternatively provide a link to the rdoc.info page.

boc-tothefuture:
I will go ahead and merge this. I do still prefer 'thing' and 'things' to 'linked_thing' and 'all_linked_things'.  If I am using an item, I know the things are linked because it is an item.  We can alias the methods later if we can reach agreement on that.

ktgeek:
> I will go ahead and merge this. I do still prefer 'thing' and 'things' to 'linked_thing' and 'all_linked_things'. If I am using an item, I know the things are linked because it is an item. We can alias the methods later if we can reach agreement on that.

Yay!

Even if we dropped `linked` from the method names we had a problem with `things` as a name because that was overriding the `things` on Object and causing a lot of bad side effects since the return wasn't what was expected. The current name is the least worst solution until things is scoped differently.

But still, yay!

PR #414, opened 2021-11-21
feat(trigger): support creating triggers for any trigger type

jimtng:
fix(guard): include attachment in guard's event. Previously the attachment isn't included in the `event` object for the guard.

[PID Controller Automation](https://www.openhab.org/addons/automation/pidcontroller/) and other add-ons will require specifying a custom trigger type uid. This PR will enable the creation of such trigger.

I have two commits here so that they both show up in the changelog as two separate entries. The guard fix depends on the changes in the trigger commit, hence why I'm including it in the same PR.


jimtng:
Hmm it's failing on OH 3.0 because pid controller automation isn't on it. Should I just remove that test? Can we run certain tests on certain OH version?

boc-tothefuture:
I didn't know generic triggers existed.

Should we drop the configuration part?  Just use double splat?

boc-tothefuture:
We would have to start using tags in the features.

boc-tothefuture:
Is there a way to test it without pid automation?  Since we are testing genetic triggers and not specifically pid automation?

jimtng:
> Is there a way to test it without pid automation? Since we are testing genetic triggers and not specifically pid automation?

We do need to test against pidautomation - or perhaps pwm controller automation (which was also introduced in OH3.1), because this PR does something different with the run event block. In pidautomation for example,`inputs` does not contain an `event` object. However, `inputs` contains `command` which is an important output data from pidautomation.

In such case, this PR creates a Struct with `attachment` and `command` (from `inputs`), that way our run { |event| .... } event variable has access to this `command` value through `event.command`.

I am wondering though whether we need to generically include all the fields from `inputs` (except those prefixed with the trigger id) because _maybe_ in the future other addons could use other variables. I haven't looked deeply enough to ascertain this.

For now, this only specifically grabs `command` from `inputs`.

So if we're going to drop 3.0, I'll create a separate PR for that, merge that, then merge this PR afterwards.

jimtng:
> Should we drop the configuration part? Just use double splat?

Thanks for the suggestion. I didn't know about double splat. I've now converted it to use double splat. Much nicer syntax, plus a hash is still possible.

lib/openhab/dsl/rules/automation_rule.rb
```diff
@@ -87,21 +87,38 @@ def cleanup
...
+          add_attachment(event, inputs)
+        end
+
         #
         # Get the trigger_id for the trigger that caused the rule creation
         #
         # @return [Hash] Input hash potentially containing trigger id
         #
         def trigger_id(inputs)
-          inputs&.keys&.grep(/\.event$/)&.first&.chomp('.event')
```
jimtng, line 104: @boc-tothefuture I'd like to ask specifically about this. Was there a particular reason you chose to use `<triggerid>.event` to extract the trigger id instead of `module` below? I wasn't really sure whether there are cases where `module` doesn't exist perhaps? I tried running all the trigger tests locally, and they seem to pass using `module`, but I thought I should check with you first.

boc-tothefuture: My locale branch is on module.  I discovered that when I added support for attachments on cron triggers..

The only thing I can think of is that I either missed it or it wasn't there before 3.1.

jimtng: Thanks for the confirmation. I'll stick with module.

boc-tothefuture:
LGTM

boc-tothefuture:
@jimtng  question for you and I guess @ccutrer  on this.  I was looking at [this](https://www.openhab.org/docs/configuration/jsr223.html) and the bottom 'generic trigger' section.  This seems different from a generic trigger.  

My question/proposal.  Should the keyword for this not be `trigger` and instead be something like `custom` or something else? I think `trigger` should probably be reserved for the generic trigger where you specify the topic, source, and events.

thoughts?

jimtng:
> the generic trigger where you specify the topic, source, and events.

I might not be aware of this. Could you please clarify what you are referring to as the 'generic trigger'?

boc-tothefuture:
> > the generic trigger where you specify the topic, source, and events.
> 
> I might not be aware of this. Could you please clarify what you are referring to as the 'generic trigger'?

Take a look at the link I posted in the comment at the bottom there is an openhab type 'core.GenericEventTrigger'

jimtng:
oops sorry I should've checked that link first.

I don't know what the "generic trigger" (i.e. with trigger type uid of `core.GenericEventTrigger`) is used for? 

I have no problems changing the naming in this PR but, perhaps the `core.GenericEventTrigger` can be implemented as `generic_event` instead, which I think is more descriptive. It's a bit confusing with the two concepts seemingly overlapping.

This PR allows you to create any trigger type, including creating a `core.GenericEventTrigger` trigger. `custom` just doesn't sound right. Maybe there's a better name?

boc-tothefuture:
> oops sorry I should've checked that link first.
> 
> I don't know what the "generic trigger" (i.e. with trigger type uid of `core.GenericEventTrigger`) is used for?
> 
> I have no problems changing the naming in this PR but, perhaps the `core.GenericEventTrigger` can be implemented as `generic_event` instead, which I think is more descriptive. It's a bit confusing with the two concepts seemingly overlapping.
> 
> This PR allows you to create any trigger type, including creating a `core.GenericEventTrigger` trigger. `custom` just doesn't sound right. Maybe there's a better name?

The generic trigger (from what I can tell) lets you listen to anything that goes across the event bus.  I was looking at using it now for doing something when an item is added, changed, or deleted in OpenHAB.   Perhaps a better keyword for 'core.GenericEventTrigger' would be 'event' as it is more descriptive of what you are trying to trigger on.

Maybe trigger is fine for this type, but would like @ccutrer or some others to weigh in too.

jimtng:
> The generic trigger (from what I can tell) lets you listen to anything that goes across the event bus. I was looking at using it now for doing something when an item is added, changed, or deleted in OpenHAB. Perhaps a better keyword for 'core.GenericEventTrigger' would be 'event' as it is more descriptive of what you are trying to trigger on.

That would be handy! I'd throw in `busevent` and `bus_event` as candidate names, because we customarily use `run do |event|` so that might become confusing.

PR #417, opened 2021-11-22
fix(comparisons): === should perform an exact type match for case statements

### LINKED ISSUE #365: Case comparisons are sensitive to specific sequence
Here are some tests that are currently failing. 

```
  Scenario Outline: event.command for Color item can be compared to accepted command types
    Given items:
      | type  | name  |
      | Color | Item1 |
    And a rule
      """
      rule 'handle command' do
        received_command Item1
        run do |event|
          logger.info('We sent a command <command>')
          logger.info("event.command.class: #{event.command.class}")
          log = case event.command
                when <cond1> then 'cond1'
                when <cond2> then 'cond2'
                when <cond3> then 'cond3'
                else 'cond-else'
                end
          logger.info("case match: #{log}")
        end
      end
      """
    When I deploy the rule
    And item "Item1" state is changed to "<command>"
    Then It should log 'case match: <log>' within 5 seconds
    Examples:
      | command  | cond1       | cond2             | cond3    | log   |
      | ON       | ON          | 1..100            | INCREASE | cond1 |
      | ON       | 0..99,100   | OFF               | ON       | cond3 |
      | OFF      | OFF         | INCREASE,DECREASE | 0        | cond1 |
      | OFF      | 1..100      | 0                 | OFF      | cond3 |
      | INCREASE | ON,DECREASE | 0..100            | INCREASE | cond3 |
      | DECREASE | OFF         | 0,INCREASE        | DECREASE | cond3 |
      | 0        | OFF         | DECREASE          | 0        | cond3 |
      | 0        | ON          | DECREASE          | 0..50    | cond3 |
      | 50       | ON          | INCREASE          | 50       | cond3 |
      | 100      | ON          | 0..99             | 100      | cond3 |
      | 100      | ON          | 0..99             | 100      | cond3 |
      | 1,2,3    | 1,2,3       | 0.100             | OFF      | cond1 |
      | 1,2,3    | ON          | 0.100             | '1,2,3'  | cond3 |

  Scenario Outline: event.command for Player item can be compared to accepted command types
    Given items:
      | type   | name  |
      | Player | Item1 |
    And a rule
      """
      rule 'handle command' do
        received_command Item1
        run do |event|
          logger.info('We sent a command <command>')
          logger.info("event.command.class: #{event.command.class}")
          log = case event.command
                when <cond1> then 'cond1'
                when <cond2> then 'cond2'
                when <cond3> then 'cond3'
                else 'cond-else'
                end
          logger.info("case match: #{log}")
        end
      end
      """
    When I deploy the rule
    And item "Item1" state is changed to "<command>"
    Then It should log 'case match: <log>' within 5 seconds
    Examples:
      | command | cond1 | cond2                            | cond3 | log   |
      | PLAY    | PLAY  | 1..100                           | ON    | cond1 |
      | PLAY    | PAUSE | NEXT,PREVIOUS,REWIND,FASTFORWARD | PLAY  | cond3 |

  Scenario Outline: event.command for Rollershutter item can be compared to accepted command types
    Given items:
      | type          | name  |
      | Rollershutter | Item1 |
    And a rule
      """
      rule 'handle command' do
        received_command Item1
        run do |event|
          logger.info('We sent a command <command>')
          logger.info("event.command.class: #{event.command.class}")
          log = case event.command
                when <cond1> then 'cond1'
                when <cond2> then 'cond2'
                when <cond3> then 'cond3'
                else 'cond-else'
                end
          logger.info("case match: #{log}")
        end
      end
      """
    When I deploy the rule
    And item "Item1" state is changed to "<command>"
    Then It should log 'case match: <log>' within 5 seconds
    Examples:
      | command | cond1     | cond2     | cond3   | log   |
      | UP      | 0         | 1..100    | DOWN,UP | cond3 |
      | UP      | 100       | STOP,MOVE | UP      | cond3 |
      | STOP    | STOP      | MOVE      | 0       | cond1 |
      | MOVE    | STOP      | UP,DOWN   | MOVE    | cond3 |
      | 0       | STOP,MOVE | UP,DOWN   | 0       | cond3 |
      | 100     | STOP,MOVE | 100       | DOWN    | cond2 |
```

PS I didn't include DimmerItem because I figured the tests for ColorItem already cover the same types for it.

_Originally posted by @jimtng in https://github.com/boc-tothefuture/openhab-jruby/issues/332#issuecomment-953370755_

jimtng:
Resolve #365 

**This could potentially be a breaking change.**

`case` comparisons must now use the underlying type.

Case in point:
```ruby
RollerShutterItem1 << DOWN
case RollerShutterItem1
when DOWN then logger.info("It will no longer reach here")
when 100 then logger.info("It will reach here: 100")
end
```

I believe this is the correct / non ambiguous behaviour. It also allows precision handling of commands sent to items that support multiple types, which is the whole reason for this PR.

Example:
```ruby
rule 'x' do
  received_command RollerShutterItem1
  run do |event|
    case event.command
    when 0..100 then logger.info("A specific position has been commanded")
    when DOWN then logger.info("DOWN command received")
    end
  end
end

RollerShutterItem1 << DOWN
```

The following will still evaluate to true, as it's not affected by this PR: `RollerShutterItem1 == 100`, `RollerShutterItem1 == DOWN` or `RollerShutterItem1.down?`

### DIFF
--- a/features/groups.feature
+++ b/features/groups.feature
@@ -125,10 +125,10 @@ Feature:  groups
 
   Scenario: Groups work in case statements
     Given groups:
-      | name         | type          | function | params       |
-      | Switches     | Switch        | AND      | ON, OFF      |
-      | Contacts     | Contact       | OR       | OPEN, CLOSED |
-      | Shutters     | Rollershutter | AND      | UP, DOWN     |
+      | name     | type          | function | params       |
+      | Switches | Switch        | AND      | ON, OFF      |
+      | Contacts | Contact       | OR       | OPEN, CLOSED |
+      | Shutters | Rollershutter | AND      | UP, DOWN     |
 
     And items:
       | type          | name       | label       | state  | groups   |
@@ -152,8 +152,8 @@ Feature:  groups
       end
 
       case Shutters
-      when UP then logger.info('All shutters are UP')
-      when DOWN then logger.info('At least one shutter is not UP')
+      when 0 then logger.info('All shutters are UP')
+      when 100 then logger.info('At least one shutter is not UP')
       end
       """
     When I deploy the rules file
@@ -222,7 +222,7 @@ Feature:  groups
       | name     | type    |
       | Switches | Contact |
     Then It should log "false" within 10 seconds
-  
+
   Scenario: Can add lazy arrays together
     Given code in a rules file
       """
--- a/lib/openhab/dsl/types/type.rb
+++ b/lib/openhab/dsl/types/type.rb
@@ -42,6 +42,20 @@ module OpenHAB
         end
 
         #
+        # Case equality
+        #
+        # @return [Boolean] if the values are of the same Type
+        #                   or item state of the same type
+        #
+        def ===(other)
+          logger.trace("(#{self.class}) #{self} === #{other} (#{other.class})")
+          other = other.state if other.respond_to?(:state)
+          return false unless instance_of?(other.class)
+
+          eql?(other)
+        end
+
+        #
         # Check equality, including type conversion
         #
         # @return [Boolean] if the same value is represented, including
--- a/features/between.feature
+++ b/features/between.feature
@@ -81,60 +81,61 @@ Feature: between
       | '<%=(Time.now + (5*60)).strftime('%H:%M:%S')%>'..'<%=(Time.now + (10*60)).strftime('%H:%M:%S')%>' | Time.now | should not |
 
 
-    Scenario Outline: Between supports Day of Month 
+  Scenario Outline: Between supports Day of Month
     Given a rule template:
       """
       require 'date'
       range = between <between>
-      in_range = range.include?(<compare>) 
+      in_range = range.include?(<compare>)
       logger.info("in range: #{in_range}")
       """
     When I deploy the rule
     Then It should log "in range: <result>" within 5 seconds
     Examples: Checks in range, before range and after range
-      | between                                                                                           | compare                                | result   |
-      | '<%=Date.today.prev_day.strftime('%m-%d')%>'..'<%=Date.today.next_day.strftime('%m-%d')%>'        | Date.today                             | true     |
-      | '<%=(Date.today + 10 ).strftime('%m-%d')%>'..'<%=(Date.today + 20).strftime('%m-%d')%>'           | Date.today                             | false    |
-      | '<%=(Date.today - 20 ).strftime('%m-%d')%>'..'<%=(Date.today - 10).strftime('%m-%d')%>'           | Date.today                             | false    |
+      | between                                                                                    | compare    | result |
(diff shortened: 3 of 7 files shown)


boc-tothefuture:
Makes sense to me. 
@ccutrer ?

jimtng:
I still have a few kinks to work out to pass all tests, but the general concept is the same.

lib/openhab/dsl/types/type.rb
```diff
@@ -41,6 +41,20 @@ def eql?(other)
...
+        #
+        # Case equality
+        #
+        # @return [Boolean] if the values are of the same Type
+        #                   or item state of the same type
+        #
+        def ===(other)
+          logger.trace("(#{self.class}) #{self} === #{other} (#{other.class})")
+          other = other.state if other.respond_to?(:state)
+          return false unless is_a?(other.class)
```
ccutrer, line 53: instead of calling is_a? (you should really be using instance_of? in this case) and calling super, just call `eql?(other)` at this point

ccutrer (commented):
while it's a bit opposite of normal (usually === is broader than ==) it makes sense in this case.

lib/openhab/dsl/types/type.rb
jimtng: Thanks for the suggestions. I'll implement them in the next push.

features/groups.feature
```diff
@@ -152,8 +152,8 @@ Feature:  groups
       end
 
       case Shutters
-      when UP then logger.info('All shutters are UP')
-      when DOWN then logger.info('At least one shutter is not UP')
+      when 0 then logger.info('All shutters are UP')
+      when 100 then logger.info('At least one shutter is not UP')
```
jimtng, line 156: This part here is a bit counter-intuitive. The comparison in a case statement for item `state` (as opposed to `command`) also requires an exact type match, in this case, because the state's type is PercentType, it must be compared against PercentType.

jimtng:
This is ready for review

ccutrer (approved):

features/groups.feature
boc-tothefuture: Does that need to be in the docs so we and others don't get confused?

jimtng: > Does that need to be in the docs so we and others don't get confused?

Ideally, we'll have a page that comprehensively explains about how state comparisons work in various different ways. For example, how and why PercentType.new(1).on? returns true vs PercentType.new(1) === ON returns false. 

I'll create a separate PR for that because I may not think of all possible scenarios.

jimtng:
Should this be flagged as a breaking change?

boc-tothefuture:
> Should this be flagged as a breaking change?

That is a judgement call.  Is this a fix of something broken or a broken change.  The API surface isn't changing so I am inclined to call it a fix, but will defer to you.

jimtng:
> > Should this be flagged as a breaking change?
> 
> That is a judgement call. Is this a fix of something broken or a broken change. The API surface isn't changing so I am inclined to call it a fix, but will defer to you.

I'm happy to call it a fix too.

PR #432, opened 2021-11-29
feat(switch): accept boolean values

### LINKED ISSUE #400: Switch Items should accept ruby Booleans
Currently you need to convert boolean into ON/OFF
I think intuitively True == ON and False == OFF

So:
```
Squeeze_Home_Audio_A_Power << (ZoneSources.members.any? { |source| source == 6 } ? ON : OFF)
```

Becomes:
```
Squeeze_Home_Audio_A_Power << ZoneSources.members.any? { |source| source == 6 }
```

boc-tothefuture:
Closes #400

### DIFF
--- a/features/switch_item.feature
+++ b/features/switch_item.feature
@@ -10,7 +10,7 @@ Feature:  switch_item
       | Switch | TestSwitch | Test Switch | <initial_state> |
     And code in a rules file
       """
-      # Invert all switches
+     # Invert all switches
       items.select { |item| item.is_a? Switch }
            .each   { |switch| if switch.off? then switch.on else switch.off end}
       """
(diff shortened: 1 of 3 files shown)


features/string_item.feature
```diff
@@ -14,6 +14,7 @@ Feature:  string_item
     Given item "StringOne" state is changed to "Hello"
     And code in a rules file
       """
+      pp StringOne.class.ancestors
```
ccutrer, line 17: debugging?

lib/openhab/dsl/items/switch_item.rb
```diff
@@ -20,6 +20,23 @@ def truthy?
           on?
         end
 
+        #
```
ccutrer, line 23: I would highly recommend using format_type instead of overriding command directly. otherwise it won't be able to work with `ensure`

ccutrer (changes requested):

lib/openhab/dsl/items/switch_item.rb
boc-tothefuture: I think it did work with ensure when I printed the ancestors it seems correct.

I will check out format command as that might cleaner.

features/switch_item.feature
```diff
@@ -21,6 +21,24 @@ Feature:  switch_item
       | ON            | OFF         |
       | OFF           | ON          |
 
+    Scenario Outline: Switches accept boolean values
+    Given items:
+      | type   | name       | label       | state           |
+      | Switch | TestSwitch | Test Switch | <initial_state> |
+    And code in a rules file
+      """
+      puts "******* Switch **********"
```
ccutrer, line 30: Debugging?

ccutrer (changes requested):
Code looks good, just has some extra debug logging left in in a test

boc-tothefuture:
format_type was much better

ccutrer (approved):

PR #445, opened 2021-12-01
docs(comparisons): document type comparisons

jimtng:
This adds the documentation for #417. It is far from perfect, so any feedback / improvements will be appreciated.


jimtng:
> I'd also mention "if at any point you want to bypass loose type conversions, use `eql?`. Just be aware that this also bypasses the implicit conversion of an Item to its state."

Thanks for the addition. I have added a subsection for it and provided some examples. Would you mind having a quick look at the examples?

ccutrer (approved):

PR #448, opened 2021-12-02
fix(types): remove java types that shadowed ruby types inside rules

### LINKED ISSUE #348: Exclude certain entities from being pushed into the ruby context by Java library
Continues discussion in #341

jimtng:
Resolve #348

I'm not sure if this is the best way

I'm also not quite sure how to test this other than checking the types manually like seeing if HSBType.new works inside a rule.


boc-tothefuture:
Will this only work inside of rules and not outside of them? 

Does it work if it is called from 'script_loaded' instead?

jimtng:
> Will this only work inside of rules and not outside of them?

It already worked outside of rules before this PR. I could write a test however.

boc-tothefuture:
Would doing it in script_loaded prevent having to do it for every rule?

jimtng:
Cleaning it up inside each execution seem to be necessary.

Is removing the const the right approach though? I would like to know how this problem occurred in the first place, perhaps we could implement a better solution.

jimtng:
For the top level clean up, I cannot place it inside script_loaded because it may be needed as the script loads / executes. I haven't figured out the spot to call the `remove_java_types` method that would work.

Even though the "java" entities are still on the top level, I can still instantiate a PercentType and HSBType, whereas I couldn't do it inside a rule without this patch.

TBH I don't know what I'm doing here :)

ccutrer:
What are the chances we can fix this on the Java side? Basically in the code that adds a $ in front of global variable names, check if it's a Java class and if it is then translate it to the Ruby proxy class before setting it. That would even fix it for gui rules that aren't using our library, and make it a actually live up to the documentation in OpenHAB.

jimtng:
> Basically in the code that adds a $ in front of global variable names

Are we dealing with global variables? I thought the problem is with classes, like PercentType, which doesn't have a $ in front of it.



> check if it's a Java class and if it is then translate it to the Ruby proxy class before setting it.

what is a "ruby proxy class"?

boc-tothefuture:
Does anyone have a clear timeline of want exactly is happening?

Like a 
1.
2.
3.

And which step causes the defect?

jimtng:
The good news is: I can call the cleanup code in script_loaded and not inside each execution. That fixed the problem inside the rules.

Outside the rules are fine, so perhaps this could be the fix, although I'd love to better understand what's actually happening.

jimtng:
I've just tried a plain GUI rule (not using require 'openhab') with the following code
```ruby
Java::OrgSlf4j::LoggerFactory.getLogger("org.openhab.core.automation.examples").error HSBType.new
```

and it produced the same error we're seeing with our openhab-jruby:

```
01:20:02.787 [ERROR] [.internal.handler.ScriptActionHandler] - Script execution of rule with UID 'A_test1' failed: Error during evaluation of Ruby in <script> at line 3: (NoMethodError) undefined method `new' for #<Java::JavaLang::Class: org.openhab.core.library.types.HSBType>
```

So this issue seems to be on the java side indeed.

ccutrer:
It's essentially https://github.com/openhab/openhab-addons/blob/82ac5ee2d90190838a3d5e540834d094e8679173/bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/JRubyScriptEngineFactory.java#L94. OH calls it to import the "presets" (https://github.com/openhab/openhab-core/blob/79edf2b9e643ad802c31b2e5f54edb93534277fd/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptExtensionManager.java#L102), and down the stack it just assigns some variables into the script engine (at https://github.com/openhab/openhab-core/blob/79edf2b9e643ad802c31b2e5f54edb93534277fd/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/AbstractScriptEngineFactory.java#L50 - the JRuby override just alters things a bit). The default preset is imported every time an engine is created (https://github.com/openhab/openhab-core/blob/79edf2b9e643ad802c31b2e5f54edb93534277fd/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java#L134).

> Are we dealing with global variables? I thought the problem is with classes, like PercentType, which doesn't have a $ in front of it.

As shown above OpenHAB is saying "make these globals accessible inside the script". For Ruby, "global" can mean different things. Classes are constants, and are always identified by an initial capital letter. lower case initials are either methods or local variables. So to make incoming "globals" with initial lower case we prepend them with `$` to make ruby global variables (see https://github.com/openhab/openhab-addons/blob/82ac5ee2d90190838a3d5e540834d094e8679173/bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/JRubyScriptEngineFactory.java#L68 - really that method should probably be looking for any initial lower case, rather than special casing a known list).

> what is a "ruby proxy class"?

Java has a `Class`, and Ruby has a `Class`. They are not the same thing. Inside JRuby, you mostly work with Ruby Class, because that's what you can actually create an instance of. But occasionally you still need access to the Java class so that you can call the methods from java.lang.Class. Every Java class will get a "ruby proxy class" created for it that is a Ruby class but allows you to create instances of it. On the Ruby side, you can get the ruby class from a java class by calling `#ruby_class`. And vice versa, you can get the java class from the ruby class by calling `java_class`. Examples (called from the top level of a script, after requiring openhab, so sidestepping the openhab presets problem):

```ruby
PercentType # => Java::OrgOpenhabCoreLibraryTypes::PercentType
PercentType.java_class # => #<Java::JavaLang::Class: org.openhab.core.library.types.PercentType>
PercentType.class # => Class
PercentType.java_class.class # => Java::JavaLang::Class
PercentType.java_class.ruby_class # => Java::OrgOpenhabCoreLibraryTypes::PercentType
```

So, end result, OpenHAB calls scopeValues with a map of things like `{ "PercentType" => org.openhab.core.library.types.PercentType.class }`. On the Java side, that is a Java class, because it's Java. But it gets assigned in to JRuby as the Java class itself. We need to catch that, and somehow get the ruby proxy class from it (probably can't just call `ruby_class` on it, because that method is added on the ruby side), and then assign it in. It's the same idea that when you call Java from Ruby JRuby silently translates between java.lang.String and a Ruby String. But because of how OpenHAB is doing the assigning, it skips JRuby doing any translation like that.

boc-tothefuture:
What makes it work in certain contexts and not others? 

scopeValues is called after the ScriptEngine is created but before the script is evaluated, right?  Once we do a require 'openhab' why does it work outside of the rules but not inside the rule context which is supposed to be evaluating it in the same context as outside the rules?

jimtng:
I tried modifying `scopeValues` and convert java class to "ruby class". The way I do it is by calling `scriptEngine.eval("java_import(org.openhab.core.library.types.HSBType).first")`. This returns the imported java object for the given class. I'm not sure if that's the right way to get the "ruby object". 

It seems to solve the ruby object, however, it's complaining about missing `==` operator in our overridden types (because we removed it in our code).

boc-tothefuture:
The Nashhorn engine is doing something similar where it is transforming anything passed in that is an instance of Class.  We may want to take a similar approach here - is that what you were doing @jimtng ?

jimtng:
> The Nashhorn engine is doing something similar where it is transforming anything passed in that is an instance of Class. We may want to take a similar approach here - is that what you were doing @jimtng ?

I am trying to replace the java objects with a ruby one. I found an even simpler syntax. Instead of using java_import, I could simply eval the java class name, e.g. `scriptEngine.eval("org.openhab.core.library.types.PointType")`. This will return the ruby class.

That part works fine, however, I came across two errors when doing `require 'openhab'`:
```
10:00:25.901 [ERROR] [ript.internal.ScriptEngineManagerImpl] - Error during evaluation of script 'file:/openhab/conf/automation/jsr223/ruby/test2.rb': Error during evaluation of Ruby in org/jruby/RubyModule.java at line 3443: (NameError) method '==' not defined in OpenHAB::DSL::Types::HSBType
```

And inside hsb_type.rb, this line caused an error:
```
        alias raw_hue hue
```
```
10:02:21.947 [ERROR] [ript.internal.ScriptEngineManagerImpl] - Error during evaluation of script 'file:/openhab/conf/automation/jsr223/ruby/test2.rb': Error during evaluation of Ruby in org/jruby/RubyModule.java at line 3369: (NameError) undefined method `hue' for class `OpenHAB::DSL::Types::HSBType'
```

I was able to get around these by:
- Commenting out all the `remove_method :==`  inside our ruby library
- commenting out the `alias raw_hue hue` and `alias argb rgb` inside hsb_type.rb and using `getHue` and `getRgb` where it needed to access the original methods

However, it's a bit strange.

boc-tothefuture:
We need to get the ruby proxies for all things that are of type class, right?  And I think you only want to call eval once for all of them otherwise there will probably be a decent size penalty for each script load?

If you look at nashhorn, they build up a single eval string with all the imports and run that.

I can share what I did so far?

I think that is what we want to do, prepending it with a "require 'java'"?

jimtng:
> I can share what I did so far?
> 
Please do.

boc-tothefuture:
```
    public void scopeValues(ScriptEngine scriptEngine, Map<String, Object> scopeValues) {
        // Empty comments prevent the formatter from breaking up the correct streams chaining
        logger.warn("Scope Values: {}", scopeValues);
        Map<String, Object> filteredScopeValues = //
                scopeValues //
                        .entrySet() //
                        .stream() //
                        .filter(map -> !FILTERED_PRESETS.contains(map.getKey())) //
                        .map(JRubyScriptEngineFactory::mapInstancePresets) //
                        .map(JRubyScriptEngineFactory::mapGlobalPresets) //
                        .collect(Collectors.toMap(map -> map.getKey(), map -> map.getValue())); //

        Map<Boolean, Map<String, Object>> partitionedMap = //
                filteredScopeValues.entrySet() //
                        .stream() //
                        .collect(Collectors.partitioningBy(entry -> (entry.getValue() instanceof Class),
                                Collectors.toMap(Entry::getKey, Entry::getValue)));

        // super.scopeValues(scriptEngine, filteredScopeValues);
        super.scopeValues(scriptEngine, partitionedMap.getOrDefault(false, new HashMap<String, Object>()));
    }
```

I was in the middle of working on this, but I have to step away, but can look at it later depending on how your progress goes.

I have filtered out the class vs. rest using the partitionBy method and right now only the non-class values are going into the env.put.

For the classes, I was going to try this:
iterate over the values of the hashmap from `partitionedMap.getOrDefault(true)` constructing a string that was prepended with "require 'java'" and then a line that was essentially:
"java_import '#{class.getName()}'"
Then join those together with a newline and call  scriptEngine.eval() on it.

I was roughly following the outline here:
https://github.com/openhab/openhab-core/blob/79edf2b9e643ad802c31b2e5f54edb93534277fd/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/NashornScriptEngineFactory.java#L61

jimtng:
That's a better idea.

I was intercepting the put and replacing the Class objects with a ruby version obtained via eval, so it does multiple evals. 

I came across another unrelated problem, have you seen this before?
```
11:01:04.788 [ERROR] [Events.Framework                     ] - FrameworkEvent ERROR
org.osgi.framework.BundleException: Could not resolve module: org.openhab.automation.jrubyscripting [291]
  Unresolved requirement: Import-Package: com.sun.nio.file; resolution:="optional"
  Unresolved requirement: Import-Package: com.sun.security.auth.module; resolution:="optional"
  Unresolved requirement: Import-Package: org.apache.tools.ant; resolution:="optional"
  Unresolved requirement: Import-Package: org.joda.convert; resolution:="optional"
  Unresolved requirement: Import-Package: sun.management; resolution:="optional"
  Unresolved requirement: Import-Package: sun.nio.ch; resolution:="optional"
  Unresolved requirement: Import-Package: jakarta.annotation
```

boc-tothefuture:
No.  What preceded that error?

jimtng:
> No. What preceded that error?

It happened when openhab is trying to load the bundle. 

```
11:57:39.126 [WARN ] [org.apache.felix.fileinstall         ] - Error while starting bundle: file:/openhab/addons/org.openhab.automation.jrubyscripting-3.2.0-SNAPSHOT.jar
org.osgi.framework.BundleException: Could not resolve module: org.openhab.automation.jrubyscripting [291]
  Unresolved requirement: Import-Package: com.sun.nio.file; resolution:="optional"
  Unresolved requirement: Import-Package: com.sun.security.auth.module; resolution:="optional"
  Unresolved requirement: Import-Package: org.apache.tools.ant; resolution:="optional"
  Unresolved requirement: Import-Package: org.joda.convert; resolution:="optional"
  Unresolved requirement: Import-Package: sun.management; resolution:="optional"
  Unresolved requirement: Import-Package: sun.nio.ch; resolution:="optional"
  Unresolved requirement: Import-Package: jakarta.annotation

	at org.eclipse.osgi.container.Module.start(Module.java:463) ~[org.eclipse.osgi-3.16.300.jar:?]
	at org.eclipse.osgi.internal.framework.EquinoxBundle.start(EquinoxBundle.java:445) ~[org.eclipse.osgi-3.16.300.jar:?]
	at org.apache.felix.fileinstall.internal.DirectoryWatcher.startBundle(DirectoryWatcher.java:1260) [bundleFile:3.7.0]
	at org.apache.felix.fileinstall.internal.DirectoryWatcher.startBundles(DirectoryWatcher.java:1233) [bundleFile:3.7.0]
	at org.apache.felix.fileinstall.internal.DirectoryWatcher.startAllBundles(DirectoryWatcher.java:1221) [bundleFile:3.7.0]
	at org.apache.felix.fileinstall.internal.DirectoryWatcher.doProcess(DirectoryWatcher.java:515) [bundleFile:3.7.0]
	at org.apache.felix.fileinstall.internal.DirectoryWatcher.process(DirectoryWatcher.java:365) [bundleFile:3.7.0]
	at org.apache.felix.fileinstall.internal.DirectoryWatcher.run(DirectoryWatcher.java:316) [bundleFile:3.7.0]
```

jimtng:
Purging my maven cache fixed the issue with bundle dependency

jimtng:
I gave this a go, and found that if I imported all of them in one eval, one error or missing class would ruin everything.

jimtng:
Noting this - that the bundle doesn't like being reloaded. It requires an openhab restart
```
Caused by: java.lang.LinkageError: loader constraint violation: when resolving method 'org.jruby.RubyString[] org.jruby.runtime.Helpers.constructRubyStringArray(org.jruby.RubyString, org.jruby.RubyString, org.jruby.RubyString)' the class loader 'bootstrap' of the current class, java/lang/Object, and the class loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @57e2c959 for the method's defining class, org/jruby/runtime/Helpers, have different Class objects for the type org/jruby/RubyString used in the signature (java.lang.Object is in module java.base of loader 'bootstrap'; org.jruby.runtime.Helpers is in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @57e2c959, parent loader java.net.URLClassLoader @57e1b0c)
```

jimtng:
The errors when importing are caused by internal packages from bindings actions. For example:

```
Error importing org.openhab.binding.chromecast.internal.action.ChromecastActions: Error during evaluation of Ruby in org/jruby/javasupport/JavaPackage.java at line 253: (NameError) missing class name org.openhab.binding.chromecast.internal.action.ChromecastActions
```

ccutrer:
You should be able to still do a single eval while doing all of the classes at once by building up a more complicated statement with rescues in it. You can just append " rescue nil" to each line if we want to ignore errors like ChromecastActions.

jimtng:
Import errors aside, it came back to the same problem when `require 'openhab'`

```
14:06:22.704 [ERROR] [ript.internal.ScriptEngineManagerImpl] - Error during evaluation of script 'file:/openhab/conf/automation/jsr223/ruby/test2.rb': Error during evaluation of Ruby in org/jruby/RubyModule.java at line 3443: (NameError) method '==' not defined in OpenHAB::DSL::Types::DateTimeType
```

ccutrer:
I'm only mildly surprised that fixing this will cause compatibility errors in the Ruby gem. I'd be more than happy to look into them and figure out how to make it compatible with both fixed and not-fixed of the Java bundle... as soon as I can find time (I'm between computers at the moment).

jimtng:
> You should be able to still do a single eval while doing all of the classes at once by building up a more complicated statement with rescues in it. You can just append " rescue nil" to each line if we want to ignore errors like ChromecastActions.

Thanks for the tip. That worked.

> I'm only mildly surprised that fixing this will cause compatibility errors in the Ruby gem. I'd be more than happy to look into them and figure out how to make it compatible with both fixed and not-fixed of the Java bundle... as soon as I can find time (I'm between computers at the moment).

You can use this jar to try it:
https://github.com/jimtng/openhab-addons/raw/jruby-scope/bundles/org.openhab.automation.jrubyscripting/org.openhab.automation.jrubyscripting-3.2.0-SNAPSHOT.jar



The code if you'd like to take a look can be found here:
https://github.com/jimtng/openhab-addons/blob/jruby-scope/bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/JRubyScriptEngineFactory.java

boc-tothefuture:
> Noting this - that the bundle doesn't like being reloaded. It requires an openhab restart
> 
> ```
> Caused by: java.lang.LinkageError: loader constraint violation: when resolving method 'org.jruby.RubyString[] org.jruby.runtime.Helpers.constructRubyStringArray(org.jruby.RubyString, org.jruby.RubyString, org.jruby.RubyString)' the class loader 'bootstrap' of the current class, java/lang/Object, and the class loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @57e2c959 for the method's defining class, org/jruby/runtime/Helpers, have different Class objects for the type org/jruby/RubyString used in the signature (java.lang.Object is in module java.base of loader 'bootstrap'; org.jruby.runtime.Helpers is in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @57e2c959, parent loader java.net.URLClassLoader @57e1b0c)
> ```

I think it is a JRuby thing? Or maybe a JRuby/OSGI thing?  I don't believe there is anything we can do about it.  We could open a defect with JRuby.

jimtng:
> I'm only mildly surprised that fixing this will cause compatibility errors in the Ruby gem. I'd be more than happy to look into them and figure out how to make it compatible with both fixed and not-fixed of the Java bundle... as soon as I can find time (I'm between computers at the moment).

@ccutrer have you had a chance to look into this? I've provided the jar and the java source above

ccutrer:
I have not yet. I started a new job this week, and am still getting settled. I haven't forgotten it though. This and semantic item helpers are at the top of my hobby-time list.

boc-tothefuture:
@jimtng can you close this PR and reopen it once I push the cleaned mirror?

Issue #370, opened 2021-11-03
Getting a related item


jimtng:
A very common pattern in my rules is to get another item of a similar name, e.g.

```ruby
triggered do |item|
  event_data = items[item.name.sub '_LastMotionType', '_EventData']`
```

Is there a way (or should we even) make this process somewhat easier? I understand that this would rely on some sort of naming pattern.

jimtng:
I'm thinking something like this but not too sure about the name

```ruby
event_data = item.sibling('_EventData')
```

ccutrer:
I think it's something to explore.  Ideally I think I'd want to be able to do something like `GreatCans_Scene.dimmer_item` (get the dimmer associated with the scene controller on the same switch), or `GuestBathShade_Rollershutter.window_contact`, but there isn't always a 1:1 mapping like this (window_contact might be a door_contact, or sometimes it might not have one, or some windows there are multiple shades that I also have to remove Blackout or Roman from the shade name to get to the window. It seems the only way to get what I really want would be to let you declare some sort of mappings (as a hash), then you could use a default proc to do more custom stuff. But is all that actually going to be much less code than a couple of sub calls? Maybe.

boc-tothefuture:
I agree that an "association" system is needed but I am not sure what that looks like. 
I have been doing mine with hashes because:
1. I don't like name based associations because they are fragile 
2. More importantly, if they are done inside of a rule they break during rule execution not during rule processing.

pacive:
If I would implement something like this I would use Item metadata. Then the code would look like:
```ruby
event_data = items[item.meta[eventDataItem]]
```
IMHO this offers much more flexibility than using the item names to find related items.

jimtng:
I'm closing this because there doesn't seem to be an elegant way of doing it, and it's not a critically needed feature, since it can be achieved in many other ways.

ccutrer:
so with my latest two PRs (plus @ktgeek's), it doesn't completely address it, but there _is_ the possibility of using the things to get related items.

I.e. I have a bunch of zwave dimmers that I often get from the scene channel triggering a rule to get to the main dimmer channel. This could be rewritten now as:

```
changed gSceneChannels.members, to: 1.3 do |event|
  dimmer_item = event.item.thing.channels["switch_dimmer"].item
  dimmer_item.ensure << 100
end
```

And the entire `thing.channels["switch_dimmer"].item` could be a great candidate to refactor into:

```
class GenericItem
[...]

PR #449, opened 2021-12-02
feat(semantics): add semantic helper methods to items

ccutrer:
refs #370

both direct wrapping of the action methods, as well as additional
convenience methods

### DIFF
--- a/lib/openhab/dsl/items/semantics.rb
+++ b/lib/openhab/dsl/items/semantics.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require_relative 'enumerable'
+require_relative 'semantics/enumerable'
 
 module OpenHAB
   module DSL
@@ -86,7 +86,7 @@ module OpenHAB
       #
       # @return [Class]
       def location_type
-        SemanticsAction.get_location_type(Self)&.ruby_class
+        SemanticsAction.get_location_type(self)&.ruby_class
       end
 
       # Gets the related Equipment Item of this Item.
@@ -168,8 +168,6 @@ module OpenHAB
         members.points(*point_or_property_types)
       end
     end
-
-    GenericItem.include(Semantics)
   end
 end
 
@@ -204,7 +202,7 @@ module Enumerable
   #   lGreatRoom.equipments.flat_map(&:members).points(Semantics::Switch)
   def points(*point_or_property_types) # rubocop:disable Metrics
     unless (0..2).cover?(point_or_property_types.length)
-      raise ArgumentError, "wrong number of arguments (given #{point_or_property_types.length}, expected 1..2)"
+      raise ArgumentError, "wrong number of arguments (given #{point_or_property_types.length}, expected 0..2)"
     end
     unless point_or_property_types.all? do |tag|
              tag < OpenHAB::DSL::Semantics::Point || tag < OpenHAB::DSL::Semantics::Property
@@ -220,8 +218,8 @@ module Enumerable
       next unless point.point?
 
       point_or_property_types.all? do |tag|
-        (tag < OpenHAB::DSL::Semantics::Point && point.point_type <= tag) ||
-          (tag < OpenHAB::DSL::Semantics::Property && point.property_type <= tag)
+        (tag < OpenHAB::DSL::Semantics::Point && point.point_type&.<=(tag)) ||
+          (tag < OpenHAB::DSL::Semantics::Property && point.property_type&.<=(tag))
       end
     end
   end
--- a/lib/openhab/dsl/items/semantics.rb
+++ b/lib/openhab/dsl/items/semantics.rb
@@ -1,0 +1,228 @@
+# frozen_string_literal: true
+
+require_relative 'enumerable'
+
+module OpenHAB
+  module DSL
+    # Module for implementing semantics helper methods on [GenericItem]
+    #
+    # Wraps https://www.openhab.org/javadoc/latest/org/openhab/core/model/script/actions/semantics,
+    # as well as adding a few additional convenience methods.
+    # Also includes Classes for each semantic tag.
+    #
+    # Be warned that the Semantic model is stricter than can actually be
+    # described by tags and groups on an Item. It makes assumptions that any
+    # given item only belongs to one semantic type (Location, Equipment, Point).
+    #
+    # See https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.semantics/model/SemanticTags.csv
+    module Semantics
+      # @!visibility private
+      # import the actual semantics action
+      SemanticsAction = org.openhab.core.model.script.actions.Semantics
+      private_constant :SemanticsAction
+
+      # import all the semantics constants
+      [org.openhab.core.semantics.model.point.Points,
+       org.openhab.core.semantics.model.property.Properties,
+       org.openhab.core.semantics.model.equipment.Equipments,
+       org.openhab.core.semantics.model.location.Locations].each do |parent_tag|
+        parent_tag.stream.for_each do |tag|
+          const_set(tag.simple_name.to_sym, tag.ruby_class)
+        end
+      end
+
+      # put ourself into the global namespace, replacing the action
+      ::Semantics = self # rubocop:disable Naming/ConstantName
+
+      # Checks if this Item is a Location
+      #
+      # This is implemented as checking if the item's semantic_type
+      # is a Location. I.e. an Item has a single semantic_type.
+      #
+      # @return [true, false]
+      def location?
+        SemanticsAction.location?(self)
+      end
+
+      # Checks if this Item is an Equipment
+      #
+      # This is implemented as checking if the item's semantic_type
+      # is an Equipment. I.e. an Item has a single semantic_type.
+      #
+      # @return [true, false]
+      def equipment?
+        SemanticsAction.equipment?(self)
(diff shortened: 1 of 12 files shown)


ccutrer:
@jimtng @boc-tothefuture : I still need to write cucumber tests, but things seem to be working with what I've got in my live OpenHAB. Just looking for early feedback if you like the direction I'm going here.

boc-tothefuture:
I haven't touched the semantic model.. are you setting that in the GUI or with files somehow?

ccutrer:
You can do either. The docs steer you towards UI (https://www.openhab.org/docs/tutorial/model.html), but you can also do it from items files. It's just a specific set of tags, with a few constraints. https://community.openhab.org/t/oh3-semantic-model-setup-via-tags-in-configuration-items-files/112520

I started using the latter, because I keep all my configuration in text files both so I can auto-generate it when possible, and to have it in source control. Way too many clicks to do anything in the UI for as many things as I have.

lib/openhab/dsl/items/semantics.rb
```diff
@@ -0,0 +1,179 @@
...
+      # Gets the related Location Item of this Item.
+      # @return [GenericItem]
+      def location
+        SemanticsAction.get_location(self)
+      end
+
+      # Returns the sub-class of [Location] this Item is tagged with.
+      # @return [Class]
+      def location_type
+        SemanticsAction.get_location_type(Self)&.ruby_class
```
jimtng, line 83: Is `Self` not a typo?

jimtng (commented):
Looks great! I like the #point(s)_for methods.

lib/openhab/dsl/items/semantics.rb
```diff
@@ -0,0 +1,179 @@
...
+      # Returns the sub-class of [Point] this Item is tagged with.
+      # @return [Class]
+      def point_type
+        SemanticsAction.get_point_type(self)&.ruby_class
+      end
+
+      # Gets the [Point] Items belonging to this Equipment Item.
+      # @return [Array<GenericItem>]
+      def points
+        return [] unless equipment? && is_a?(GroupItem)
```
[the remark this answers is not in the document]
ccutrer: Hmm, I didn't think Points were allowed to belong directly to a Location. But I see I was wrong. So yeah, I'll need to rework this a bit.

As for "letting the underlying openhab code", there _isn't_ a method to on the Semantics script action to get all the points that belong to an Equipment or Location.

lib/openhab/dsl/items/semantics.rb
```diff
@@ -0,0 +1,179 @@
...
+      # to (or none, if it doesn't belong to any).
+      #
+      # Note that an Equipment Item will return itself.
+      #
+      # @return [GenericItem, Array<GenericItem>, nil]
+      def equipment
+        return SemanticsAction.get_equipment(self) unless location? && is_a?(GroupItem)
+
+        members.select(&:equipment?)
+      end
```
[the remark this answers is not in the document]
ccutrer: Doh, I didn't realize you could have sub-equipment. So yeah, we should use `equipments` and `equipment`. It's not the best grammar, but is also how OpenHAB names their method for getting the list of tags that can be equipment.

lib/openhab/dsl/items/semantics.rb
ccutrer: Yes, indeed it is. This is why I need tests :)

jimtng: I'm curious, mainly because I am still not familiar with `ruby_class`. Some of the calls in other methods didn't convert the returned object into ruby_class like this one, and I'm wondering what the difference.

lib/openhab/dsl/items/semantics.rb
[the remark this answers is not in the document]
jimtng: I am actually hoping that they would allow an equipment item to also be a point and have property, as in assign an equipment, a point and a property into the one item. It seems redundant to have to create a Group to become the equipment group for one point. I haven't voiced the request yet, since I've only just started learning about the semantic model.

lib/openhab/dsl/items/semantics.rb
```diff
@@ -0,0 +1,179 @@
...
+      # Returns the sub-class of [Equipment] this Item is tagged with.
+      # @return [Class]
+      def equipment_type
+        SemanticsAction.get_equipment_type(self)&.ruby_class
+      end
+
+      # Gets the related Location Item of this Item.
+      # @return [GenericItem]
+      def location
+        SemanticsAction.get_location(self)
```
jimtng, line 77: I _think_ a point (and probably also an equipment) _can_ belong to multiple locations. I'm not sure if it's a bug but openhab allows it and it shows up in the mainui -> settings -> model tree.

lib/openhab/dsl/items/semantics.rb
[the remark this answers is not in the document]
ccutrer: It seems to work for some things, but not a others. It's really quite annoying. Cause yeah having to introduce an intermediate group for the equipment a huge pain when you have hundreds of single-point-equipments. And by "work" I mean "shows up and is usable in the MainUI on any of the locations, equipment, or points tab."

jimtng:
I've been thinking about how to utilise this semantic model to solve #370 
However, it seems that the current implementation of semantic model is lacking, especially the selection of valid properties. 

What do you think if we were to allow custom tags in the search done within `points_for`? This will make it infinitely more usable. As it is now I'm a bit stuck trying to figure out how to use the semantic model with the limited number of choice for Point/Property.

Case in point:
- I have a smart light bulb that has Power, Brightness, Colortemp, Colorhsb
- I have a tasmota button that I'd like to get under the same "light" equipment so I can control the light by getting ButtonItem.point_for(Light, Power), point_for(Light, Brightness), Colortemp and Color.

But, notice there isn't a property for Brightness, nor Color. The only applicable properties are `Light` (which I suppose I could use for the on/off) and `ColorTemperature`.

I figured I could still use `point_for` if I supplanted the system with custom tags, especially for property.

This means we'd need to relax the requirements for points_for to only require the argument to contain one Point but the other one can be a simple tag (not necessarily a property).

ccutrer:
I agree on some points. But I think I want to go Rails-like, allowing easily composable query methods.  Basically like this:

`item.points_for(Semantics::Control).grep(ColorItem).tagged("MyCustomTag").first` (`grep` is already part of Enumerable, we would just need to add a `tagged` method, and decide if it should live on Enumerable, Array, or force a custom array descendant to wrap the return type of any method we want to be able to start such a chain.)

This means you don't have to maintain a single method that knows how to handle every possible way of traversing item structures.  `points_for` only deals with the semantic model.

Other "query" methods I can think of are `in_group` (basically a chainable synonym for `&`), and something like `channel_items`.

jimtng:
> I agree on some points. But I think I want to go Rails-like, allowing easily composable query methods. Basically like this:
> 
> `item.points_for(Semantics::Control).grep(ColorItem).tagged("MyCustomTag").first` (`grep` is already part of Enumerable, we would just need to add a `tagged` method, and decide if it should live on Enumerable, Array, or force a custom array descendant to wrap the return type of any method we want to be able to start such a chain.)
> 
> This means you don't have to maintain a single method that knows how to handle every possible way of traversing item structures. `points_for` only deals with the semantic model.
> 
> Other "query" methods I can think of are `in_group` (basically a chainable synonym for `&`), and something like `channel_items`.

Brilliant!

boc-tothefuture:
I think we will to close this PR for me to push the updates removing the large binaries.  

@ccutrer can you close and reopen once I push the cleaned mirror?

jimtng:
@ccutrer are you still working on this?

jimtng:
> we would just need to add a `tagged` method, and decide if it should live on Enumerable, Array, or force a custom array descendant to wrap the return type of any method we want to be able to start such a chain.)

By having it inside Enumerable, won't that affect all instances of Enumerable? If that's the case, I would prefer to have a custom array descendant.

lib/openhab/dsl/items/semantics.rb
```diff
@@ -0,0 +1,228 @@
...
+       point_or_property_types.count { |tag| tag < OpenHAB::DSL::Semantics::Property } > 1
+      raise ArgumentError, 'point_or_property_types cannot both be a subclass of Point or Property'
+    end
+
+    select do |point|
+      next unless point.point?
+
+      point_or_property_types.all? do |tag|
+        (tag < OpenHAB::DSL::Semantics::Point && point.point_type <= tag) ||
+          (tag < OpenHAB::DSL::Semantics::Property && point.property_type <= tag)
```
jimtng, line 224: I came across a situation where point.property_type is nil, so this line raised an error `semantics.rb at line 227: (NoMethodError) undefined method `<=' for nil:NilClass`

```suggestion
          (tag < OpenHAB::DSL::Semantics::Property && point.property_type&.<=(tag))
```

Perhaps this would also apply to the previous line in the case where point_type could return nil.

lib/openhab/dsl/items/semantics.rb
jimtng: When an item belongs to more than one locations, openhab picks only one.

jimtng:
@boc-tothefuture this PR is ready to be merged now. It's a fantastic addition to the library, and I have started using it. I believe this resolves #370

PR #450, opened 2021-12-03
feat(datetime): support ZonedDateTime objects as commands for DateTime items

### LINKED ISSUE #222: DateTime Item doesn't accept ZonedDateTime
This code failed:

```ruby
DateTimeItem1 << ZonedDateTime.now
```

```
13:02:38.797 [WARN ] [ab.core.model.script.actions.BusEvent] - Cannot convert '2021-03-05T13:02:38.788380+10:00[Australia/Brisbane]' to a state type which item 'DateTimeItem1' accepts: [DateTimeType, UnDefType].
```

I understand that this is actually an openhab-core problem. Do we want to intercept and correct it?

The same error occurred with #update too

boc-tothefuture:
closes #222

### DIFF
--- a/lib/openhab/dsl/items/date_time_item.rb
+++ b/lib/openhab/dsl/items/date_time_item.rb
@@ -2,6 +2,7 @@
 
 require 'forwardable'
 require 'time'
+require 'java'
 
 require 'openhab/dsl/items/comparable_item'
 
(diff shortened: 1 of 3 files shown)


lib/openhab/dsl/items/date_time_item.rb
```diff
@@ -53,6 +55,7 @@ def coerce(other)
         # @!visibility private
         def format_type(command)
           return command.iso8601 if command.respond_to?(:iso8601)
+          return DateTimeFormatter::ISO_OFFSET_DATE_TIME.format(command) if command.is_a? Java::JavaTime::ZonedDateTime
```
ccutrer, line 58: I feel like we've standardized on java.time.ZonedDateTime format for referencing java classes where possible.

ccutrer (approved):

PR #453, opened 2021-12-04
feat(hsb): add from/to hash support

### LINKED ISSUE #305: Support for conversion to/from hashes for Color Items
As documented in #34 and #295

boc-tothefuture:
closed #305

### DIFF
--- a/lib/openhab/dsl/items/color_item.rb
+++ b/lib/openhab/dsl/items/color_item.rb
@@ -48,9 +48,50 @@ module OpenHAB
         # convert it to an HSBType
         # @!visibility private
         def format_type(command)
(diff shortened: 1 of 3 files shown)


lib/openhab/dsl/items/color_item.rb
```diff
@@ -48,10 +48,51 @@ def coerce(other)
         # convert it to an HSBType
         # @!visibility private
         def format_type(command)
+          return format_hash(command.to_h) if command.respond_to?(:to_h)
```
ccutrer, line 51: I'd recommend using to_hash here, not to_h. It will avoid round-tripping to a hash-and-back if you pass in an HSBType (which now responds to to_h).

ccutrer (changes requested):

PR #454, opened 2021-12-04
feat(location): support hashes for locations

### LINKED ISSUE #311: add hash to/from methods for Location
See #230

boc-tothefuture:
closes #311

### DIFF
--- a/lib/openhab/dsl/types/point_type.rb
+++ b/lib/openhab/dsl/types/point_type.rb
@@ -17,6 +17,7 @@ module OpenHAB
         # @param longitude [DecimalType, QuantityType, StringType, Numeric]
         # @param altitude [DecimalType, QuantityType, StringType, Numeric]
         def initialize(*args) # rubocop:disable Metrics
+          args = from_hash(args.first.to_hash) if args.first.respond_to? :to_hash
           if (2..3).cover?(args.length)
             args = args.each_with_index.map do |value, index|
               if value.is_a?(DecimalType) || value.is_a?(StringType)
@@ -119,6 +120,13 @@ module OpenHAB
         end
 
         #
+        # Convert the PointType to a hash
+        # @return [Hash] with keys latitude/longitude/altitude
+        def to_h
+          { latitude: latitude, longitude: longitude, altitude: altitude }
+        end
+
+        #
         # Calculate the distance in meters from other, ignoring altitude.
         #
         # This algorithm also ignores the oblate spheroid shape of Earth and
@@ -138,7 +146,7 @@ module OpenHAB
 
         # coerce an object to a PointType
         # @return [PointType]
-        def coerce_single(other)
+        def coerce_single(other) # rubocop:disable Metrics
           logger.trace("Coercing #{self} as a request from #{other.class}")
           if other.is_a?(PointType)
             other
@@ -148,7 +156,22 @@ module OpenHAB
             other.state
           elsif other.respond_to?(:to_str)
             PointType.new(other.to_str)
+          elsif other.respond_to?(:to_hash)
+            PointType.new(other.to_hash)
           end
+        end
+
+        #
+        # Convert hash into ordered arguments for constructor
+        #
+        def from_hash(hash)
+          keys = [%i[lat long alt], %i[latitude longitude altitude]]
+          keys.each do |key_set|
+            values = hash.transform_keys(&:to_sym).values_at(*key_set)
+
+            return *values.compact if values[0..1].all?
+          end
+          raise ArgumentError, "Supplied arguments (#{hash}) must contain one of the following sets #{keys}"
         end
       end
     end
--- a/docs/usage/items/location.md
+++ b/docs/usage/items/location.md
@@ -21,7 +21,12 @@ Represents GPS coordinates
 Update the location Item
 
 ```ruby
-Location << '30,20'
+Location << '30,20'   # latitude of 30, longitude of 20
+Location << {lat: 30, long: 30}
+Location << {lat: 30, long: 30, alt: 80}
+Location << '30,20,80' # latitude of 30, longitude of 20, altitude of 80
+Location << {latitude: 30, longitude: 30}
+Location << {latitude: 30, longitude: 30, altitude: 80}
 ```
 
 or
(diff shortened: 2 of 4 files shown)


jimtng:
does it support
```
Location << '1,2,3' # for lat, long, alt
Location << 1,2,3
```

boc-tothefuture:
It does since the hash stuff just converts to an array.  But I can update docs for the array based

lib/openhab/dsl/types/point_type.rb
```diff
@@ -16,7 +16,14 @@ class PointType
...
         # @param longitude [DecimalType, QuantityType, StringType, Numeric]
         # @param altitude [DecimalType, QuantityType, StringType, Numeric]
-        def initialize(*args) # rubocop:disable Metrics
+        # @param :latitude [DecimalType, QuantityType, StringType, Numeric]
+        # @param :lat [DecimalType, QuantityType, StringType, Numeric]
+        # @param :longitude [DecimalType, QuantityType, StringType, Numeric]
+        # @param :long [DecimalType, QuantityType, StringType, Numeric]
+        # @param :altitude [DecimalType, QuantityType, StringType, Numeric]
+        # @param :alt [DecimalType, QuantityType, StringType, Numeric]
+        def initialize(*args, **kargs) # rubocop:disable Metrics
```
ccutrer, line 25: Normally you use "kwargs", not "kargs".  Though I'm also wondering if you should name them or not (not being using a splat as you have). Or even if we should use kwargs or only accept an _actual_ hash (which should functionally be the same from the caller's point of view, though there are land mines in Ruby 2.6, 2.7, and 3.0 with hash-as-last-arg-in-combination-with-splats-and-kwarg-splats). I think I'm honestly leaning that way, given how we're expecting it to be an atomic hash with _either_ one set of opts _or_ a different set of opts, instead of being several optional modifying params, each of them optional in their own right.

lib/openhab/dsl/types/point_type.rb
```diff
@@ -148,7 +162,21 @@ def coerce_single(other)
...
+          elsif other.respond_to?(:to_hash)
+            PointType.new(other.to_hash)
+          end
+        end
+
+        #
+        # Convert possible keyword arguments into ordered arguments for constructor
+        #
+        def from_kargs(kargs)
+          keys = [%i[lat long alt], %i[latitude longitude altitude]]
```
ccutrer, line 174: it's possible to create one without altitude, right? Just lat and long.

lib/openhab/dsl/types/point_type.rb
```diff
@@ -148,7 +162,21 @@ def coerce_single(other)
...
+          keys.each do |key_set|
+            values = kargs.transform_keys(&:to_sym).values_at(*key_set).compact
+            return *values if values.length >= 2
```
ccutrer, line 177: Oh yeah, that's what you're trying to do here. But it's not going to work right, cause if you pass say latitude and _altitude_, you'll have two values, but altitude will be interpreted as longitude.

ccutrer (changes requested):

lib/openhab/dsl/types/point_type.rb
```diff
@@ -16,7 +16,14 @@ class PointType
...
         # @param altitude [DecimalType, QuantityType, StringType, Numeric]
-        def initialize(*args) # rubocop:disable Metrics
+        # @param :latitude [DecimalType, QuantityType, StringType, Numeric]
+        # @param :lat [DecimalType, QuantityType, StringType, Numeric]
+        # @param :longitude [DecimalType, QuantityType, StringType, Numeric]
+        # @param :long [DecimalType, QuantityType, StringType, Numeric]
+        # @param :altitude [DecimalType, QuantityType, StringType, Numeric]
+        # @param :alt [DecimalType, QuantityType, StringType, Numeric]
+        def initialize(*args, **kargs) # rubocop:disable Metrics
+          args = from_kargs(kargs) if kargs.length.positive?
```
ccutrer, line 26: unless kwargs.empty?

ccutrer (changes requested):

ccutrer:
> It does since the hash stuff just converts to an array.  But I can update docs for the array based

I'm not sure that's right. Yes, the constructor can take those, but format_type and coerce are only looking if the object responds to to_hash (which an Array does not)

boc-tothefuture:
> > It does since the hash stuff just converts to an array.  But I can update docs for the array based
> 
> I'm not sure that's right. Yes, the constructor can take those, but format_type and coerce are only looking if the object responds to to_hash (which an Array does not)

Misspoke, by "array" I mean the CSV string that @jimtng was referring to in his question.  I updated the docs to show that you can send altitude in the CSV string.

boc-tothefuture:
> does it support
> 
> ```
> Location << '1,2,3' # for lat, long, alt
> Location << 1,2,3
> ```
> 
> If it does, it should be mentioned in the docs / examples

It supports the first example you give, not the second.  I don't think the second example '1,2,3' compiles as valid ruby, you would have to wrap it in brackets.

ccutrer:
> > does it support
> 
> > 
> 
> > ```
> 
> > Location << '1,2,3' # for lat, long, alt
> 
> > Location << 1,2,3
> 
> > ```
> 
> > 
> 
> > If it does, it should be mentioned in the docs / examples
> 
> 
> 
> It supports the first example you give, not the second.  I don't think the second example '1,2,3' compiles as valid ruby, you would have to wrap it in brackets. 
> 
> 
Yeah I was assuming that was a typo. `Location << [1, 2, 3]` seems like reasonable syntax to me.

ccutrer (approved):

jimtng:
Pardon my ignorance... is this different to <<, because it works:
```
def a(*b)
  logger.error b
end

a 1, 2, 3
```

it shows b as an array [1,2,3]

ccutrer:
It's different because you're not passing an array, you're passing 3 arguments that are converted to an array by the splat. The shovel operator (as with Boolean operators in Ruby) only takes a single right hand argument unless you use parentheses, so even taking a splat to shovel would not work. Even if you could I'd worry about operator precedence (shovel is surprisingly low on that list) vs. comma, if it was ever wrapped in another method call.

jimtng:
> It's different because you're not passing an array, you're passing 3 arguments that are converted to an array by the splat. The shovel operator (as with Boolean operators in Ruby) only takes a single right hand argument unless you use parentheses, so even taking a splat to shovel would not work. Even if you could I'd worry about operator precedence (shovel is surprisingly low on that list) vs. comma, if it was ever wrapped in another method call.

Thanks for the explanation. It makes sense now, especially with operator precedence. It had got me many times.

PR #455, opened 2021-12-05
fix(monthday): fix equality comparison against string and allow parsing single digit month-days

### LINKED ISSUE #377: MonthDay comparisons
Thanks for adding MonthDay! 

I'd like to suggest some additional features:

```ruby
MonthDay.now == '11-05'
MonthDay.now > '11-05'
MonthDay.now.between? '11-01'..'11-05' # this is similar to TimeOfDay.now.between?
```

jimtng:
Fix #377 - MonthDay.of(10, 10) == '10-10' wasn't working before.

Allows MonthDay.parse('2-3') format, which is the same as MonthDay.parse('02-03') for the lazy people like me. Also allows leading `--` if provided.

Added docs.

### DIFF
--- a/docs/usage/misc/month_day.md
+++ b/docs/usage/misc/month_day.md
@@ -1,0 +1,55 @@
+---
+layout: default
+title: MonthDay
+nav_order: 6
+has_children: false
+parent: Misc
+grand_parent: Usage
+---
+
+# MonthDay
+
+MonthDay class from [java.time.MonthDay](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/MonthDay.html) can be used in rules for month-date related logic. Notable Methods:
+
+| Method       | Parameter  | Description                                                                                                                                                                  |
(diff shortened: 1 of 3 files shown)


boc-tothefuture:
LGTM - 

From a consistency standpoint (and doesn't have to be in this PR - we could open another issue if we agree) should it support keyword constructors like TimeOfDay.  So MonthDay.new(m: 8, d: 9)?  

Unrelated to this - I also wonder if we should update TimeOfDay to be more like MonthDay which is a thinner veneer around LocalTime.

jimtng:
> From a consistency standpoint (and doesn't have to be in this PR - we could open another issue if we agree) should it support keyword constructors like TimeOfDay. So MonthDay.new(m: 8, d: 9)?

I was wondering about this too. I think we should. 

> 
> Unrelated to this - I also wonder if we should update TimeOfDay to be more like MonthDay which is a thinner veneer around LocalTime.

Yes, but I was looking for, and couldn't find, that issue you created about including Rails' Time stuff because that would be exciting.

PR #469, opened 2021-12-27
fix(item_groups): fix Item#groups returning nil sometimes

ccutrer:
OpenHAB allows you to reference a group item that doesn't exist
(so that it doesn't matter which file your define the group item
in), but we shouldn't be returning nils from `groups`.


boc-tothefuture:
@ccutrer can you rebase to get the test fixes I pushed up?

PR #470, opened 2021-12-27
fix(items): don't exclude GroupItems from `items` global hash

ccutrer:
OpenHAB's DSL rules doesn't, and it can be surprising and annoying when you're
trying to find an item by using Enumerable methods (i.e. `select`)
against items and your item doesn't show up just because it happens
to be a GroupItem. especially since hash lookup (`items["GroupItem"]`)
doesn't have the same restriction.

### DIFF
--- a/features/groups.feature
+++ b/features/groups.feature
@@ -229,4 +229,4 @@ Feature:  groups
       logger.info("Item+Group Count: #{(items + groups).count}")
       """
     When I deploy the rules file
-    Then It should log 'Item+Group Count: 8' within 5 seconds
+    Then It should log 'Item+Group Count: 13' within 5 seconds
--- a/lib/openhab/dsl/items/item_registry.rb
+++ b/lib/openhab/dsl/items/item_registry.rb
@@ -35,7 +35,7 @@ module OpenHAB
         # Explicit conversion to array
         # @return [Array]
         def to_a
-          $ir.items.grep_v(org.openhab.core.items.GroupItem) # rubocop:disable Style/GlobalVars
+          $ir.items.to_a # rubocop:disable Style/GlobalVars
         end
       end
 


boc-tothefuture:
Willing to have a discussion on this, although the design was intentional.  While openhab DSL doesn't treat groups distinctly from items, I think that is a mistake as in most cases they are treated and operated on distinctly.  Groups have members, etc.  I always know when I am trying to get a group and I am never thinking, "Oh I want to get an item which just happens to be a group". 

Can you provide a use case where it makes sense to mix items and groups together?  Also is putting groups back in with items better than `(items+groups).select`?

ccutrer:
any time you're looking for things tagged a certain way, where you don't care if the implementation is a single item, or a group of time. this happens all. the. time.

examples:
 * I have many rooms with one window/one shade. I also have many rooms with multiple windows/shades. I have automations around both the states of the windows, and to control the shades in those rooms. In the room with multiples, I have the individual items, and I also have a group item that contains all the individuals for that room. given a rule that is supposed to control the shades for the room, I can just do `items.select { |i| i.tags.include?("Shade") }` and I'll get the single item that represents all the shades for the room, and it might be a GroupItem or a RollershutterItem, I care not.
 * I have a few virtual 3-ways for lights. Basically, it's two (or more) smart switches, where one (or more!) control the actual lights, but if you change 1, the others immediately follow. I create a GroupItem that aggregates them together. But 99% of my switches are single items, so I'm not going to create a GroupItem for every single one of them just so my rules can always rely on it being a group.
 * Similar situation, I have an alarm contact on every door. But I have one door that's a double-slider. It has _two_ contacts, because that was the only way to physically mount (wired) contacts. But I never care which side is open or closed - I just present in my UI a single unified `Group:Contact:AND(CLOSED,OPEN)` item. Again, my rules don't need to care whether they're working with a ContactItem or a GroupItem. They just get the items that are members of gDoor.
 *  When you look at `GroupItem#members` there's no partitioning of members that are groups vs. members that aren't. It's _super_ jarring when the top-level `items` that's in many ways the same type of object behaves differently.
 * If we're having a `groups` top-level, we might as well have `dimmers` and `rollershutters` and `switches`... Okay yeah, that's ridiculous. Not once have I ever thought to use `groups`. Sometimes I use `items.grep(DimmerItem)` or the like - which seems just as obvious to me to use `items.grep(GroupItem)` instead of `groups`.
 * again, re-iterating that literally everything else in OpenHAB, from rules DSL to documentation to configuring things in the UI, a GroupItem is just a particular type of GenericItem. it doesn't have a separate inheritance tree. nowhere else does it filter out GroupItems (though there are a few places that only show groups). so surprising to have a central, top-level object do this

boc-tothefuture:
@jimtng your thoughts?

jimtng:
I'm in favour of having groups in the items, just can't explain it as well as ccutrer did. It would be a surprise for new users to find that groups isn't included, because as has been said, groupitem is also an item.

boc-tothefuture:
Two against one, so majority rules here.  

The tagging example was a good one.

boc-tothefuture:
@ccutrer could you make the a feat and a mark it breaking - then we can merge it.

boc-tothefuture:
@ccutrer can you rebase and mark it breaking?

jimtng:
@boc-tothefuture the commit message has been updated to mark it as breaking

boc-tothefuture:
I see that, commit lint is failing.

jimtng:
Oh, I see, the `!` should be after the parentheses, i.e. 

```
fix(items)!: don't exclude GroupItems from items global hash 

OpenHAB doesn't, and it can be surprising and annoying when you're
trying to find an item by using Enumerable methods (i.e. `select`)
against items and your item doesn't show up just because it happens
to be a GroupItem. especially since hash lookup (`items["GroupItem"]`)
doesn't have the same restriction.
`

Issue #466, opened 2021-12-25
Item points to a stale instance


jimtng:
I came across a weird problem.

I have a rule file that thinks the status of an item was OFF even when it changed to ON.
The only way to make the rule "see" the actual item status was to reload the rule file.

What could be the cause of this problem?

jimtng:
I should add, that I have this as a constant outside of the rule block:

```ruby
SHOWER_LIGHT = { MasterBathRoom_Humidity => Shower_Light, MainBathRoom_Humidity => MainBathRoom_Light_Power }
```

Both `Shower_Light` and `MainBathRoom_Light_Power` didn't return the correct item status inside the rule body. They were referred to as:

```
rule 'x' do
  updated SHOWER_LIGHT.keys
  triggered do |item|
    light = SHOWER_LIGHT[item]
    logger.info("#{light.name}: #{light}")
  end
end
```

I suspect this has something to do with it. However, I've only started noticing this problem since 3.2. The same rule has been running fine in the past.

jimtng:
I believe the cause is this: after the rule was loaded, I might have made some changes to the items file definition that might have caused the items to get 'refreshed' within openhab, which might have resulted in a new instance of the item, meanwhile the rule is still referring to the item's old instance (because the constant is not being re-evaluated).

I guess the moral of the story is only use this form of item constant when the items aren't being tinkered with. When the items definitions are constantly being updated, it's best to dynamically resolve the item inside rule's invocations.

However it makes me wonder about rules' triggers and whether they too would be referring to old items instances and could stop triggering.

jimtng:
> However it makes me wonder about rules' triggers and whether they too would be referring to old items instances and could stop triggering.

I've just realised the answer to this: rules triggers are based on item names as string, so it would continue to work even when the item objects changed.

boc-tothefuture:
Is it the lookup that is stale or the trigger?

jimtng:
> Is it the lookup that is stale or the trigger?

The trigger is not stale.

It isn't the lookup that's stale either. It's the fact that the lookup was done at script loading time is what makes it stale.

boc-tothefuture:
Hmm.  That could be very confusing for users.  I wonder if we should change the implementation so that we aren't returning the OpenHab item for those lookups bit a proxy/delegate object that fetches right before any operations are performed on it.

ccutrer:
[...]

PR #474, opened 2021-12-29
feat(item): support reloading of references items

boc-tothefuture:
Implementation for discussion around #466

### DIFF
--- a/lib/openhab/core/item_proxy.rb
+++ b/lib/openhab/core/item_proxy.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+require 'delegate'
+require 'forwardable'
+
+module OpenHAB
+  module Core
+    # Class is a proxy to underlying item
+    class ItemProxy < Delegator
+      extend Forwardable
+      def_delegators :__getobj__, :class, :is_a?, :kind_of?, :instance_of?
+
+      #
+      # Set the proxy item (called by super)
+      #
+      def __setobj__(item)
+        # Convert name to java version for faster lookups
+        @item_name = item.name.to_java
+      end
+
+      #
+      # Lookup item from item registry
+      #
+      def __getobj__
+        $ir.get(@item_name) # rubocop:disable Style/GlobalVars
+      end
+    end
+  end
+end
--- a/features/comparisons.feature
+++ b/features/comparisons.feature
@@ -583,5 +583,4 @@ Feature: comparisons
       logger.info("Total tests: #{tests.count}, passed: #{ok_count}, failed: #{tests.count - ok_count}")
       """
     When I deploy the rules file
-    Then It should log "All tests passed" within 2 seconds
-    And It should not log "Test ERROR" within 2 seconds
+    Then It should log "All tests passed" within 10 seconds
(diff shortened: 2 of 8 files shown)


jimtng:
What a brilliant solution!

boc-tothefuture:
> What a brilliant solution!

Well, accept that it doesn't fully work :( 

```
            case other
            when Types::QuantityType,
              Types::DateTimeType,
              Items::DateTimeItem,
              Items::NumericItem
              other == self
```
fails and I am not sure if the best way to fix it.  I guess we could override '===' in GenericItem and the base Type class? Although I am not 100% sure what we need to do to override.  @ccutrer is this something that is possible or once we do that will we have other issues?

lib/openhab/core/item_proxy.rb
```diff
@@ -0,0 +1,33 @@
...
+
+require 'delegate'
+require 'forwardable'
+
+module OpenHAB
+  module Core
+    # Class is a proxy to underlying item
+    class ItemProxy < Delegator
+      extend Forwardable
+      def_delegators :__getobj__, :class, :is_a?, :tap, :then, :yield_self, :hash, :kind_of, :to_s
```
ccutrer, line 11: Why are you forwarding tap, then, and yield_self to the actual item? If it's because BasicObject (and thus SimpleDelegator) don't have those methods, you should re-implement them so that the actual item doesn't escape.

lib/openhab/core/item_proxy.rb
```diff
@@ -0,0 +1,33 @@
...
+      def initialize(item)
+        # Convert name to java version for faster lookups
+        @item_name = item.name.to_java
+      end
+      # rubocop:enable Lint/MissingSuper
+
+      #
+      # Lookup item from item registry
+      #
+      # rubocop:disable Style/GlobalVars
```
ccutrer, line 26: You should use single-line disable here (a comment at the end of the line you're disabling). Both because it's simpler, and because this line will leak into the Yard docs

ccutrer (commented):
@jimtng yup, that's the problem with delegation that I've never solved adequately. It's not GenericItem#=== you need to override... it's Module#=== (maybe you _could_ do GenericItem.===). Which might be feasible, but your overridden method needs to be able to identify that it's the proxy class. At that point I've overridden is_a? to return true if it's given the proxy class, or call super (or equivalent). I feel like I ran into problems at that point, but I can't remember what they are.

lib/openhab/core/item_proxy.rb
boc-tothefuture: tap/then/and yield_self were probably not correct to forward as they should be handled by the delegator.  I *think* we need to forward most methods in 'Module' i.e. class, is_a?, instance_of?, etc.  but I was mostly looking at the list of methods in Module and guessing.

ccutrer: yeah the rest of the list looked correct to me

boc-tothefuture:
I added a test for this use case (red/green testing worked).. So it is ready for review. 

On a related note @jimtng do you have any idea why the persistence tests are so finnicky?

jimtng:
LGTM

And a Happy New Year!

PR #480, opened 2022-01-04
fix(month_day_range): incorrect result when range crosses into the next year

### LINKED ISSUE #479: MonthDay range doesn't properly adjust ending_year/current_year
This code:
```
# Wrap to next year if ending day of month is before starting day of month
ending_year = ending < start ? next_year : current_year
```
Doesn't work correctly. 
1. It adjust ending year when it shouldn't -- right now it always increments the ending year even in cases when it shouldn't. e.g. 12/31 and 1/2 when testing on today (1/1) should be 12/31/2021 and 1/2/2022.  Instead it is 12/31/2022 and 1/2/2023.
2. There is no method to adjust backwards the starting_year. 

The current year needs to play into the decision about the start and end year. 

I don't quite know what the algorithm should be.  @ccutrer or @jimtng not sure if either of you have ideas on what the algorithm is.  I looked at it for a few minute and it didn't jump off the page to me.  I might come back after a break and see if that helps.

jimtng:
Fix #479 

@boc-tothefuture if you've got a neater fix, I'll be happy to close this PR in favour of yours.

### DIFF
--- a/docs/usage/misc/month_day.md
+++ b/docs/usage/misc/month_day.md
@@ -11,15 +11,16 @@ grand_parent: Usage
 
 MonthDay class from [java.time.MonthDay](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/MonthDay.html) can be used in rules for month-date related logic. Notable Methods:
 
-| Method       | Parameter  | Description                                                                                                                                                                  |
-| ------------ | ---------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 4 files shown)


jimtng:
This pr works because the year doesn't matter. It's just an offset for the monthdays.

I haven't checked if this is to be compared against an actual localdate, it should raise an exception I think, because it shouldn't be possible.

Removing the year completely could be a cleaner solution.

jimtng:
Closing - in favour of #483

PR #483, opened 2022-01-06
fix(monthday): fix monthday error when rolling over to another year

### LINKED ISSUE #479: MonthDay range doesn't properly adjust ending_year/current_year

boc-tothefuture:
Closes #479

Take a different approach (ignoring year) than #480

@jimtng are you ok with this approach?

### DIFF
--- a/lib/openhab/dsl/time/month_day.rb
+++ b/lib/openhab/dsl/time/month_day.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'java'
+
 module OpenHAB
   module DSL
     # Support for time related functions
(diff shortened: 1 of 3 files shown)


lib/openhab/dsl/time/month_day.rb
```diff
@@ -8,40 +10,21 @@ module Between
...
-
-          # Wrap to next year if ending day of month is before starting day of month
-          ending_year = ending < start ? next_year : current_year
+          start_range = DayOfYear.new(month_day: start, range: start..ending)
+          ending_range = DayOfYear.new(month_day: ending, range: start..ending)
+          logger.trace "Month Day Range Start(#{start}) - End (#{ending}) - Created from (#{range})"
 
-          start_range = MonthDayRangeElement.new(month_day: start, year: current_year)
-          ending_range = MonthDayRangeElement.new(month_day: ending, year: ending_year)
           range.exclude_end? ? (start_range...ending_range) : (start_range..ending_range)
```
jimtng, line 26: Not a suggestion, just a question for my edification: why is the above preferred to:
```ruby
Range.new(start_range, ending_range, range.exclude_end?)
```

lib/openhab/dsl/time/month_day.rb
```diff
@@ -52,54 +35,80 @@ def self.range?(range)
...
           end
-          # rubocop:enable Metrics/AbcSize
+        end
+      end
+
+      java_import java.time.Month
+      # Extend Month with helper method
+      class Month
+        # Calcalute and memoize the maximum number of days in a year before this month
+        # @return [Number] maximum nummber of days in a year before this month
```
jimtng, line 109:
```suggestion
        # Calculate and memoize the maximum number of days in a year before this month
        # @return [Number] maximum number of days in a year before this month
```

lib/openhab/dsl/time/month_day.rb
```diff
@@ -52,54 +35,80 @@ def self.range?(range)
...
+        def max_days_before
+          @max_days_before ||= Month.values.select { |month| month < self }.sum(&:max_length)
```
jimtng, line 111: This would save us from running the select loop, perhaps at the expense of readability - your call
```suggestion
          @max_days_before ||= Month.values.sum(&:max_length) - max_length
```

jimtng (commented):
This is a great and neat implementation. I like it.

PR #489, opened 2022-01-13
refactor(logging)!: rename log prefix from jsr223.jruby to org.openhab.automation.jruby

### LINKED ISSUE #464: Change the logger prefix to org.openhab.automation.script
Currently the logger prefix jsr223.jruby needs to be manually set to INFO in the karaf console, otherwise the logging would not be visible.  It seems that many new users stumbled on this issue. JS Scripting uses `org.openhab.automation.script` as its logger prefix. Should we change our logger to match?

jimtng:
Resolve #464 

jsr223.jruby is a custom log prefix that would require users to specifically enable in the karaf console. This is an extra step that could be avoided to improve user's first time experience.

org.openhab prefix is set to INFO by default

In Openhab 3.2 distro, a new prefix was created `org.openhab.automation.script`, however it is set to TRACE by default, making it impractical for us to use.

This PR chose `org.openhab.automation.jruby` as a unique log prefix to use, relying on the `org.openhab` default as INFO.

### DIFF
--- a/docs/usage/misc/logging.md
+++ b/docs/usage/misc/logging.md
@@ -9,25 +9,23 @@ grand_parent: Usage
 
 # Logging
 
-Logging is available everywhere through the logger object. 
+Logging is available everywhere through the logger object.
 
 Logging placed outside of a rule will have the name of the file appened to the logger name. The following entries are in a file named 'log_test.rb'
 
 ```ruby
-logger.trace('Test logging at trace') # 2020-12-03 18:05:20.903 [TRACE] [jsr223.jruby.log_test               ] - Test logging at trace
-logger.debug('Test logging at debug') # 2020-12-03 18:05:32.020 [DEBUG] [jsr223.jruby.log_test               ] - Test logging at debug
-logger.warn('Test logging at warn') # 2020-12-03 18:05:41.817 [WARN ] [jsr223.jruby.log_test               ] - Test logging at warn
-logger.info('Test logging at info') # 2020-12-03 18:05:41.817 [INFO ] [jsr223.jruby.log_test               ] - Test logging at info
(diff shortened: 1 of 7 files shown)


jimtng:
Failing on gem_install

jimtng:
@boc-tothefuture I can't figure out why it's failing on gem_install. Would you mind taking a look please?

boc-tothefuture:
I can look this weekend - it does pass locally for me.

jimtng:
I figured out why gem_install was failing. This test clears GEM_HOME and installs the openhab-jruby library from rubygems.org, i.e. not the version in the PR that contained the changes to logger prefix. The rubygems.org version is still logging to jsr223.jruby and in this PR, that prefix is not enabled in karaf, so its logs never showed up.

I stumbled on the same issue in another PR that's caused by this.

It's fixed by not loading a rule after cleaning the GEM_HOME.

boc-tothefuture:
Awesome - great find.  This needs to be marked as breaking correct?  It will impact people's running setups.

jimtng:
> This needs to be marked as breaking correct? It will impact people's running setups.

I suppose if they have set jsr223.jruby to other than INFO, it would change things for them, but I don't think it would be breaking as such - it doesn't change the behaviour/outcome of their rules unless their rules read from / rely on the contents of the log file?

boc-tothefuture:
Yes, I think it is still breaking.  If they set something to debug level in the old path it won't be in debug after this upgrade.

jimtng:
I've added the breaking flag

jimtng:
After all these breaking changes, we'll have to update the installation documentation, possibly also the one for the bundle.

PR #490, opened 2022-01-14
fix(java_class): fix java to ruby class

### LINKED ISSUE #482: convert java_class to ruby_class
The jrubyscripting bundle that shipped with openhab 3.2 contains a bug where Openhab injected global scope which contains java classes. As a result, they are not usable in JRuby, unless the classes are re-imported. This in turns breaks rules that don't use openhab-jruby helper library. 

The openhab-jruby helper library re-imports most of the relevant classes which masked the problem.

In December 2021, a new PR made it into the release of Openhab 3.2 which added ZonedDateTime, Duration and ChronoUnit. This caused problems in rules that use those java classes without re-importing them.

I made a patch to the jrubyscripting bundle to intercept the objects / classes and replace them with a jruby class. This jar file can be used for testing:
https://github.com/jimtng/openhab-addons/raw/jruby-scope/bundles/org.openhab.automation.jrubyscripting/org.openhab.automation.jrubyscripting-3.2.0-SNAPSHOT.jar

The source code can be found here:
https://github.com/jimtng/openhab-addons/blob/jruby-scope/bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/JRubyScriptEngineFactory.java

I'll clean it up before submitting a PR to openhab-addons.

_Originally posted by @jimtng in https://github.com/boc-tothefuture/openhab-jruby/issues/448#issuecomment-985964964_

jimtng:
Resolve #482 

This PR makes openhab-jruby library compatible with the fix to be made in the jrubyscripting bundle that addresses #482. This PR should be compatible with the bundle before the fix (i.e. the bundle that shipped with openhab 3.2) as well.

### DIFF
--- a/features/gem_install.feature
+++ b/features/gem_install.feature
@@ -5,6 +5,9 @@ Feature:  gem_install
     Given Clean OpenHAB with latest Ruby Libraries
 
   @reset_library
+  # This will install the current release version from rubygems.org
+  # Running a rule here may cause errors if the new version is not
+  # compatible with the release version
   Scenario: Install OpenHAB helper library
     Given OpenHAB is stopped
     And GEM_HOME is empty
@@ -14,12 +17,6 @@ Feature:  gem_install
       org.openhab.automation.jrubyscripting:gems=openhab-scripting=~>4.0
       org.openhab.automation.jrubyscripting:rubylib=<%= ruby_lib_dir %>
       """
-    And code in a rules file:
-      """
-      logger.info("OpenHAB helper library is at #{OpenHAB::VERSION}")
-      """
     When I start OpenHAB
     Then It should log 'Installing Gem: openhab-scripting' within 180 seconds
-    And I deploy the rules file
-    Then It should log 'OpenHAB helper library is at' within 200 seconds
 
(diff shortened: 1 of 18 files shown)


ccutrer (approved):

jimtng:
Note that I tested this against the new bundle that converted the global scope from java class to ruby class. This was done by making the build environment use the new addon. So this PR should work on either version of the bundle.

boc-tothefuture:
So we need a test?  There are a few tests that use DateTime, want to just uncomment the java_import on those?

We can do it in another or if you prefer?

jimtng:
> So we need a test? There are a few tests that use DateTime, want to just uncomment the java_import on those?
> 
> We can do it in another or if you prefer?

Could you please elaborate? What test, and which line to uncomment?

boc-tothefuture:
Sorry, I meant delete the line, not uncomment.  On mobile here, let me see if I can find it.

jimtng:
Testing against the new bundle should be in another PR, once the new bundle is available.



> Sorry, I meant delete the line, not uncomment. On mobile here, let me see if I can find it.

Are you referring to the date_time_type.rb where it imports ZonedDateTime?

boc-tothefuture:
Nevermind, I thought we had a feature test where to make it work we had to do a java_import in the body of the rest but I don't see it.

LGTM.

PR #491, opened 2022-01-14
fix(exception_handling): re-raising java exception caused an error

### LINKED ISSUE #406: re_raise_with_backtrace failed when a Java exception occurred
I happened to cause a `Java::JavaLang::ClassCastException` exception in my code, and the `re_raise_with_backtrace` failed with an error:
```
Error during evaluation of script 'file:/Users/jimmy/Projects/openhab-jruby/tmp/openhab/conf/automation/jsr223/ruby/personal/cucumber_test20211120-58372-1z6lli.rb': Error during evaluation of Ruby in /Users/jimmy/Projects/openhab-jruby/tmp/openhab/conf/automation/lib/ruby/gem_home/gems/openhab-scripting-4.14.0/lib/openhab/dsl/rules/rule.rb at line 77: (TypeError) exception class/object expected
```

Do we need to encapsulate the Java exception into a ruby Exception before re-raising it?

jimtng:
Resolve  #406

### DIFF
--- a/lib/openhab/log/logger.rb
+++ b/lib/openhab/log/logger.rb
@@ -82,6 +82,16 @@ module OpenHAB
         error
       end
 
+      #
+      # Print error and stack trace without calls to internal classes
+      #
+      # @param [Exception] error A rescued error
+      #
+      def log_exception(exception, rule_name)
+        exception = clean_backtrace(exception)
+        error { "#{exception.message} (#{exception.class})\nIn rule: #{rule_name}\n#{exception.backtrace&.join("\n")}" }
+      end
+
       private
 
       #
--- a/features/rule_language.feature
+++ b/features/rule_language.feature
@@ -177,7 +177,7 @@ Feature: rule_language
       | not_if  |
 
   @log_level_changed
-  Scenario: Native java exceptions are handled
+  Scenario: Native java exceptions are handled during rule execution
     Given log level INFO
     And code in a rules file
       """
@@ -196,6 +196,23 @@ Feature: rule_language
     And It should log "RUBY.test" within 5 seconds
     And It should log "RUBY.<main>" within 5 seconds
 
+  Scenario: Native java exceptions are handled during rule creation
+    Given log level INFO
+    And code in a rules file
+      """
+      def test
+        Java::JavaLang::Integer.parseInt('k')
+      end
+
+      rule 'test' do
+        test
+      end
+      """
+    When I deploy the rules file
+    Then It should log 'For input string: "k" (Java::JavaLang::NumberFormatException)' within 5 seconds
+    And It should log 'In rule: test' within 5 seconds
+    And It should log "RUBY.test" within 5 seconds
+    And It should log "RUBY.<main>" within 5 seconds
 
   Scenario: OpenHAB config directory is available
     Given code in a rules file:
--- a/features/step_definitions/openhab.rb
+++ b/features/step_definitions/openhab.rb
@@ -40,7 +40,7 @@ Then('It should log a line matching regex {string} within {int} seconds') do |re
   end
 end
 
-# rubocop:disable Metrics/LineLength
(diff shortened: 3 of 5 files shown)


lib/openhab/log/logger.rb
```diff
@@ -79,7 +79,7 @@ def clean_backtrace(error)
           backtrace = error.stack_trace.reject { |line| JAVA_INTERNAL_CALL_REGEX.match? line.to_s }
           error.set_stack_trace(backtrace)
         end
-        error
+        error.is_a?(Exception) ? error : JavaException.new(error.message)
```
ccutrer, line 82: I'd recommend passing the original error to JavaException.new, and then in an initialize method in that class store the entire error to an ivar, and only pass the message to super. You should also look if jruby already has a similar Ruby class defined

jimtng: Thanks for your suggestion. What should be done with that ivar?

The first problem is that I don't know how to reproduce this issue. So far I very rarely (but not never) encountered it on my production system. However, I see this error most often in the cucumber tests on the github workflow.

Also upon further investigation, this `error` came from `StandardError => e`

https://github.com/boc-tothefuture/openhab-jruby/blob/409c7a07586dc0a2cdf828317a72790a90ccf51a/lib/openhab/dsl/rules/rule.rb#L51

So it seems that it should've already been an instance of StandardError, but why re-raising it can cause 
```
(TypeError) exception class/object expected
```

ccutrer: Just accessible with an attr_reader. As for how it could happen... no idea.

ccutrer: Maybe add a rescue around the raise that if it does a TypeError, log the details of what you're trying to raise and then re-raise (the TypeError).

Perhaps there's a funny jruby thing going on that StandardError#=== is catching something Java-ish, but internally it can't re-raise

jimtng: Who is going to consume / use that attribute? Do I just pick an arbitrary name, e.g. `exception_object` or something?

ccutrer: Anyone interested in the original error. On second thought, there might be an implicit `cause` accessor that's populated. Please to double check.

jimtng: > Maybe add a rescue around the raise that if it does a TypeError, log the details of what you're trying to raise and then re-raise (the TypeError).

That's actually a brilliant idea!

jimtng: I've removed the JavaException class for now and just log the error that caused the TypeError as you suggested.

jimtng:
Note: need to change the commit message to something like:
`refactor(exception_handling): log the exception that caused TypeError`

ccutrer (changes requested):
Ah. Yes, this is not the right way. https://github.com/ccutrer/homie-mqtt/blob/6fcfe51e76bc23494c6f67ce5e28422b3a9087a4/lib/mqtt/homie/device.rb#L140 is the proper way to change the backtrace of an existing exception then re-raise it.

jimtng:
Yes, it was calling `raise e, message` with e an exception object, not a class, hence TypeError.

jimtng:
I figured out how to trigger this error: An error that occurred during rule creation, not execution. That's the only place re_raise_with_backtrace is called. I've now added a cucumber test for this scenario.

Upon further investigation, other places that called clean_backtrace don't re-raise the exception - I suppose due to the difficulties we're facing here.

So I opted to do the same, and use print_backtrace instead, which is now moved to logger to DRY it.

I am still unsure whether print_backtrace (which calls logger.error) is the ideal way to do it, or whether it should be re-raised. However, I'm happy with this for now unless anyone has objections.

jimtng:
I renamed `print_backtrace` to `log_exception`. This is ready to be reviewed / merged.

PR #498, opened 2022-01-16
feat(rules): support ranges as rule conditions

boc-tothefuture:
This PR supports ranges as rule conditions for from and to

Additionally, to support this some refactoring was done:
1. Change from @trigger_delays to @trigger_conditions 
2. Simplification of automation rule, but making TriggerDelay a TriggerCondition and moving it into its own class

### DIFF
--- a/docs/usage/triggers/changed.md
+++ b/docs/usage/triggers/changed.md
@@ -20,7 +20,9 @@ grand_parent: Usage
 Changed accepts Items, Things or Groups. 
 To and from accept arrays to match multiple states
 
-The from and to values operate exactly as they do in the DSL and Python rules with the exception of operating on Things.  If changed element being used as a trigger is a thing than the to and from values will accept symbols and strings, where the symbol matches the [supported status](https://www.openhab.org/docs/concepts/things.html). 
+The from and to values operate like they do in the DSL and Python rules with some enhancements:
+1. If the changed element being used as a trigger is a thing, the `to` and `from` values will accept symbols and strings, where the symbol matches the [supported status](https://www.openhab.org/docs/concepts/things.html). 
+2. Support for ranges.
 
 The for parameter provides a method of only executing the rule if the value is changed for a specific duration.  This provides a built-in method of only executing a rule if a condition is true for a period of time without the need to create dummy objects with the expire binding or make or manage your own timers.
 
@@ -50,6 +52,23 @@ rule 'Execute rule when item is changed to specific number, from specific number
 end
 ```
 
+Works with ranges:
+```ruby
+rule 'Execute rule when item is changed to a range of numbers, from a specific range of numbers, for specified duration' do
+  changed Alarm_Mode, from: 8..10, to: 12..14, for: 12.seconds
+  run { logger.info("Alarm Mode Updated")}
+end
+```
+
+Works with endless ranges:
+```ruby
+rule 'Execute rule when item is changed to any number greater than 12'
+  changed Alarm_Mode, to: (12..)   # Parenthesis required for endless ranges
+  run { logger.info("Alarm Mode Updated")}
+end
+```
+
+
 Works with things as well:
 ```ruby
 rule 'Execute rule when thing is changed' do
@@ -57,7 +76,6 @@ rule 'Execute rule when thing is changed' do
    run { |event| logger.info("Thing #{event.uid} status <trigger> to #{event.status}") }
 end
 ```
-
 
 Real world example:
 ```ruby
--- a/docs/usage/triggers/updated.md
+++ b/docs/usage/triggers/updated.md
@@ -17,7 +17,7 @@ grand_parent: Usage
 
 Changed accepts Items, Things or Groups. 
 
-The to value restricts the rule from running to only if the updated state matches. If the updated element being used as a trigger is a thing than the to and from values will accept symbols and strings, where the symbol matches the [supported status](https://www.openhab.org/docs/concepts/things.html). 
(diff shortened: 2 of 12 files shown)


docs/usage/triggers/changed.md
```diff
@@ -20,7 +20,9 @@ grand_parent: Usage
 Changed accepts Items, Things or Groups. 
 To and from accept arrays to match multiple states
 
-The from and to values operate exactly as they do in the DSL and Python rules with the exception of operating on Things.  If changed element being used as a trigger is a thing than the to and from values will accept symbols and strings, where the symbol matches the [supported status](https://www.openhab.org/docs/concepts/things.html). 
+The from and to values operate like they do in the DSL and Python rules with some enhabnacements.
+1. If changed element being used as a trigger is a thing than the to and from values will accept symbols and strings, where the symbol matches the [supported status](https://www.openhab.org/docs/concepts/things.html). 
+2. Support for ranges
```
jimtng, line 25:
```suggestion
The from and to values operate like they do in the DSL and Python rules with some enhancements:
1. If the changed element being used as a trigger is a thing, the `to` and `from` values will accept symbols and strings, where the symbol matches the [supported status](https://www.openhab.org/docs/concepts/things.html). 
2. Support for ranges.
```

docs/usage/triggers/updated.md
```diff
@@ -17,7 +17,7 @@ grand_parent: Usage
 
 Changed accepts Items, Things or Groups. 
 
-The to value restricts the rule from running to only if the updated state matches. If the updated element being used as a trigger is a thing than the to and from values will accept symbols and strings, where the symbol matches the [supported status](https://www.openhab.org/docs/concepts/things.html). 
+The `to` value restricts the rule from running to only if the updated state matches. If the updated element being used as a trigger is a thing than the to and from values will accept symbols and strings, where the symbol matches the [supported status](https://www.openhab.org/docs/concepts/things.html).  The `to` value supports ranges.
```
jimtng, line 20:
```suggestion
The `to` value restricts the rule from running only if the updated state matches. If the updated element being used as a trigger is a thing, the `to` and `from` values will accept symbols and strings, where the symbol matches the [supported status](https://www.openhab.org/docs/concepts/things.html).  The `to` value supports ranges.
```

lib/openhab/dsl/rules/triggers/conditions/range.rb
```diff
@@ -0,0 +1,74 @@
...
+            # Process rule
+            # @param [Hash] inputs inputs from trigger
+            #
+            def process(mod:, inputs:) # rubocop:disable Lint/UnusedMethodArgument - mod is unused here but required
+              logger.trace("Checking #{inputs} against condition trigger #{self}")
+              yield if check_from(inputs: inputs) && check_to(inputs: inputs)
+            end
+
+            #
+            # Check if from condition match the in inputs
```
jimtng, line 36:
```suggestion
            # Check if from condition match the inputs
```

jimtng (changes requested):
Great work! You must have read my mind. I was hesitant to raise the request for this feature, but I'm glad you went ahead and implemented it. This will simplify some of my rules.

I'd suggest adding some more tests that involve the edges of the range and also ranges with excluded end.

boc-tothefuture:
> Great work! You must have read my mind. I was hesitant to raise the request for this feature, but I'm glad you went ahead and implemented it. This will simplify some of my rules.
> 
> I'd suggest adding some more tests that involve the edges of the range and also ranges with excluded end.

I did it after looking at the fridge temperature rule you posted -- although this won't solve that specific case. 

Did you have any specific tests in mind? I wasn't trying to test if the Ruby Range class works, just that we process ranges correctly.

jimtng:
> I did it after looking at the fridge temperature rule you posted -- although this won't solve that specific case.

Exactly, that's the rule that made me want to have this feature. I was just thinking about how to use this for that rule, perhaps in combination with `otherwise`. I was also thinking of using `for` (changed duration feature) but that would require the inverse of this PR, i.e. 
```rb
changed Fridge_Temperature, not_to: '2 °C'..'6 °C', for: 5.minutes
```

This feature will also be handy to check for SunElevation. I have a rule that gradually changes a light's colour temperature  based on changes in SunElevation during the twilight hours.



> Did you have any specific tests in mind? I wasn't trying to test if the Ruby Range class works, just that we process ranges correctly.

Nothing specific, and I understand that this is not specifically for testing the range itself. Happy to drop this request.

jimtng:
I currently implemented my colour temperature changing rule like this:
```rb
rule 'Dawn/Dusk: Change light temperature' do
  changed Sun_Elevation
  only_if { ELEVATION_RANGE.cover? Sun_Elevation }
```

boc-tothefuture:
> > I did it after looking at the fridge temperature rule you posted -- although this won't solve that specific case.
> 
> Exactly, that's the rule that made me want to have this feature. I was just thinking about how to use this for that rule, perhaps in combination with `otherwise`. I was also thinking of using `for` (changed duration feature) but that would require the inverse of this PR, i.e.
> 
> ```ruby
> changed Fridge_Temperature, not_to: '2 °C'..'6 °C', for: 5.minutes
> ```
> 
> This feature will also be handy to check for SunElevation. I have a rule that gradually changes a light's colour temperature based on changes in SunElevation during the twilight hours.
> 
> > Did you have any specific tests in mind? I wasn't trying to test if the Ruby Range class works, just that we process ranges correctly.
> 
> Nothing specific, and I understand that this is not specifically for testing the range itself. Happy to drop this request.

Let's move this part to another issue.  I will merge the range here and we can talk about other implementations in another issue.

Issue #493, opened 2022-01-15
Implement error handing for Rest class


jimtng:
The testing infrastructure needs better error handling.

Error:
```
  Background:                                      # features/stop_move_type.feature:4
    Given Clean OpenHAB with latest Ruby Libraries # features/step_definitions/openhab.rb:6
      no implicit conversion of String into Integer (TypeError)
      ./features/support/openhab.rb:169:in `block in delete_rules'
      ./features/support/openhab.rb:168:in `delete_rules'
      ./features/step_definitions/openhab.rb:7:in `nil'
      features/stop_move_type.feature:5:in `Clean OpenHAB with latest Ruby Libraries'
```

jimtng:
I'm not sure what caused this error - it seems that the Rest.rules didn't return the expected json data. I'm wondering if it's because openhab isn't ready yet. It's just a guess, but I can add `start_openhab` in the `Given('Clean OpenHAB with latest Ruby Libraries')`. `start_openhab` will call the rake openhab:start which will wait until openhab is ready.

Do you have any objections to this @boc-tothefuture ?

It will change the cucumber behaviour in that it will start openhab if it's not already started, whereas previously it would just quit with a bunch of errors.

boc-tothefuture:
I have also found that sometimes OpenHab just doesn't respond to a rest request.. not sure if it had to do with it being started.

I am not sure if OpenHab start will break any of our tests?  It will probably slow them all down however.

jimtng:
I just came across another error while cucumber is waiting for the rule to be loaded. This all relates to #425 

Error: 
https://github.com/boc-tothefuture/openhab-jruby/runs/4824019929?check_suite_focus=true

```
    When I deploy the rules file                            # features/step_definitions/openhab_rules.rb:117
      Rule not added (RuntimeError)
      ./features/support/support.rb:10:in `wait_until'
      ./features/step_definitions/openhab_rules.rb:61:in `wait_for_rule'
      ./features/step_definitions/openhab_rules.rb:75:in `deploy_ruby_file'
      ./features/step_definitions/openhab_rules.rb:50:in `deploy_rule'
      ./features/step_definitions/openhab_rules.rb:118:in `nil'
      features/comparisons.feature:585:in `I deploy the rules file'
```

boc-tothefuture:
An alternative, and I thought I implemented this was to retry those failed requests.

jimtng:
[...]

PR #499, opened 2022-01-16
test(rest): retry rest requests on rules deletion

boc-tothefuture:
Performs retry as discussed in #493

### DIFF
--- a/features/support/openhab.rb
+++ b/features/support/openhab.rb
@@ -177,14 +177,23 @@ def delete_things
   openhab_client('openhab:things clear')
 end
 
-def delete_rules
+def delete_rules # rubocop:disable Metrics/MethodLength
   FileUtils.rm Dir.glob(File.join(rules_dir, '*.rb'))
   deleted = false
-  Rest.rules.each do |rule|
-    uid = rule['uid']
-    Rest.delete_rule(uid)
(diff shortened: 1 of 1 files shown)


features/support/openhab.rb
```diff
@@ -177,14 +177,23 @@ def delete_things
...
-    deleted = true
+  begin
+    retries ||= 0
+    Rest.rules.each do |rule|
+      uid = rule['uid']
+      Rest.delete_rule(uid)
+      deleted = true
+    end
+  rescue StandardError => e
+    raise e unless (retries += 1) < 3
```
ccutrer, line 191: You should just `raise` not `raise e`. The latter resets the backtrace as if the exception is thrown from this point, whereas the former preserves the original backtrace.

ccutrer (approved):

jimtng:
Thanks for this. I was wondering if we could implement a retry within the Rest class itself, by redefining the `#get` method, so that all the other calls would benefit from it?

boc-tothefuture:
I was a bit concerned about doing that because I wasn't sure what the possible side effects would be.  I was thinking to try this here and see if we ever got failures anywhere else before making it broader.

jimtng (approved):

PR #500, opened 2022-01-17
refactor(cron): use trigger handler instead of secondary rule

### LINKED ISSUE #446: Determine if secondary rule for cron attach triggers can be replaced with new trigger type
Determine if this is feasible after creating new directory watch trigger

boc-tothefuture:
This closes #446 

Simplifies the code to not have a different rule for cron type triggers with attachments, instead a cron trigger handler is created that supplies the trigger id - enabling cron processing in rules to be the same as the other triggers.


jimtng:
Another great PR!

PR #501, opened 2022-01-17
feat(conditions): support procs/lambdas for trigger conditions

boc-tothefuture:
This permits the use of procs and lambda procs in the from/to conditions for complex state matching cases (not a value, regex, etc). 

Deduplicates some code  between the duration and and other conditions.


boc-tothefuture:
@jimtng 
I think this would no permit to change this:
```
require 'openhab'

rule 'Fridge Temperature' do
  description 'Warn when the fridge temperature is abnormal for more than 5 mins'
  changed Fridge_Temperature
  triggered do |item|
    if item.between?('2 °C', '6 °C') # the acceptable range depends on the placement of the thermometer
      timers[item]&.each(&:cancel)
      next
    end

    next if timers[item] # do not reschedule the timer

    after(5.minutes, id: item) do |timer|
      message = "Warning, the #{item.id} is not within normal parameters. "
      message << "The #{item.id} is currently #{item.state.format('%.1f %unit%')}"
      gEchoTTS << message
      timer.reschedule unless timer.cancelled?
    end
  end
end
```

Into: 
```
require 'openhab'

rule 'Fridge Temperature' do
  description 'Warn when the fridge temperature is abnormal for more than 5 mins'
  changed Fridge_Temperature, to: ->to { !to.between?('2 °C', '6 °C')  }, for: 5.minutes
  triggered do |item|
      message = "Warning, the #{item.id} is not within normal parameters. "
      message << "The #{item.id} is currently #{item.state.format('%.1f %unit%')}"
      gEchoTTS << message
    end
end
```

jimtng:
You read my mind again! I was thinking about exactly this

jimtng:
What I was wondering was whether we should do 
```ruby
changed Item, from: lambda, to: lambda
```

or

```ruby
changed Item, conditions: lambda { |from, to| code }
```

Yes, the first syntax is simpler, easier, and requires the least amount of change, and I prefer it too. Just throwing the second syntax out there for consideration as well. WDYT?

jimtng (commented):
This would also be great for `command:` in received_command trigger

boc-tothefuture:
Right now I think the from/to aligns better with the rest of the dsl.  

Also, with the second method you have to always define from and to even if you only use one of them.

I haven't found a good use case where you need to compare the from and two where they aren't independent. 

However, we could always add 'conditions' later - they wouldn't be mutually exclusive.

jimtng:
> Into:
> 
> ```
> require 'openhab'
> 
> rule 'Fridge Temperature' do
>   description 'Warn when the fridge temperature is abnormal for more than 5 mins'
>   changed Fridge_Temperature, to: ->to { !to.between?('2 °C', '6 °C')  }, for: 5.minutes
>   triggered do |item|
>       message = "Warning, the #{item.id} is not within normal parameters. "
>       message << "The #{item.id} is currently #{item.state.format('%.1f %unit%')}"
>       gEchoTTS << message
>     end
> end
> ```

The only problem with this is that the "warning" will never be issued if the fridge temperature keeps changing in less than 5 minutes.

boc-tothefuture:
> This would also be great for `command:` in received_command trigger

I will open another issue for that.

jimtng:
> I haven't found a good use case where you need to compare the from and two where they aren't independent.

I guess any time you need to compare from and to whilst taking advantage of the `for` feature, e.g. to detect that the temperature continues to increase (from < to) for 5 minutes. I agree that should we decide to add this (and I'm not sure that we should), it would be in a separate PR. In any case, the "for" feature is the key here, otherwise it could just be done in a guard clause.

PR #513, opened 2022-01-20
test(delete_rules): fix error when deleting all rules

jimtng:
A possible fix for #493 

~~It seems that deleting the rules via REST is redundant after deleting the rule files and might in fact cause some sort of problems in openhab. This PR removes the rule deletion via REST and just waits longer for openhab to complete the rules removal.~~ 

Need to run the checks a few times to see if a problem might still occur during delete_rules


jimtng:
Closing this and combining it into #514

PR #524, opened 2022-02-02
fix(hsbtype): comparison against another hsbtype only checks for brightness

jimtng:
The solution was provided by @ccutrer


ccutrer (approved):
I was really confused because https://boc-tothefuture.github.io/openhab-jruby/usage/items/color/ shows we already have to_a. but then I realized that's ColorItem. so, another PR, but we should drop that into HSBType (especially since it's fancier).

jimtng:
@ccutrer, thanks for pointing that out. I wasn't familiar with the whole color stuff. I've moved to_h and to_a into hsb_type in a second commit. WDYT?

ccutrer (approved):

PR #527, opened 2022-02-21
fix(logger): move logger class into core module to prevent unintentional modification

### LINKED ISSUE #523: Conflict with httparty
My script doesn't run after I require 'httparty'

Example:
```ruby
require 'openhab'
logger.info('Before httparty')
require 'httparty'
logger.info('After httparty')
```

It will log `Before httparty` but not `After httparty`

My script seems to work fine if I put the require httparty line before require openhab. I haven't yet tested whether this will break httparty, as I haven't started using it.

Why is this happening though?

jimtng:
Fix #523

Because OpenHAB::Log gets included into the top level, the OpenHAB::Log::Logger class then lives on the top level. Due to its generic name, it can be reopened/modified by other libraries that innocently creates their own Logger class. In some cases this can cause errors.

Some libraries known to cause such a problem are httparty and koala.

This PR moves OpenHAB::Log::Logger class to OpenHAB::Core::Logger avoid this issue.

### DIFF
--- a/features/logging.feature
+++ b/features/logging.feature
@@ -99,4 +99,15 @@ Feature:  logging
     And item "Foo" state is changed to "5"
     Then It should log only "Baz" at level "INFO" from 'org.openhab.automation.jruby.log_file.log_test' within 8 seconds
 
+  Scenario: Logger works after requiring third party gems
+    Given code in a rules file
+      """
+      gemfile do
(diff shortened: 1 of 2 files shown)


ccutrer (approved):

jimtng:
Change: Instead of renaming the class, I moved it into OpenHAB::Core.

PR #529, opened 2022-02-23
fix(dsl): dsl methods leaked into other objects

### LINKED ISSUE #294: sort out DSL method availability
DSL methods have been added to the object model in several ways inconsistently, and their visibility needs to be sorted out.
 (constants too, but I want to address methods for now with this issue).

There should be three levels:
 * Modules `extend`ed onto `main` (the top-level object). These methods are _only_ available from the top level, or other top level methods/blocks. This should include things like `rule`.
 * Modules `include`d onto `Object` (implicitly what happens when you `include` onto `main`): These methods are available on _any_ object, anywhere, called in any way. I'm not sure what, if anything, truly belongs here. Candidates are entity lookup (`method_missing` and `const_missing` support that allows you to reference items and things as first-class named objects). These are kind of dangerous because any library a person might use will see these methods. And while anything they define will override stuff we add here, it still introduces possibilities for mis-understandings. Case in point - `OpenHAB::DSL::Items` is `include`d onto object, so the `items` method is available on _any_ object, and even tripped our tests because we have something calling `SomeGroupItem.items` which was returning _all items in OpenHAB_ rather than the expected items belonging just to that group item.
 * Modules `include`d onto `Object`, but with instance methods marked as `module_function`. `module_function` both marks the instance method as private, and copies it to the module's singleton. The end result is that the methods are callable explicitly (like `OpenHAB::DSL::Items.items`) _or_ as a "bare" method call (`items`), but never as an instance method on an object (`SomeGroupItem.items` -- even though technically the method exists, it's not callable because it's private -- this is how all the methods in `Kernel` are defined, like `puts`). Clearly I believe `items` belongs in this category :).

As an aside, even if something is in the first category, it can still be included into other contexts like `RuleConfig` for use in the block for setting up a rule. Right now this is exactly how `OpenHAB::Core::EntityLookup` works - it's `extend`ed onto `main`, and `include`d into `RuleConfig`.

`rule` is clearly in the first category.

I really can't think of anything that should be in the second category. But there are several things that I'm not sure if they should be in the first category or the third. Should `items` be available _anywhere_, or only at top-level, or within rule config and rule execution blocks? Basically, should this work in my own class?:

```ruby
class Something
  def my_method
    items["MySwitch"]
  end
end

rule "when something happens" do
  changed MySwitch
  run do
    Something.new.my_method
  end
end
```

Seems like most other methods fit in this category - actions, `items`, `things`, `groups`, `unit`.

jimtng:
Resolve #294

These were the methods that leaked into other objects
`:scriptLoaded, :scriptUnloaded, :unit, :with_unit, :timers, :after, :things, :store_states, :rule, :thread_local, :persistence, :items, :groups, :__conf__, :between, :say, :actions, :play_sound, :actions_for_thing, :notify, :logger`

* This PR removed all the leaks and removed 'pp' pretty_print inclusion
* Move most non-user classes into `Support` submodule of the corresponding module so they aren't available on the top level
* Clean up YARD documentation 
* Add `rake openhab:client` to conveniently launch the karaf client.
* Add `a raw rule:` cucumber phrase

After this PR is applied, these are the consts introduced after require 'openhab'
```
:ALL_DAY
:Addrinfo
:Audio
:AutomationRule
:Base64
:BasicSocket
:BigDecimal
:BigMath
:ColorItem
:ComparableItem
:Configuration
:Contact
:ContactItem
:Date
:DateTime
:DateTimeFormatter
:DateTimeFormatterBuilder
:DateTimeItem
:Dimmer
:DimmerItem
:Ensure
:Ephemeris
:Events
:Exec
:Forwardable
:GenericItem
:GroupItem
:Guard
:HTTP
:IPAddr
:IPSocket
:ImageItem
:ItemEquality
:LocalTime
:Locale
:LocationItem
:MIDNIGHT
:Marcel
:Metadata
:Month
:MonthDay
:MonthDayRange
:NOON
:Net
:NumberItem
:NumericItem
:NumericType
:OpenHAB
:OpenSSL
:Pathname
:Persistence
:PersistenceExtensions
:Ping
:PlayerItem
:Property
:Quantity
:RollershutterItem
:RuleConfig
:RuleTriggers
:ScriptExecution
:SecureRandom
:Semantics
:ServerSocket
:Set
:SingleForwardable
:Singleton
:Socket
:SocketError
:SortedSet
:StringIO
:StringItem
:Support
:Switch
:SwitchItem
:TCPServer
:TCPSocket
:TCPserver
:TCPsocket
:Terse
:Thing
:ThingStatus
:Things
:TimeOfDay
:TimeOfDayRangeElement
:TimedCommand
:Timeout
:TimeoutError
:Timer
:Transformation
:Triggers
:Type
:UDPSocket
:UDPsocket
:UNIXServer
:UNIXSocket
:UNIXserver
:UNIXsocket
:URI
:Voice
:Zlib
```

Top level constants that are removed by this PR
```
:AutomationRule						      <
:BusEvent						      <
:COMMAND_ALIASES					      <
:Ensure							      <
:Events							      <
:Groups							      <
:Guard							      <
:ItemRegistry						      <
:PP							      <
:PREDICATE_ALIASES					      <
:PrettyPrint						      <
:Property						      <
:ReentrantTimer						      <
:RuleConfig						      <
:RuleTriggers						      <
:StateStorage						      <
:Terse							      <
:TimerManager						      <
:TimerSet						      <
:Triggers						      <
```

### DIFF
--- a/features/rule_language.feature
+++ b/features/rule_language.feature
@@ -234,3 +234,37 @@ Feature: rule_language
       """
     When I deploy the rules file
     Then It should log /Rule UID: '.+'/ within 5 seconds
+    
+  Scenario: DSL methods don't leak into other objects
+    Given a raw rule:
+      """
+      original_methods = Class.methods
+
+      require 'openhab'
+
+      leaked_methods = (Class.methods - original_methods).sort
+      logger.info "Leaked methods: #{leaked_methods}"
+      """
+    When I deploy the rules file
+    Then It should log "Leaked methods: []" within 5 seconds
+
+  @wip
+  Scenario: Check constants introduced by the library
+    Given a raw rule:
+      """
+      def all_constants
+        Object.ancestors.map(&:constants).concat(Object.constants).flatten.uniq.sort
+      end
+
+      original_constants = all_constants
+
+      require 'openhab'
+
+      added_constants = (all_constants - original_constants).sort
+
+      # added_constants.each {|s| logger.info "#{s} => #{Object.const_get(s).class.name}" }
+      logger.info "Added constants: #{added_constants}"
+      logger.info "OK"
+      """
+    When I deploy the rules file
+    Then It should log "OK" within 5 seconds
--- a/features/step_definitions/openhab_rules.rb
+++ b/features/step_definitions/openhab_rules.rb
@@ -21,7 +21,8 @@ def identifying_log_line(uid)
 end
 
 def prepend_identifying_log_line_to_rule(code, uid)
-  code.insert require_openhab.length, %[\n\nlogger.info("#{identifying_started_log_line(uid)}")\n\n]
+  pattern = require_openhab.gsub %(['"]), %(['"]) # allow raw rules with `require "openhab"` and `require 'openhab'`
+  code.sub!(/#{pattern}/, %[#{require_openhab}\n\nlogger.info("#{identifying_started_log_line(uid)}")\n\n])
 end
 
 def append_identifying_log_line_to_rule(code, uid)
@@ -74,6 +75,13 @@ def deploy_ruby_file(code:, directory:, filename: '', check_position: :end, chec
   create_log_markers(code, uid) if check
   atomic_rule_write(code, File.join(directory, filename))
   wait_for_rule(log_line) if check
+end
+
+# A raw rule is one where we don't automatically insert `require 'openhab'`
+# It must be inserted manually in the code doc_string by the test author.
+# This gives the author control over what goes before the require line.
+Given('a raw rule(:)') do |doc_string|
+  @rule = doc_string
 end
 
 Given('a rule(:)') do |doc_string|
--- a/lib/openhab.rb
+++ b/lib/openhab.rb
@@ -19,16 +19,13 @@ module OpenHAB
   #
   # @param [Object] base Object to decorate with DSL and helper methods
   #
-  #
-  # Number of extensions and includes requires more lines
   def self.extended(base)
     OpenHAB::Core.wait_till_openhab_ready
-    base.extend Log
     base.extend OpenHAB::Core::ScriptHandling
+    base.extend OpenHAB::Core::ScriptHandlingCallbacks
     base.extend OpenHAB::Core::EntityLookup
     base.extend OpenHAB::DSL
 
-    base.send :include, OpenHAB::Core::ScriptHandlingCallbacks
     logger.debug "OpenHAB JRuby Scripting Library Version #{OpenHAB::VERSION} Loaded"
 
     OpenHAB::Core.add_rubylib_to_load_path
@@ -36,7 +33,4 @@ module OpenHAB
 end
 
 # Extend caller with OpenHAB methods
-
-# rubocop: disable Style/MixinUsage
-extend OpenHAB
-# rubocop: enable Style/MixinUsage
+extend OpenHAB # rubocop:disable Style/MixinUsage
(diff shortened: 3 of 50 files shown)


jimtng:
Code to get the list of consts:
```ruby
        def all_constants
          Object.ancestors.map(&:constants).concat(Object.constants).flatten.uniq.sort
        end

        original_consts = all_constants

        require 'openhab'

        logger.error (all_constants - original_consts).to_s
```

boc-tothefuture:
Wow - great work.  

I think we can drop PP as most people log, it got in there I think while debugging.

boc-tothefuture:
What is the difference between "raw rule:" and "Code in a rules file:" ?

jimtng:
> What is the difference between "raw rule:" and "Code in a rules file:" ?

raw rule doesn't insert `require 'openhab'`
One must type it in the test code

```
"""
require 'openhab'

logger.info "my test"
"""
```

jimtng:
> Wow - great work.
> 
> I think we can drop PP as most people log, it got in there I think while debugging.

I have now dropped pp from automation_rule.rb

jimtng:
The added submodule `Support` is showing up on the top level, and I don't think this is desirable. 

I'm exploring different options, e.g. moving them to `OpenHAB::Core`

jimtng:
I've done some more rearranging. 
* Do not include OpenHAB::DSL
* Hide the `Support` module inside OpenHAB::DSL instead of under their corresponding module
* Do not include OpenHAB::DSL::Rules because this will promote all the submodules of Rules to become top level modules: Property, Triggers, Guard, Terse
* Move the `rule()` method into a submodule of Rules::StandardRule and include it
* Minor refactor on StatesStorage

I'm leaving the multiple commits for review in case some things need to be reverted. 

I will squash before merge.

boc-tothefuture:
Can you change "raw rule" to "a file in the rules directory"?

jimtng:
> Can you change "raw rule" to "a file in the rules directory"?

I'm not sure I understand this.

```
Given a file in the rules directory:
   """
   rule code here?
   """
```

That doesn't make sense, because you're not giving it a file, you're give it a block of code

boc-tothefuture:
"file content in rules directory"?

jimtng:
The purpose of the new phrase is something that doesn't prepend `require 'openhab'` into the code

Given the two names: "file content in rules directory" vs "code in a rules file" how could you intuitively tell that they're different and that the prior wouldn't insert `require 'openhab'` and the latter would?

boc-tothefuture:
> The purpose of the new phrase is something that doesn't prepend `require 'openhab'` into the code
> 
> Given the two names: "file content in rules directory" vs "code in a rules file" how could you intuitively tell which of the two would insert the `require` and which wouldn't?

Given this change "code in a rules file' isn't good anymore either.  I was thinking that would have to change - but wasn't going to saddle this PR with the change.

jimtng:
I've rearranged the code in logger.rb so the changes are easier to see.

boc-tothefuture:
@jimtng is changing from raw rule still on your to-do list or do we need more discussion?

jimtng:
> @jimtng is changing from raw rule still on your to-do list or do we need more discussion?

I thought that we'll review / change the cucumber step names in a separate PR after this one is merged? I am happy to do it but I don't know if we've come to a conclusion about what they should ideally be called.

jimtng:
I have now squashed all the commits into one. Should be ready to merge unless there are any objections?

PR #532, opened 2022-02-27
refactor(on_off_type): reduce to one method call for #!

jimtng:
I noticed when calling SwitchItem.toggle, one comparison was done when the item was in the ON state, but two comparisons were done when the item was in the OFF state.

* When SwitchItem1 is ON and SwitchItem1.toggle is called:
```
23:15:06.403 [TRACE] [.openhab.core.library.types.OnOffType] - (Java::OrgOpenhabCoreLibraryTypes::OnOffType) ON == ON (Java::OrgOpenhabCoreLibraryTypes::OnOffType)
23:15:06.406 [TRACE] [openhab.core.library.items.SwitchItem] - Sending Command OFF to Switch1
```

* When SwitchItem1 is OFF and SwitchItem1.toggle is called:
```
23:15:08.773 [TRACE] [.openhab.core.library.types.OnOffType] - (Java::OrgOpenhabCoreLibraryTypes::OnOffType) OFF == ON (Java::OrgOpenhabCoreLibraryTypes::OnOffType)
23:15:08.776 [TRACE] [.openhab.core.library.types.OnOffType] - (Java::OrgOpenhabCoreLibraryTypes::OnOffType) OFF == OFF (Java::OrgOpenhabCoreLibraryTypes::OnOffType)
23:15:08.777 [TRACE] [openhab.core.library.items.SwitchItem] - Sending Command ON to Switch1
2
```

This PR makes it so there's only one comparison to be done in either case.

### DIFF
--- a/lib/openhab/dsl/types/on_off_type.rb
+++ b/lib/openhab/dsl/types/on_off_type.rb
@@ -20,8 +20,7 @@ module OpenHAB
         # Invert the type
         # @return [OnOffType] +OFF+ if +ON+, +ON+ if +OFF+
         def !
-          return OFF if on?
-          return ON if off?
+          on? ? OFF : ON
         end
       end
     end


ccutrer (commented):
This is not strictly equivalent. It was written this way so that if the item is in NULL or UNDEF, it will still return nil. Maybe we want to raise an error in that case though?

jimtng:
The OnOffType is an enum. Whilst a java enum object can be null, in such case, OnOffType#! won't be called (because the object is null?)

Moving up to the SwitchItem level, it already checks for state? and returns ON when state? returns false (state is NULL/UNDEF).

Have I missed something?

jimtng:
If this PR is wrong, then we should at least create a test case that fails it.

ccutrer (approved):
Nope you're not wrong. I was thinking this was on the SwitchItem level where UnDefType is valid.

PR #538, opened 2022-03-04
fix(ensure): implement ensure.update

jimtng:
I noticed that ensure.update is in the docs but wasn't implemented.

Also GroupMembers doesn't have #update. At this point I don't think it's worth adding, as I don't think anyone needs it.


ccutrer (approved):

boc-tothefuture:
@jimtng I think one of the other PRs created a merge conflict here.

jimtng:
I've rebased the branch

PR #550, opened 2022-03-08
feat(item_equality): add the ability to compare item objects

### LINKED ISSUE #541: Item comparison - possible confusion
I know, I know... I want everything!

Currently this code works, it compares an item against its state
```ruby
rule 'x' do
  changed SwitchItem1
  triggered do |item|
    case item
    when ON then logger.info("Hey It's on")
    when OFF then logger.info("Hey it's off")
    end
  end
end
```

One might not realise that and started writing code like this, comparing the triggering item against the possible items to act upon:
```ruby
rule 'y' do
  changed Switches.members
  triggered do |item|
    case item
    when SwitchItem1 then logger.info("Hey I'm switch item 1")
    when SwitchItem2 then logger.info("Hey I'm switch item 2")
    end
  end
end
```

Innocently that code would appear to be reasonable. After all, you're comparing item against item, but it would not work as "expected" because it'll compare the **states** of SwitchItem1 and SwitchItem2, against the triggering item.

So when SwitchItem1 was ON and SwitchItem2 was OFF
Then SwitchItem2 changed to ON, triggering rule y, it would match the first when clause because triggering item's state (ON) matches SwitchItem1's state (also ON) which happens to be the first when clause.

Also should `SwitchItem1 == SwitchItem2` be true when their states are the same? What would break if we changed this to become false, so event.item == SwitchItem1 would only be true if event.item is SwitchItem1.

To compare the states, it should be `SwitchItem1.state == SwitchItem2.state`

jimtng:
Fix #541 

* Add GenericItem#item method that returns GenericItemObject
* ItemEquality #== uses `eql?` when its argument is a GenericItemObject, i.e. `ItemName.item`
* This means users can use ItemName.item to ensure that they are operating on the item, not the state for equality comparison purposes
* Add ItemEquality #=== so that grep operates on the item without needing `.item`
* Update comparisons.md doc to clarify this issue

### DIFF
--- a/docs/usage/misc/comparisons.md
+++ b/docs/usage/misc/comparisons.md
@@ -1,6 +1,6 @@
 ---
 layout: default
-title: Type Comparisons
+title: Comparisons
 nav_order: 12
 has_children: false
 parent: Misc
@@ -9,7 +9,36 @@ grand_parent: Usage
 
 # Item and State Type Comparisons
 
-Some OpenHAB item types can accept different command types. For example, a Dimmer item can accept a command with an `OnOffType`, `IncreaseDecreaseType` or a `PercentType`. However, ultimately an item only stores its state in its native type, e.g. a Dimmer item's native type is PercentType.
+Comparison of two items implicitly compares their states. For example:
+
+```ruby
+if Item1 == Item2
+  logger.info('The state of Item1 is the same as the state of Item2')
+end
+```
+
+Comparing the actual items can be achieved using `#item` method of either operand:
+
+```ruby
+Item1.item == items['Item1'].item   # This would always return true
+Item1.item == Item2.item            # This would always return false
+
+# When Item1, Item2 and Item3 all have the same state:
+[Item1, Item2].include?(Item3.item) # => false - Check for the existence of Item3 inside the array
+[Item1, Item2].include?(Item3)      # => true - State check
+```
+
+Using `#item` is not necessary for hash keys because Hash keys are indexed by their hash codes.
+
+```ruby
+# When Item1 and Item2 contain the same state:
+hash = { Item1 => 'value1', Item2 => 'value2' }
+hash[Item2] # => 'value2' this will correctly match Item2's value in the hash
+```
+
+Some OpenHAB item types can accept different command types. For example, a Dimmer item can accept a command 
+with an `OnOffType`, `IncreaseDecreaseType` or a `PercentType`. However, ultimately an item only stores its 
+state in its native type, e.g. a Dimmer item's native type is PercentType.
 
 ## Loose Type Comparisons
 
--- a/docs/usage/misc/comparisons.md
+++ b/docs/usage/misc/comparisons.md
@@ -9,38 +9,103 @@ grand_parent: Usage
 
 # Item and State Type Comparisons
 
-Comparison of two items implicitly compares their states. For example:
+Comparisons of items implicitly compare the states. 
 
 ```ruby
-if Item1 == Item2
-  logger.info('The state of Item1 is the same as the state of Item2')
-end
-```
-
-Comparing the actual items can be achieved using `#item` method of either operand:
-
-```ruby
-Item1.item == items['Item1'].item   # This would always return true
-Item1.item == Item2.item            # This would always return false
-
 # When Item1, Item2 and Item3 all have the same state:
-[Item1, Item2].include?(Item3.item) # => false - Check for the existence of Item3 inside the array
-[Item1, Item2].include?(Item3)      # => true - State check
+
+# State comparison
+Item1 == Item2                      # => true - when they both have the same state
+Item1 == items['Item2']             # => true - same as above
+
+Item1 == ON                         # => true if Item1's state is ON
+ON == Item1                         # => true if Item1's state is ON
+
+case Item1
+when OFF then logger.info('Item1 is OFF')
+when ON then logger.info('Item1 is ON')
+end
+
+NumberItem1 > 5                     # Item can be compared against numbers
+Outside_Temp > '24 °C'              # with Units of Measurement
+StringItem1 == 'single'             # or string
+DateItem1 < Time.now                # or date/time
+DateItem1 < '2022-01-01'            # datetime against string
 ```
 
-Using `#item` is not necessary for hash keys because Hash keys are indexed by their hash codes.
+## Item Comparisons
+
+To compare item objects, use the `#item` method on either item:
 
 ```ruby
-# When Item1 and Item2 contain the same state:
-hash = { Item1 => 'value1', Item2 => 'value2' }
-hash[Item2] # => 'value2' this will correctly match Item2's value in the hash
+# When Item1, Item2 and Item3 all have the same state:
+
+# Item comparison
+Item1 == Item2.item                 # => false - different items
+Item1 == items['Item2'].item        # => false - different items
+Item1 == items['Item1'].item        # => true - same item
+# Using eql?
+Item1.eql?(Item2)                   # => false - different items
+Item1.eql?(items['Item1'])          # => true - same item
+
+# Case statement. Given Item1, Item2 and Item3 are OFF
+result =  case Item1                # => 'off' 
+          when OFF then 'off'       #   OFF === Item1 is true, first match
+          when Item2 then 'i2'
+          when Item1 then 'i1'
+          end
+
+result =  case Item1.item           # => 'i1' 
+          when OFF then 'off'       #   OFF === Item1.item   => false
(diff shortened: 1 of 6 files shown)


ccutrer:
so this fixes case statements, but doesn't fix `[Item1, Item2].include?(Item1)`, right?

jimtng:
> so this fixes case statements, but doesn't fix `[Item1, Item2].include?(Item1)`, right?

I forgot about this. I've added a test now but it failed. This calls the #== instead of #===. How can this be solved?

jimtng:
Array#include?, Hash#value? are affected, but there may be other cases too.

ccutrer:
You need to modify #== so that if other is a GenericItem it calls eql? exclusively (Instead of only returning true if eql? returns true, otherwise falling through which is what I think it does right now). With that the explicit implementation of === shouldn't be necessary.

jimtng:
That would fail the comparisons like
Number10 == Number10A (in comparisons.feature)

ccutrer:
> That would fail the comparisons like
> Number10 == Number10A (in comparisons.feature)

Yes it would. IMO it's a bug that Number10 compares equal to Number10A.

jimtng:
How about Number5 < Number10? Should that not be supported either?

ccutrer:
Inequality comparisons don't make sense in the context of item to item comparisons (only states, and only the case case of some types). I personally wouldn't allow them because they get you in the habit of thinking that items are comparable as states.

Keep in mind that this is my _opinion_, and isn't necessarily "right". There are arguments for the convenience of treating items as states in many contexts, I just don't think they're worth the confusion that arises in other contexts from doing so.

jimtng:
That would then extend to arithmetics like Number10 - Number5

In your opinion then it should be Number10.state - Number5.state?

@boc-tothefuture could you weigh in on this?

ccutrer:
Yuuup. Same reasoning would apply to those, and is where my argument starts to fall apart as you have to call #state more and more (and I freely admit to not calling state in several places in my own rules). And then I counter that if the state is nil (NULL or UNDEF), implicitly delegating to state just makes it even harder to debug what the problem is, because the error comes inside method_missing on GenericItem. Even worse to me is that the methods an Item responds to can change as the state changes to different types. Not necessarily a violation of anything, but can be quite surprising.

jimtng:
It does seem more "proper" to call state, but at the expense of convenience.

If we move to this, it would be a breaking change.

Alternatively, we could stay as is and sacrifice not being able to do array#include? because it is currently ambiguous whether an Item1 is the item object or its state. Convenience comes at a price

jimtng:
A proposal:

Considering that:
* arithmetic and state comparisons are far more commonly done than things like [Item1, Item1].include?(Item1)
* We'd like to keep this convenience(?)

As an alternative, we could do something like `[Item1, Item2].include?(Item1.obj) (or something that's more "idiomatic") to compare object instead of its value/state.

boc-tothefuture:
I am not a fan of having to call state for all comparisons.  Most often, at least for me, what I am comparing in an item is the state.

jimtng:
> I am not a fan of having to call state for all comparisons.  Most often, at least for me, what I am comparing in an item is the state. 

I somewhat agree and like the convenience of this. But how do you propose we compare items?

boc-tothefuture:
I will take a look at it this weekend.

boc-tothefuture:
`[Item1, Item2].include?(Item1.obj)`

I prefer that approach over breaking state/value comparisons.  Is `.item` better? Does `.class` currently work in this instance if we are seeing if it is a specific type?

jimtng:
.item is good I think

jimtng:
I've added GenericItem#item to help with comparing item objects, i.e.

```ruby
[Item1, Item2].include?(Item1.item)
```

boc-tothefuture:
@jimtng - LGTM. I did have a question.  For the 'item' method my initial thought was that it would just delegate to `java_class`.  Any idea what the pros/cons of creating the GenericItem object are instead?

jimtng:
> @jimtng - LGTM. I did have a question. For the 'item' method my initial thought was that it would just delegate to `java_class`. Any idea what the pros/cons of creating the GenericItem object are instead?

TBH I didn't realise that we could do that. Hold off, let me try it out.

jimtng:
@boc-tothefuture if `other` is `Item.java_class`, how can I get the instance object from `other` in order to compare it against `self`? My understanding is that `java_class` gives me the class, not the instance.

boc-tothefuture:
> @boc-tothefuture if `other` is `Item.java_class`, how can I get the instance object from `other` in order to compare it against `self`? My understanding is that `java_class` gives me the class, not the instance.

You are right, `java_class` is not the correct thing, I was thinking about getting the undecorated java object - but I am not sure why that would be better because I suspect that would mess up other comparisons. 

I think what you have is good.

jimtng:
In that case, no further changes are needed

jimtng:
I've just removed the test for `Item.item.label` - it's not going to be used in real life I think.

jimtng:
Update:

* Support the scenario where the left side of the operator is Item.item, e.g. `Item.item == Item`
* Support filtering items using `grep`, e.g. `MyGroup.grep_v(Item1.item)` and explain the difference in comparisons.md

jimtng:
The second commit adds support for using grep to filter for item objects

ccutrer:
> The second commit adds support for using grep to filter for item objects

And this is where I risk being a broken record, and point out that we're going to keep having gotchas like this that don't work, because I really feel it's the wrong object model to allow direct comparison between items and states, even though it is super convenient for many cases. Even "you have to use .item to compare items not states" is a super weird gotcha.

jimtng:
> > The second commit adds support for using grep to filter for item objects
> 
> And this is where I risk being a broken record, and point out that we're going to keep having gotchas like this that don't work, because I really feel it's the wrong object model to allow direct comparison between items and states, even though it is super convenient for many cases. Even "you have to use .item to compare items not states" is a super weird gotcha.

I totally agree that this is really a big bummer, I really do. 

I just can't think of a better way other than making a breaking change, so that everything we do, must change from
`if NumberItem1.positive?` -> `if NumberItem1.state.positive?`

It's a big change, and I don't know the answer.

What you're proposing does sound like the **proper** way to do it.

@boc-tothefuture this is worth re-thinking again going forward because this issue is **core** to the DSL syntax we're building here.

I also understand the convenience of treating the items like the state, but it does come at a price like this.

If we are going towards having to use `.state` to deal with the item's value, then this PR can be scrapped as it's not necessary.

boc-tothefuture:
> > > The second commit adds support for using grep to filter for item objects
> > 
> > 
> > And this is where I risk being a broken record, and point out that we're going to keep having gotchas like this that don't work, because I really feel it's the wrong object model to allow direct comparison between items and states, even though it is super convenient for many cases. Even "you have to use .item to compare items not states" is a super weird gotcha.
> 
> I totally agree that this is really a big bummer, I really do.
> 
> I just can't think of a better way other than making a breaking change, so that everything we do, must change from `if NumberItem1.positive?` -> `if NumberItem1.state.positive?`
> 
> It's a big change, and I don't know the answer.
> 
> What you're proposing does sound like the **proper** way to do it.
> 
> @boc-tothefuture this is worth re-thinking again going forward because this issue is **core** to the DSL syntax we're building here.
> 
> I also understand the convenience of treating the items like the state, but it does come at a price like this.
> 
> If we are going towards having to use `.state` to deal with the item's value, then this PR can be scrapped as it's not necessary.

I am always happy to have a discussion and reconsider.  I would want to look at all the use cases (and the frequency of each) to see what is the best path.  I know in my rules, the frequency is super highly tilted towards checking the state for almost all comparisons.  One of the things I disliked about the DSL and Python was always having to type out state.

jimtng:
> I would want to look at all the use cases (and the frequency of each) to see what is the best path. I know in my rules, the frequency is super highly tilted towards checking the state for almost all comparisons.

I would say that that's the case in my rules too. It is very rare that I had to compare the item objects, and the normal use is to compare their states. The only thing I came across, and tripped up on, would be grep(Item).

One of the many reasons I like jruby scripting is the convenience of not calling .state all the time.

While I understand @ccutrer's objections, I would like to move to keep the ability to use direct items for state comparison as it is now.

boc-tothefuture:
@jimtng I know I suggested '.item'. Do you think '.type' might be better? Specifically in the case example?

jimtng:
> @jimtng I know I suggested '.item'. Do you think '.type' might be better? Specifically in the case example?

`.type` might imply the item type. If you see `Kitchen_Light.type` you might think it means SwitchItem / ColorItem / DimmerItem

boc-tothefuture:
Ok, sounds good to me.

PR #553, opened 2022-03-09
docs(links): adapt links to jekyll 3

jimtng:
As it is now, the relative links on the current docs site are broken.

In my previous PR #546 I've converted the links to using jekyll's `link` tag. Jekyll 4 will automatically add the `baseurl` to it, but [GitHub Pages still uses Jekyll 3](https://pages.github.com/versions/). Unfortunately Jekyll 3 doesn't add the baseurl so it needs to be added manually. 

This PR:
* Add the baseurl prefix to the `link` tags
* Update the Gemfile to match the version.
* Due to a newer version of rubocop, several new offenses had to be addressed
* Rubocop wanted this to be added to the gemspec - I'm not quite sure what impacts it would have: `spec.metadata['rubygems_mfa_required'] = 'true'`. I've disabled this rule.

I had to run `bundler update` to resolve the version differences, which resulted in a lot of changes in Gemfile.lock


openhab-scripting.gemspec
```diff
@@ -18,6 +18,7 @@ Gem::Specification.new do |spec|
   spec.metadata['homepage_uri'] = spec.homepage
   spec.metadata['source_code_uri'] = 'https://github.com/boc-tothefuture/openhab-jruby'
   spec.metadata['changelog_uri'] = 'https://github.com/boc-tothefuture/openhab-jruby/blob/main/CHANGELOG.md'
+  spec.metadata['rubygems_mfa_required'] = 'true'
```
ccutrer, line 21: This might be problematic for automatic publishing of the gem via GitHub actions.

jimtng: > This might be problematic for automatic publishing of the gem via GitHub actions.

I agree. I have removed it and disabled that rubocop rule.

Issue #460, opened 2021-12-10
Store item dependencies for a rule in 'deps' or 'dependencies'


boc-tothefuture:
Related to #459 I often have the following in my rules

```
rule 'foo' do
  changed [bar, baz, qux]
  run { # do something amazing if all these have a state otherwise blow up }
  only_if [bar, baz, qux]
  otherwise { # maybe log or do something else }
end
```

Proposal:

```
rule 'foo' do
  changed [bar, baz, qux]
  run { # do something amazing if all these have a state otherwise blow up }
  only_if deps
end
```

I am not quite sure if having deps in the config object will actually work here because it is run in the context of the base caller object. We wouldn't want to leak deps into that object either.  We could fix that with an additional guard that is called `only_if_deps` or some better name.

ccutrer:
I see no reason why you couldn't make `dependencies` available in the rule config block like that. You just couldn't use them out-of-order before they're defined, which doesn't seem like much of a limitation. I would definitely use `dependencies` instead of `deps` for it, though, to reduce collision chances with local variables.

PR #562, opened 2022-03-18
feat(states): add state? helper method

jimtng:
# state?

`state?` returns true if all the given items have a valid state (not UNDEF or NULL).

| Parameter       | Description                                                                                                                                                                    |
| --------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| `items`         | One or more items to check                                                                                                                                                     |
| `things:` | Default: `false` - do not check for linked things status. When `things:` is set to `true`, only return true when all things linked to the items are in the ONLINE state. |

```ruby
if state? Item1, Item2 
  average = (Item1 + Item2)/2 # Neither Item1 nor Item2 is UNDEF/NULL
end
```

or in a rule guard

```ruby
rule 'calculate' do
  changed Item1, Item2
  only_if { state? Item1, Item2 }
  run { logger.info "Average of Item1 and Item2: #{(Item1 + Item2) / 2}}" }
end
```

It can take an array of items:

```ruby
calculated_items = [ Item1, Item2 ]

rule 'calculate' do
  changed calculated_items
  only_if { state? calculated_items }
  run { logger.info "Sum of Item1 and Item2: #{calculated_items.sum}" }
end
```

A possible solution to #460 although this doesn't automatically capture the dependencies from the triggers, but it is usable in a normal code, not just tied to a rule structure.

### DIFF
--- a/docs/usage/misc/state.md
+++ b/docs/usage/misc/state.md
@@ -1,0 +1,46 @@
+---
+layout: default
+title: Valid State
+has_children: false
+parent: Misc
+grand_parent: Usage
+---
+
+# state?
+
+`state?` returns true if all the given items have a valid state (not UNDEF or NULL).
+
+| Parameter | Description                                                                                                                                                              |
+| --------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `items`   | One or more items to check                                                                                                                                               |
+| `things:` | Default: `false` - do not check for linked things status. When `things:` is set to `true`, only return true when all things linked to the items are in the ONLINE state. |
+
+## Examples
+
+```ruby
+if state? Item1, Item2 
+  average = (Item1 + Item2)/2 # Neither Item1 nor Item2 is UNDEF/NULL
+end
+```
+
+or in a rule guard
+
+```ruby
+rule 'calculate' do
+  changed Item1, Item2
+  only_if { state? Item1, Item2 }
+  run { logger.info "Average of Item1 and Item2: #{(Item1 + Item2) / 2}}" }
+end
+```
+
+It can take an array of items:
+
+```ruby
+calculated_items = [ Item1, Item2 ]
+
+rule 'calculate' do
+  changed calculated_items
+  only_if { state? calculated_items }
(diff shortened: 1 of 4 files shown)


boc-tothefuture:
Nice PR.  
I had been thinking about this and wondering if it something to attach to the triggers, but wasn't sure how that would look.

```
changed foo1, foo2, states: true
```
I don't love the feel of it, but it prevents having to capture stuff into an array.


I don't think adding this PR (assuming we can figure out the name), precludes adding something to the rule language.

jimtng:
I think valid_states? sounds better

jimtng:
I renamed this to `state?`. It complements `Item.state?`

jimtng:
> Nice PR. I had been thinking about this and wondering if it something to attach to the triggers, but wasn't sure how that would look.
> 
> ```
> changed foo1, foo2, states: true
> ```
> 
> I don't love the feel of it, but it prevents having to capture stuff into an array.
> 

We could "capture" all the items that `changed` and `updated` are tracking within a rule and access it using `tracked_items`. This way you could add `only_if { state? rule.tracked_items } ` guard

In the scenario where you want other items to also be valid, simply add `only_if { state? rule.tracked_items, AnotherItem }`

boc-tothefuture:
> > Nice PR. I had been thinking about this and wondering if it something to attach to the triggers, but wasn't sure how that would look.
> > 
> > ```
> > changed foo1, foo2, states: true
> > ```
> > 
> > I don't love the feel of it, but it prevents having to capture stuff into an array.
> > 
> 
> We could "capture" all the items that `changed` and `updated` are tracking within a rule and access it using `tracked_items`. This way you could add `only_if { state? rule.tracked_items } ` guard
> 
> In the scenario where you want other items to also be valid, simply add `only_if { state? rule.tracked_items, AnotherItem }`
> 

I think that is a good solution.

jimtng:
Thanks for the suggestion. check_things didn't sound great.

state? Item, thing: true

Doesn't seem right either.

How about things_check?

jimtng:
I suppose in the interest of brevity, `things` could work.

PR #563, opened 2022-03-29
test: test against a new addon

jimtng:
Testing an addon that converts any lower case value to a global


jimtng:
The [Addon PR](https://github.com/openhab/openhab-addons/pull/12539) has been merged

PR #568, opened 2022-04-03
fix(zoneddatetime): make ZonedDateTime available on OpenHAB 3.2

jimtng:
Move the java_import for ZonedDateTime to the top of dsl.rb so it's available for user scripts.
Remove instances of `java_import java.time.ZonedDateTime` in other places

This makes it no longer necessary to import ZonedDateTime within user scripts even on openhab 3.2.
This also deduplicates the multiple imports in submodules.

By "importing", I mean re-assigning the ZonedDateTime constant to the Ruby class version. I'm not sure how "expensive" this operation is, but I suspect it's very "cheap" / fast to execute.

### DIFF
--- a/docs/examples/how_do_i.md
+++ b/docs/examples/how_do_i.md
@@ -741,16 +741,7 @@ Exec.executeCommandLine('/path/to/program')
 
 ### Use ZonedDateTime
 
-Prior to OpenHAB 3.3, ZonedDateTime needs to be imported before use. From OpenHAB 3.3+ the following classes are available without having to import them manually:
-
-* `ZonedDateTime`
-* `ZoneId`
-* `Duration`
-* `ChronoUnit`
-
 ```ruby
-java_import java.time.ZonedDateTime # Not needed on OpenHAB 3.3+
-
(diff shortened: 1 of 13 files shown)


boc-tothefuture:
What is the reason for using ZDT instead of the ruby time/date objects and converting to ZDT where necessary?

jimtng:
Sometimes using ZDT is unavoidable, other times, zdt has nicer methods that Ruby Time doesn't have. For example, plus_minutes, plus_hours, etc without having to do maths. It also naturally uses java Duration for deltas.

jimtng:
Currently failing on `on_start` execution context.

boc-tothefuture:
> Sometimes using ZDT is unavoidable, other times, zdt has nicer methods that Ruby Time doesn't have. For example, plus_minutes, plus_hours, etc without having to do maths. It also naturally uses java Duration for deltas.

Makes sense.

jimtng:
This was a bit epic to fix. I had to re-import in rule executions and also in timer executions. I wonder if it's worth doing. Is the performance impact acceptable?

I added a guard in the import method to avoid importing if ZDT is already correct.

PR #575, opened 2022-04-06
feat(timers[]): add #reschedule, rename #cancel_all to #cancel

jimtng:
Most of the time, only one reentrant timer is created / manipulated, so adding a helper method to cancel and reschedule would be nice.

* Add timers[id].reschedule as an alias for timers[id].each(&:reschedule)
* Add timers[id].cancel as an alias for timers[id].each(&:cancel) - the previous timers[id].cancel_all is made an alias for cancel.
* Minor tweak to the timers doc


jimtng:
I'm torn between this implementation (i.e. just have timers[id].cancel and timers[id].reschedule) vs delegating timer object's methods to timers[id].to_a.first

PR #583, opened 2022-04-18
feat(semantics): incorporate flat_map into Enumerable#points

jimtng:
This will remove the need to call flat_map before calling Enumerable#points

Before:
```
Item.location
.equipments
.flat_map(&:members)
.points
```

after:
```
Item.location
.equipments
.points
```

Also add Enumerable#members, which is an idea from @ccutrer
This helps when doing something like `gLocation.equipments.members.member_of(NonSemanticGroup)`


jimtng:
@ccutrer, I added Enumerabl#members in the second commit

PR #598, opened 2022-05-14
Optimize ci by reusing runners

### LINKED ISSUE #310: Decrease test time by reusing runners

jimtng:
Resolve #310

* I started off from @boc-tothefuture's work.
* a @log_level_changed was missing in one of the scenarios in rule_language.feature. The scenario set the log level to INFO causing subsequent tests using trace log to fail
* added `index` to the matrix array and use it in openhab dump file. Previously it used the very long feature names which also contain illegal characters `:` and `/`

Other non-essential changes:
* Added --no-heap-dump to dev:dump-create to reduce the openhab dump file size (in the case of test failures)
* Updated test against 3.3.0.M5
* Replaced / simplified several ruby setup actions with ruby/setup-ruby + bundler-cache in workflow.yml
* Close the file handle in truncate_log - I'm not sure whether the file handles would've gotten closed anyway by Ruby's GC at the end of the method call, but it's better to be sure

This brings the test time down from 20m to 12m - 15m

I believe this should be ready to be merged

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -1,6 +1,6 @@
 name: Openhab-JRuby-Scripting
 
-on: 
+on:
   pull_request:
   push:
     branches:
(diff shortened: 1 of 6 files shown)


jimtng:
It seems that some tests are still failing!

jimtng:
There seems to be two problems:

1. The fact that some (specific) tests failed - they are the same tests that fail consistently
2. cucumber --retry in test.rake doesn't seem to be doing what it's supposed to do. What I believe we want it to do:
    * if it failed, retry, and if the second attempt is successful, continue to the next test
    * if subsequent retry is successful (up to the specified limit 3), then consider the test successful, and exit with error code 0

   What it actually does:
   * if it failed, retry 3 times, even when all the subsequent attempts were successful, still do it 3 times in addition to the initial failure, making it run 4 times.
   * Even when subsequent attempts were successful, exit with exit code 1, causing the entire thing to fail.

I found an old issue that doesn't seem to be resolved:
https://github.com/cucumber/cucumber-ruby/issues/1000

As for the issue with (1), I am not quite sure why those rules failed. I downloaded the dump file but the retries seem to have overwritten the logs, unless I've missed something.

jimtng:
I finally found out why the rules were failing! One of the scenarios in rule_language.feature was missing `@log_level_changed` tag, so when the log level was changed to INFO it stayed that way. Subsequent unrelated tests that use trace logging will then fail.

PR #601, opened 2022-06-09
feat: add support for persistence 'between' methods

jimtng:
Add support for https://github.com/openhab/openhab-core/pull/2931


boc-tothefuture:
Should the methods returning a boolean have a question mark on the end?

jimtng:
I've changed the boolean methods to use a question mark and aliased the non question mark version.

I've also added yard docs for the other methods.

PR #606, opened 2022-06-23
Jruby94 compatibility

jimtng:
This PR makes changes to the code base to be compatible with JRuby 9.4 / Ruby 3.x. It can be merged now because it is backwards compatible and should pass our current tests against jruby 9.3.

A separate draft PR will be created to add specific jruby 9.4 tests. That PR doesn't need to be merged yet until jruby 9.4 is in "beta".

### DIFF
--- a/lib/openhab/dsl/items/ensure.rb
+++ b/lib/openhab/dsl/items/ensure.rb
@@ -88,6 +88,7 @@ module OpenHAB
             @item.__send__(method, *args, &block)
           end
         end
+        ruby2_keywords :method_missing if respond_to? :ruby2_keywords
 
         # .
         def respond_to_missing?(method, include_private = false)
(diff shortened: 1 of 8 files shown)


lib/openhab/dsl/items/ensure.rb
```diff
@@ -81,11 +81,15 @@ def <<(command)
         end
 
         # activate +ensure_states+ before forwarding to the wrapped object
-        def method_missing(method, *args, &block)
+        def method_missing(method, *args, **kwargs, &block)
```
ccutrer, line 84: the proper way to fix this is with ruby2_keywords. normally you'd run into issues with ruby 2.7, but since jruby is skipping ruby 2.7, this _might_ be okay, but still concerns me

jimtng: I've pushed a version that uses ruby2_keywords. It passed the jruby94 test too

PR #610, opened 2022-06-30
fix(terse): improve automatic naming of terse rules

ccutrer:
construct a useful name if possible, otherwise use the full location of the execution block, including line number.

### DIFF
--- a/features/rule_language.feature
+++ b/features/rule_language.feature
@@ -91,12 +91,12 @@ Feature: rule_language
     And a rule
       """
       changed TestSwitch do
-        logger.trace("switch changed")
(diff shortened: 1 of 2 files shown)


boc-tothefuture:
Do we need to update any user docs for this?

ccutrer:
I don't think so. It's just nice for when file based rules show up in the UI

ccutrer:
actually, scratch that. looks like UI just shows the filename still. so... I'm not sure what this is useful for, other than when I'm doing IRB debugging, to be able to look up rules

ccutrer:
🤦 🤦 . Yes, it does work for the UI. My openhab install just wasn't using this code when the rules file was loaded

ccutrer:
superseded by https://github.com/boc-tothefuture/openhab-jruby/pull/617

PR #613, opened 2022-07-03
fix(ensure): with QuantityType and plain number

ccutrer:
previously it was getting converted to a straight string, which is
not comparable against a QuantityType. we might as well send
raw types through


ccutrer:
@boc-tothefuture : any comment on this one?

PR #618, opened 2022-07-18
chore(rubocop): explicitly allow openhab globals

ccutrer:
instead of having to disable it at each usage


jimtng:
This is a great idea!

PR #619, opened 2022-07-22
fix(rule): openHAB doesn't reload rules on save

### LINKED ISSUE #620: Rules reloading is broken in 4.43.1
When a rule file is modified and saved, the new rule doesn't seem to take effect. The old rule continues to run.

To reproduce:

```ruby
rule 'x' do
  received_command(SwitchItem1)
  run { logger.info 'T1' }
end

SwitchItem1.on
```

Then change T1 to T2, then save. Change it again to T3 and save. Openhab will continue to log `T1`

jimtng:
Fix #620
rule ids need to be unique for openHAB to load the new rule

### DIFF
--- a/lib/openhab/dsl/rules/rule.rb
+++ b/lib/openhab/dsl/rules/rule.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'method_source'
+require 'securerandom'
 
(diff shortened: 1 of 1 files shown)


jimtng:
Fixed in #622

PR #622, opened 2022-07-25
fix(rules): fix reloading of rules

### LINKED ISSUE #620: Rules reloading is broken in 4.43.1

ccutrer:
fixes #620

we were changing the ID of the action after it was registered, so
when it unregistered it couldn't find its handler, and was orphaning
it. only to reuse the old handler on the reload because the id hadn't
been changed yet

### DIFF
--- a/lib/openhab/dsl/rules/rule.rb
+++ b/lib/openhab/dsl/rules/rule.rb
@@ -89,7 +89,7 @@ module OpenHAB
           rule = AutomationRule.new(config: config)
           Rule.script_rules << rule
           added_rule = add_rule(rule)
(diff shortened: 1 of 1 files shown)


jimtng (approved):
Thanks!

PR #623, opened 2022-07-26
feat(rules): improve rule name inference

ccutrer:
* push it down into rule itself, not just terse, but catch rules
   that only have a single inferrable trigger
 * support duration on item triggers
 * support "every" triggers
 * support channel triggers
 * support multiple things for many parameters

### DIFF
--- a/lib/openhab/dsl/rules/name_inference.rb
+++ b/lib/openhab/dsl/rules/name_inference.rb
@@ -1,0 +1,133 @@
+# frozen_string_literal: true
+
+module OpenHAB
+  module DSL
+    module Rules
+      # Contains helper methods for inferring a rule name from its triggers
+      module NameInference
+        # Trigger Type UIDs that we know how to generate a name for
+        KNOWN_TRIGGER_TYPES = [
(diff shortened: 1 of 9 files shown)


jimtng (commented):
This is very nice!

lib/openhab/dsl/rules/name_inference.rb
```diff
@@ -0,0 +1,133 @@
...
+          end
+
+          # formulate a readable rule name such as "TestSwitch received command ON" if possible
+          def infer_rule_name(config) # rubocop:disable Metrics
+            known_triggers, unknown_triggers = config.triggers.partition do |t|
+              KNOWN_TRIGGER_TYPES.include?(t.type_uid)
+            end
+            return nil unless unknown_triggers.empty?
+
+            cron_triggers = known_triggers.select { |t| t.type_uid == 'jsr223.jruby.CronTrigger' }
```
jimtng, line 35:
```suggestion
            cron_triggers = known_triggers.select { |t| t.type_uid == Triggers::Cron::CRON_TRIGGER_MODULE_ID }
```

PR #639, opened 2022-09-09
bring in rspec-openhab-scripting gem

ccutrer:
and convert (most) _type.feature tests to rspec in-process tests.

By running tests in-line, and in-process, massive speed increases are possible, as well as ease of writing tests. You don't have to wrap everything in a rules file, that then gets deployed, wait for OpenHAB to detect it, and then check the results by continually checking the log file. This also allows more particular testing (like the rspec `be` matcher in the `PercentType` spec that also checks for type equality).

For this initial proof-of-concept, I converted 28+ cucumber scenarios to specs. The runtime on my computer:

```
cody@opennuc:~/src/openhab-jruby$ time rspec

Randomized with seed 8442
............................

Finished in 18.4 seconds (files took 2.48 seconds to load)
28 examples, 0 failures

Randomized with seed 8442

real	0m26.233s
user	1m46.645s
sys	0m2.334s
```

The vast majority of that time is booting up - the actual specs (all of them, not each one) took a fraction of a second to run. This is in stark contrast to cucumber's infrastructure that takes 30-60s to start openhab, and then a minimum of 5 seconds per scenario.


spec/examples.txt
```diff
@@ -0,0 +1,30 @@
+example_id                                                    | status | run_time        |
```
ccutrer, line 1: I need to ignore this file, but for now it's perfect for demonstrating how fast each individual spec is.

spec/openhab/dsl/types/decimal_type_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+RSpec.describe 'OpenHAB::DSL::Types::DecimalType' do
+  it 'is inspectable' do
+    expect(DecimalType.new(10)).to eq '10'
```
jimtng, line 5: missing `.inspect` here?

ccutrer: yup, whoops. TIL `DecimalType.new(10) == '10'` is true. fixed, and also changed to `eql` matcher to be more particular in the future.

ccutrer:
@jimtng @boc-tothefuture : this is ready for review now!

ccutrer:
Note that with this commit, you can also do `bundle console`, and get a JRuby REPL _inside_ an OpenHAB environment. It's amazing for exploring how things work when you're implementing new features/fixing bugs.

.github/workflows/workflow.yml
```diff
@@ -201,7 +241,7 @@ jobs:
       # otherwise the gem build will fail because it is not a development instance
       - uses: actions/setup-ruby@v1
         with:
-          ruby-version: ${{ env.RUBY_VERSION }}
+          ruby-version: '2.6.8'
```
jimtng, line 244: Is there a reason for not using the `env.RUBY_VERSION`? This includes the `ruby-version` additions above.

jimtng (commented):
I like the speed improvements and the more straight forward tests. 

I had a look at the test run and noticed that there's a lack of details in the test output. I suppose failures would show up there. I've just gotten used to seeing all the details from cucumber.

Lastly, we'd obviously lose the pickles pages for the tests that are no longer done in cucumber, such as https://boc-tothefuture.github.io/openhab-jruby/pickles/decimal_type.html
I don't usually look at the pickles output and I doubt anyone does.

.github/workflows/workflow.yml
ccutrer: because I explicitly want MRI 2.6.8 here. I had to remove the .ruby-version file so that for the rspec tests it would even be allowed to run with JRuby 9.3.8.0, but in so doing I have to be explicit for the other uses. And I still want to use MRI for these both to make sure nothing changes in them, and also because MRI still boots much faster than JRuby, so for reasons when strict compatibility (just running cucumber, but the tests themselves run on JRuby inside OpenHAB; running rubocop, etc.) isn't necessary, it's still preferred.

ccutrer:
Oops. Accidentally pushed to the wrong repo, "merging this". Need to undo.

PR #647, opened 2022-10-06
test: bring in rspec-openhab-scripting gem

ccutrer:
:sigh: yet again.

PR #653, opened 2022-11-30
Revert "test: bring in rspec-openhab-scripting gem"

ccutrer:
Reverts boc-tothefuture/openhab-jruby#647. It's not compatible with openHAB 3.4.0, and the upstream rspec-openhab-scripting gem is no longer maintained seperately. Easiest to just revert it.


boc-tothefuture:
Thanks for the PR!

This was reverted as part of #651 which was just merged.

I am going to close this PR since it is a duplicate.

